PR #8, opened 2017-02-20
feat: add Traced

bumbleblym:
i took a stab at adding Traced, based on Writer/Reader.


gcanti:
Hi @bumbleblym, thanks!

PR #14, opened 2017-03-06
fix: fix compose generic type positions

bumbleblym:
use the same generic type scheme for `curry`

this happens to be a breaking change when `compose` is used with 5 or more functions, but actual return type would have been different from inferred type.


gcanti:
Nasty bug.. thanks a lot @bumbleblym

gcanti:
Released a patch.

@bumbleblym out of curiosity, are you using fp-ts for a real world project? fp-ts started as a R&D project but then I succesfully used it in [io-ts](https://github.com/gcanti/io-ts) (leveraging the `Either` monad) and [monocle-ts](https://github.com/gcanti/monocle-ts) (functional optics)

bumbleblym:
wish i could tell you i was, but i'm only hacking on this package out of mathematical curiosity. i haven't used these types in actual projects, whether from this package or other packages/languages.

by the way, have you tried [jsverify](https://github.com/jsverify/jsverify)?

PR #20, opened 2017-03-26
New technique to fake higher kinded types (v0.2)

gcanti:
This is a major refactoring (and a breaking change). 

I found a new technique to fake higher kinded types which provides a (way) better type inference. Also the code is simpler and cleaner.

If anyone is willing to check the proposed implementation I'll be glad to share more details.

**The problem**

Now if we define a generic `lift` function we get

```ts
import * as option from 'fp-ts/lib/Option'
import { StaticFunctor } from 'fp-ts/lib/Functor'
import { HKT } from 'fp-ts/lib/HKT'

function lift<F, A, B>(functor: StaticFunctor<F>, f: (a: A) => B): (fa: HKT<F, A>) => HKT<F, B> {
  return fa => functor.map(f, fa)
}

const double = (n: number): number => n * 2

const f = lift(option, double)
```

The problem is that `f` has type `(fa: HKT<"Option", number>) => HKT<"Option", number>` so we must cast `HKT<"Option", number>` to `Option<number>` manually (or with module augmentations as explained in the README).

With the new technique we get

```ts
import * as option from '../src/Option'
import { StaticFunctor } from '../src/Functor'
import { HKT } from '../src/HKT'

function lift<F extends keyof HKT<any>, A, B>(functor: StaticFunctor<F>, f: (a: A) => B): (fa: HKT<A>[F]) => HKT<B>[F] {
  return fa => functor.map(f, fa)
}

const double = (n: number): number => n * 2

const f = lift(option, double)
```

Now `f` has the correct type `(fa: option.Option<number>) => option.Option<number>` !

**The solution**

Here's the gist

Higher kinded types are represented by a unique string literal (called `URI`).

There's a central type dictionary where a mapping `URI` -> concrete type is stored

```ts
// file ./HTK.ts
export interface HKT<A> {}
```

Instances can be defined (anywhere) using a feature called [Module Augmentation](https://www.typescriptlang.org/docs/handbook/declaration-merging.html) so there's no danger of name conflict (the typechecker checks for duplicates).

Here's a mapping between the string literal `'Option'` and the concrete type `Option<A>`

```ts
// file ./Option.ts
declare module './HKT' {
  interface HKT<A> {
    Option: Option<A>
  }
}

export const URI = 'Option'
export type URI = typeof URI
export type Option<A> = None<A> | Some<A>

export class None<A> {
  readonly _URI: URI
  map<B>(f: (a: A) => B): Option<B> {
    return none
  }
  ...
}

export Some<A> {
  readonly _URI: URI
  // fantasy-land implementation
  map<B>(f: (a: A) => B): Option<B> {
    return new Some(f(this.value))
  }
  ...
}

// static-land implementation
export function map<A, B>(f: (a: A) => B, fa: Option<A>): Option<B> {
  return fa.map(f)
}
```

Concrete types can be retrieved by their `URI` using a feature called [Index types](https://www.typescriptlang.org/docs/handbook/advanced-types.html).

Type classes are implemented following (when possible) both the [static-land](https://github.com/rpominov/static-land) spec and the [fantasy-land](https://github.com/fantasyland/fantasy-land) spec.

 Here's the definition of the type class `Functor` (following the static-land spec)

```ts
export interface StaticFunctor<F extends keyof HKT<any>> {
  URI: F
  map<A, B>(f: (a: A) => B, fa: HKT<A>[F]): HKT<B>[F]
}
```

/cc @danielepolencic This new technique mainly affects (in a good way) the static-land implementation, so you might be interested


gcanti:
I checked this implementation against

- [io-ts](https://github.com/gcanti/io-ts) (no updates required)
- [monocle-ts](https://github.com/gcanti/monocle-ts) (some minor updates)

and against two new libraries

- [elm-ts](https://github.com/gcanti/elm-ts)
- [parser-ts](https://github.com/gcanti/parser-ts)

Everything seems fine so far. I'm leaning towards a v0.2 release

gcanti:
Good news: I'm now able to define monad transformers with decent type inference

**Example**: `OptionT`

```ts
import { HKT, HKTS } from '../src/HKT'
import { FantasyMonad, StaticMonad } from '../src/Monad'
import { Option } from '../src/Option'
import * as option from '../src/Option'

declare module '../src/HKT' {
  interface HKT<A> {
    OptionT: OptionT<any, A>
  }
}

export const URI = 'OptionT'

export type URI = typeof URI

export class OptionT<M extends HKTS, A> implements FantasyMonad<URI, A> {
  readonly _URI: URI
  of: (a: A) => OptionT<M, A>
  constructor(public readonly monad: StaticMonad<M>, public readonly value: HKT<Option<A>>[M]) {
    this.of = of<M>(monad)
  }
  map<B>(f: (a: A) => B): OptionT<M, B> {
    return new OptionT<M, B>(this.monad, this.monad.map((o: Option<A>) => o.map(f), this.value))
  }
  ap<B>(fab: OptionT<M, (a: A) => B>): OptionT<M, B> {
    return fab.chain(f => this.map(f)) // <- derived
  }
  chain<B>(f: (a: A) => OptionT<M, B>): OptionT<M, B> {
    return flatten(this.map(f))
  }
}

export function flatten<M extends HKTS, A>(mma: OptionT<M, OptionT<M, A>>): OptionT<M, A> {
  return new OptionT<M, A>(mma.monad, mma.monad.chain<Option<OptionT<M, A>>, Option<A>>((o: Option<OptionT<M, A>>) => o.fold(
    () => mma.monad.of<Option<A>>(option.none),
    x => x.value
  ), mma.value))
}

export function of<M extends HKTS>(monad: StaticMonad<M>): <A>(a: A) => OptionT<M, A> {
  return <A>(a: A) => new OptionT<M, A>(monad, monad.of(option.of(a)))
}

export function map<M extends HKTS, A, B>(f: (a: A) => B, fa: OptionT<M, A>): OptionT<M, B> {
  return fa.map(f)
}

export function ap<M extends HKTS, A, B>(fab: OptionT<M, (a: A) => B>, fa: OptionT<M, A>): OptionT<M, B> {
  return fa.ap(fab)
}

export function chain<M extends HKTS, A, B>(f: (a: A) => OptionT<M, B>, fa: OptionT<M, A>): OptionT<M, B> {
  return fa.chain(f)
}
```

/cc @gabro

gcanti:
I changed a few signatures in order to get better type inference

danielepolencic:
Congrats, good progress!

gcanti:
@danielepolencic thanks! I worked hard and I'm really excited for this release :)

sledorze:
Must appreciated! Thanks for the outstanding efforts :)

PR #45, opened 2017-04-26
fix Some.reduce so it calls f

leemhenson:
Hi. I'm switch a project over from flow and flow-static-land to ts and fp-ts. In updating a section that used `Maybe` from flow-static-land to use `Option`, i noticed that the behaviour changed. It looks to me like `reduce` on `Some` isn't quite right, as it never calls `f`. This is different to how it was in flow-static-land. Is that intentional and should I be using `fold` directly instead? 

I'm still very much in the training-wheels phase with a lot of this stuff, but it strikes me that `reduce` and `fold` do very similar things. When should one be used over the other? Is it just that `reduce` should be used when you don't want to have to deal with converting to `Lazy<A>` yourself?

### DIFF
--- a/src/Option.ts
+++ b/src/Option.ts
@@ -9,7 +9,7 @@ import { StaticExtend, FantasyExtend } from './Extend'
 import { StaticSetoid } from './Setoid'
 import { StaticTraversable, FantasyTraversable } from './Traversable'
 import { StaticAlternative, FantasyAlternative } from './Alternative'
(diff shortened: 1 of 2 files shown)


gcanti:
> looks to me like reduce on Some isn't quite right, as it never calls f

Indeed is a bug! Thanks @leemhenson. I'll release a patch asap

gcanti:
Released in https://github.com/gcanti/fp-ts/releases/tag/0.2.2

PR #48, opened 2017-04-28
add Validation chain

sledorze:


gcanti:
What's different about `Validation` (w.r.t. `Either`) is that it is not a monad, but it's an applicative functor

https://hackage.haskell.org/package/validation

(`AccValidation` -> `Validation` in fp-ts)

> The AccValidation data type is isomorphic to Either, but has an instance of Applicative that accumulates on the error side. That is to say, if two (or more) errors are encountered, they are appended using a Semigroup operation.
As a consequence of this Applicative instance, there is no corresponding Bind or Monad instance. AccValidation is an example of, An applicative functor that is not a monad

sledorze:
That s perfectly clear and make sense!

gcanti:
@sledorze I thought a lot about Either and Validation and I no longer believe that Validation can't have a monad instance

sledorze:
@gcanti curious about what has changed your mind?

gcanti:
@sledorze The main argument is that the "hand-made" `ap` and the `ap` derived from `chain` won't be "equal", But what's the meaning of "equal" here?

If we are talking about validations I can define a suitable `Setoid<L>` instance (e.g. the trivial one) under which those two `ap`s appear to be equal. Switching back and forth between `Either<L, A>` and `Validation<L, A>` is not meaningful.

I'm for adding `chain` to Validation.

raveclassic:
@gcanti 
> I'm for adding chain to Validation.

Maybe we could pick another name?

gcanti:
@raveclassic I mean an actual monad instance, I'd like being able to use Validation where a `<M>(M: Monad<M>)` constraint is required

PR #59, opened 2017-05-03
improved docs

danielepolencic:
The current docs are written from the perspective of the library's author.

This PR is more user friendly and assumes the audience is a developer interested in using the library.


gcanti:
Thanks @danielepolencic

PR #63, opened 2017-05-06
fix errors in Either 'app' and Apply 'applyFirst'

sledorze:
I've based the additional tests from results of Purescript implementations and fixed the code accordingly.

### DIFF
--- a/src/Apply.ts
+++ b/src/Apply.ts
@@ -1,6 +1,6 @@
 import { HKT, HKTS, HKT2, HKT2S } from './HKT'
 import { StaticFunctor, FantasyFunctor } from './Functor'
-import { Curried2, Curried3, Curried4, identity } from './function'
+import { Curried2, Curried3, Curried4, identity, constant } from './function'
 
 export interface StaticApply<F extends HKTS> extends StaticFunctor<F> {
   ap<A, B>(fab: HKT<(a: A) => B>[F], fa: HKT<A>[F]): HKT<B>[F]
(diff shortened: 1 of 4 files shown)


gcanti:
@sledorze two nitpicks before merging: I think we could use `constant` instead of defining a new function every time `applyFirst` is called. Perhaps we could also extract `() => identity` from `applySecond`

```ts
const constIdentity = () => identity

export function applySecond<F extends HKT2S>(apply: StaticApply<F>): <L, A, B>(fa: HKT2<L, A>[F], fb: HKT2<L, B>[F]) => HKT2<L, B>[F]
export function applySecond<F extends HKTS>(apply: StaticApply<F>): <A, B>(fa: HKT<A>[F], fb: HKT<B>[F]) => HKT<B>[F]
export function applySecond<F extends HKTS>(apply: StaticApply<F>): <A, B>(fa: HKT<A>[F], fb: HKT<B>[F]) => HKT<B>[F] {
  return <A, B>(fa: HKT<A>[F], fb: HKT<B>[F]) => apply.ap(apply.map(constIdentity, fa), fb)
}
```

What do you think?

gcanti:
However we can do it later, I suspect I left around other micro-optimizations like this, for example here https://github.com/gcanti/fp-ts/blob/master/src/Monoid.ts#L25

src/Apply.ts
```diff
@@ -13,13 +13,13 @@ export interface FantasyApply<F extends HKTS, A> extends FantasyFunctor<F, A> {
...
 export function applySecond<F extends HKTS>(apply: StaticApply<F>): <A, B>(fa: HKT<A>[F], fb: HKT<B>[F]) => HKT<B>[F] {
-  return <A, B>(fa: HKT<A>[F], fb: HKT<B>[F]) => apply.ap<B, B>(apply.map(() => identity, fa), fb)
+  return <A, B>(fa: HKT<A>[F], fb: HKT<B>[F]) => apply.ap<B, B>(apply.map<A, (x: B) => B>((_: A) => identity, fa), fb)
 }
 
 export function applyFirst<F extends HKT2S>(apply: StaticApply<F>): <L, A, B>(fa: HKT2<L, A>[F], fb: HKT2<L, B>[F]) => HKT2<L, A>[F]
 export function applyFirst<F extends HKTS>(apply: StaticApply<F>): <A, B>(fa: HKT<A>[F], fb: HKT<B>[F]) => HKT<A>[F]
 export function applyFirst<F extends HKTS>(apply: StaticApply<F>): <A, B>(fa: HKT<A>[F], fb: HKT<B>[F]) => HKT<A>[F] {
-  return <A, B>(fa: HKT<A>[F], fb: HKT<B>[F]) => apply.ap<A, A>(apply.map(() => identity, fb), fa)
+  return <A, B>(fa: HKT<A>[F], fb: HKT<B>[F]) => apply.ap<B, A>(apply.map<A, (x: B) => A>((a: A) => (b: B) => a, fa), fb)
```
sledorze: oh.. that one :)

src/Apply.ts
```diff
@@ -13,13 +13,13 @@ export interface FantasyApply<F extends HKTS, A> extends FantasyFunctor<F, A> {
 export function applySecond<F extends HKT2S>(apply: StaticApply<F>): <L, A, B>(fa: HKT2<L, A>[F], fb: HKT2<L, B>[F]) => HKT2<L, B>[F]
 export function applySecond<F extends HKTS>(apply: StaticApply<F>): <A, B>(fa: HKT<A>[F], fb: HKT<B>[F]) => HKT<B>[F]
 export function applySecond<F extends HKTS>(apply: StaticApply<F>): <A, B>(fa: HKT<A>[F], fb: HKT<B>[F]) => HKT<B>[F] {
-  return <A, B>(fa: HKT<A>[F], fb: HKT<B>[F]) => apply.ap<B, B>(apply.map(() => identity, fa), fb)
+  return <A, B>(fa: HKT<A>[F], fb: HKT<B>[F]) => apply.ap<B, B>(apply.map<A, (x: B) => B>((_: A) => identity, fa), fb)
```
sledorze: @gcanti I am not sure if we should help or not the compiler (and curious reader) with type annotations (compiler, for speed).

sledorze:
@gcanti I prefer picky code reviews like that rather than "let's do it later.." :)

gcanti:
Ok, great! ;)

PR #86, opened 2017-05-15
add refinements, closes #68

### LINKED ISSUE #68: Library enhancement with typescript Predicates?
I am wondering how well it would fit into fp-ts to take advantage of narrowing with native predicate types.
For instance, I'm often using that kind of type-safe function:

```
const filterPredicate = <T, U extends T>(pred: (x: T) => x is U, arr: Array<T>): Array<U> =>
  arr.filter(pred) as Array<U>

```

If we adapt for for higher kinds, I don't really know how we would name it.

What do you think @gcanti ?

gcanti:
@sledorze I'm not sure how the `and` combinator could be useful in the case of custom guards, have you got a compelling example?

### DIFF
--- a/src/Array.ts
+++ b/src/Array.ts
@@ -10,7 +10,7 @@ import { liftA2 } from './Apply'
 import { Option } from './Option'
 import * as option from './Option'
 import { StaticOrd, toNativeComparator } from './Ord'
-import { Predicate, identity, constant, curry, Lazy, Endomorphism } from './function'
+import { Predicate, identity, constant, curry, Lazy, Endomorphism, Refinement } from './function'
 
 declare module './HKT' {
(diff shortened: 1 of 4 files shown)


gcanti:
@sledorze refinements seems to make sense only for primitives and class inheritance, AFAIK the `and` combinator wouldn't be correct

**or**

`B1 <: A` and `B2 <: A` implies `B1 | B2 <: A`

**and**

if `B1 <: A` and `B2 <: A` then is not true that `B1 & B2 <: A`

sledorze:
@gcanti thanks for the mathematical insight, you're right, even if some usages would be correct, the 'and' combinator is not sound and would accept incorrect usages.. let's ditch it!

PR #140, opened 2017-06-23
Allow chain to accumulate error types

DylanRJohnston:
Allows chaining Either methods to accumulate a union of error types.

Example

```
import * as T from 'io-ts'
import {Either, left, right} from 'fp-ts/lib/Either

const isBuffer = (x: any): Either<S3Error, Buffer> =>
    Buffer.isBuffer(x) ? right(x) : left(S3Error.NotBuffer)

export const validateData = (data: any):  =>
    isBuffer(data)
    .map(x => x.toString())
    .chain(x => T.validate(x, Project)) 
```

Without this chain is unable to match S3Error and T.ValidationError[]

I think this strictly speaking breaks FantasyLand spec but it seems useful to unify the error types.

Alternative solution is to cast up the Either type when you call the method, but it requires to to add all the other errors that could come along.

```
import * as T from 'io-ts'
import {Either, left, right} from 'fp-ts/lib/Either

const isBuffer = (x: any): Either<S3Error, Buffer> =>
    Buffer.isBuffer(x) ? right(x) : left(S3Error.NotBuffer)

export const validateData = (data: any):  =>
    (isBuffer(data) as Either<S3Error | T.ValidationError[], Buffer>)
    .map(x => x.toString())
    .chain(x => T.validate(x, Project)) 
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "fp-ts",
-  "version": "0.3.5",
+  "version": "0.3.6",
   "description": "Functional programming in TypeScript",
   "files": [
     "lib"
--- a/src/Either.ts
+++ b/src/Either.ts
@@ -48,7 +48,7 @@ export class Left<L, A>
   ap<B>(fab: Either<L, (a: A) => B>): Either<L, B> {
     return (isLeft(fab) ? fab : this) as any
   }
-  chain<B>(f: (a: A) => Either<L, B>): Either<L, B> {
+  chain<L2, B>(f: (a: A) => Either<L2, B>): Either<L2 | L, B> {
     return this as any
   }
   bimap<L2, B>(f: (l: L) => L2, g: (a: A) => B): Either<L2, B> {
@@ -116,7 +116,7 @@ export class Right<L, A>
     }
     return fab as any
   }
-  chain<B>(f: (a: A) => Either<L, B>): Either<L, B> {
+  chain<L2, B>(f: (a: A) => Either<L2, B>): Either<L2 | L, B> {
     return f(this.value)
   }
   bimap<L2, B>(f: (l: L) => L2, g: (a: A) => B): Either<L2, B> {
@@ -183,7 +183,7 @@ export function ap<L, A, B>(fab: Either<L, (a: A) => B>, fa: Either<L, A>): Eith
   return fa.ap(fab)
 }
 
-export function chain<L, A, B>(f: (a: A) => Either<L, B>, fa: Either<L, A>): Either<L, B> {
+export function chain<L, L2, A, B>(f: (a: A) => Either<L2, B>, fa: Either<L, A>): Either<L2 | L, B> {
   return fa.chain(f)
 }
 
--- a/src/Either.ts
+++ b/src/Either.ts
@@ -51,7 +51,10 @@ export class Left<L, A>
   ap_<B, C>(this: Either<L, (a: B) => C>, fb: Either<L, B>): Either<L, C> {
     return fb.ap(this)
   }
-  chain<L2, B>(f: (a: A) => Either<L2, B>): Either<L | L2, B> {
+  chain<B>(f: (a: A) => Either<L, B>): Either<L, B> {
+    return this as any
+  }
+  chain_<L2, B>(f: (a: A) => Either<L2, B>): Either<L | L2, B> {
     return this as any
   }
   bimap<L2, B>(f: (l: L) => L2, g: (a: A) => B): Either<L2, B> {
@@ -122,7 +125,10 @@ export class Right<L, A>
   ap_<B, C>(this: Either<L, (a: B) => C>, fb: Either<L, B>): Either<L, C> {
     return fb.ap(this)
   }
-  chain<L2, B>(f: (a: A) => Either<L2, B>): Either<L | L2, B> {
+  chain<B>(f: (a: A) => Either<L, B>): Either<L, B> {
+    return f(this.value)
+  }
+  chain_<L2, B>(f: (a: A) => Either<L2, B>): Either<L | L2, B> {
     return f(this.value)
   }
   bimap<L2, B>(f: (l: L) => L2, g: (a: A) => B): Either<L2, B> {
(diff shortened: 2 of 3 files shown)


OliverJAsh:
I'm also experiencing some pain in this area. I've noticed that I'm having to add type annotations in many places where it would ideally be inferred. Example and workaround:

``` ts
import * as either from 'fp-ts/lib/Either';
import { Either } from 'fp-ts/lib/Either';

const typecheck = <A>(a: A) => a;

const myE = either.right<string, string>('foo');
const myFn = (s: string): Either<number, string> => either.right(s);

{
    const myE2 = myE.chain(myFn) // error: Type 'number' is not assignable to type 'string'.
}
// Workaround 1: annotate chained either
{
    const myE2 = typecheck<Either<number | string, string>>(myE)
        .chain(myFn)
}
// Workaround 2: annotate left value
{
    const myE2 = myE
        .mapLeft(error => typecheck<number | string>(error))
        .chain(myFn)
}
```

OliverJAsh:
@gcanti Do you have any ideas on this?

sledorze:
@OliverJAsh @DylanRJohnston That's not compatible with the definition of Chain.
Expect bad things to happen when you do not stick with definitions.......

sledorze:
@DylanRJohnston 
I understand all what you're saying :)
'Either<L, _>' (<= partial type application) is a Monad, if you decide to change the type of the Left side, then, it's no more a Monad (monad is parametrized over one Type param), meaning you can no more abstract over it as a Monad, that's all.
What you're proposing is a handy combinator but not the `chain` function of the `Chain` type class.

Oh I've seen your changes :)

It's interesting to note that within a language with Union types, those patterns are now available that were not possible in other FP without this feature.

We should maybe think more about all this new power and how to tame it.. (new type classes for HKT2 types?)
Maybe stealth from somewhere.. I haven't looked into it.
@gcanti have you?

DylanRJohnston:
@sledorze I understand that Monads in Haskell are over one type parameter, but there's no reason why that has the to be the case in general. The categorical theory behind it allows it. Edward Kmett talks about doing something similar in this talk: `https://www.youtube.com/watch?v=Klwkt9oJwg0`.

sledorze:
@DylanRJohnston oh sure, you can have some morphisms to make that a monad but you need to apply them beforehand, that's not the case here.

OliverJAsh:
I've been thinking about this some more. I think the issue at the heart of this is the way tagged unions work in TypeScript. Example:

``` ts
{
    interface TwitterApiError { type: 'TwitterApiError' }
    const fetchTwitter = (): Either<TwitterApiError, string> => either.right('foo');

    interface GitHubApiError { type: 'GitHubApiError' }
    const fetchGitHub = (): Either<GitHubApiError, string> => either.right('foo');

    type ParentError = TwitterApiError | GitHubApiError;
    const fetchTwitterAndGitHub = (): Either<ParentError, string> =>
        fetchTwitter()
            // error: Type '"GitHubApiError"' is not assignable to type '"TwitterApiError"'.
            .chain(fetchGitHub)
}
```

As you can see, this is a problem when you want to compose functions that return `Either`s with different types on the left hand side.

Is there any way we can make this a better experience, without breaking any FP laws?

SimonMeskens:
FP laws promise us there can only ever be one error, right? So the output type of Either.chain should be an Either with the left side being of either type.

`chain<L2, B>(f: (a: A) => Either<L2, B>): Either<L | L2, B>`

Because Either always specifically makes sure only one error propagates, this type should still be correct, without breaking any FP laws. In case of L = L2, TypeScript will reduce `ConcreteL | ConcreteL` to just ConcreteL. If we go up the abstraction, chain is only defined over B, so the Monad laws don't care how we treat the left side of an Either.

SimonMeskens:
So, basically, there's no reason for chain_, you can just define the regular chain as accumulating errors and still be correct.

OliverJAsh:
In Elm I can write:

``` elm
fetchFoo x = Err "foo"
fetchBar x = Err 1
type FetchFooAndBarError string number = FooError string | BarError number
fetchFooAndBar =
  fetchFoo "foo" |>
    Result.mapError FooError |>
    Result.andThen
      (\s -> (Result.mapError BarError (fetchBar s)))
```

`Result.andThen` (`chain`) knows that `FooError` and `BarError` are part of the `FetchFooAndBarError` tagged union, so it doesn't error.

The same thing in TypeScript with fp-ts would need manual type widening:

``` ts
const fetchFoo = (): Either<string, string> => either.of('foo');
const fetchBar = (): Either<number, string> => either.of('foo');

class FooError {
    constructor(public error: string) {}
}
class BarError {
    constructor(public error: number) {}
}
type FetchFooAndBarError = FooError | BarError;
const fetchFooAndBar = (): Either<FetchFooAndBarError, string> =>
    fetchFoo()
        .mapLeft(error => new FooError(error))
        // Manually widen type of Left generic
        .mapLeft((error): FetchFooAndBarError => error)
        .chain(() => fetchBar().mapLeft(error => new BarError(error)));
```

gcanti:
@OliverJAsh Type inference, being local, needs a hint

```ts
export const fetchFooAndBar = (): Either<FetchFooAndBarError, string> =>
  fetchFoo()
    .mapLeft<FetchFooAndBarError>(error => new FooError(error))
    .chain(() => fetchBar().mapLeft(error => new BarError(error)))
```

otherwise infers `mapLeft<FooError>`

DylanRJohnston:
@OliverJAsh, if you want to use it now I just have this module extension defined in my project.

```typescript
import { Either } from 'fp-ts/lib/Either'

declare module 'fp-ts/lib/Either' {
  export interface Left<L, A> {
    chain<B, L2>(f: (a: A) => Either<L2, B>): Either<L | L2, B>
  }

  export interface Right<L, A> {
    chain<B, L2>(f: (a: A) => Either<L2, B>): Either<L | L2, B>
  }

  export function of<R>(r: R): Either<never, R>
}

export * from 'fp-ts/lib/Either'
```

DylanRJohnston:
Closing due to lack of interest

OliverJAsh:
For anyone interested, here is an explanation of why TypeScript does not infer a union for generics:

https://stackoverflow.com/a/39905723/5932012

OliverJAsh:
I've found that using [`unionize`](https://github.com/pelotom/unionize) to represent tagged union types aka sum types helps a lot with this—you don't have to manually widen to the union types, because the record constructors return the union type. Here is an example:

``` ts
{
    type Cases<T> = {
        foo: () => T;
        bar: () => T;
    };
    const match = <T>(cases: Cases<T>) => {};

    {
        class Foo {
            readonly type = 'foo';
        }
        class Bar {
            readonly type = 'bar';
        }
        type Type = Foo | Bar;
        match(
            /*
            Error:
            Argument of type '{ foo: () => Foo; bar: () => Bar; }' is not assignable to parameter of type 'Cases<Foo>'.
                Types of property 'bar' are incompatible.
                    Type '() => Bar' is not assignable to type '() => Foo'.
                        Type 'Bar' is not assignable to type 'Foo'.
                        Types of property 'type' are incompatible.
                            Type '"bar"' is not assignable to type '"foo"'.
            */
            {
                foo: () => new Foo(),
                bar: () => new Bar(),
            },
        );
    }

    {
        const Type = unionize({
            Foo: ofType<{ fooId: string }>(),
            Bar: ofType<{ barId: string }>(),
        });
        // No error!
        match({
            foo: () => Type.Foo({ fooId: 'foo' }),
            bar: () => Type.Bar({ barId: 'bar' }),
        });

        // This works because the record constructors for the sum type return the union type, not
        // the record type:
        const foo = Type.Foo({ fooId: 'foo' });
        foo.tag; // "Foo" | "Bar"
    }
}
```

PR #141, opened 2017-06-24
Correct TaskEither fold method

OliverJAsh:


OliverJAsh:
@gcanti Updated, let's see if it passes now.

PR #155, opened 2017-07-15
add type-level dictionaries in order to reduce the number of overload…

gcanti:
…ings


SimonMeskens:
The typechecker already reports a bug with the codebase I think. It complains that State<S, A> is not conform to HKT2<L, A>, because State uses `_S` instead of `_L`. I assume that we change that in State.ts?

gcanti:
This is great, good catch!

`Store` wasn't compliant too

gcanti:
I'm experimenting random errors in `Applicative.ts` and `HKT.ts` (I say random because my local build succeeded though travis and my vscode complain). Note the last two messages

error

SimonMeskens:
Can you try something for me? In VSCode, press F1, type in "typescript", click "TypeScript: Select TypeScript Version" and click "Use Workspace Version". See if that fixes it in your VSCode. If it does, then that means Travis is using the incorrect version of TypeScript?

SimonMeskens:
I can confirm that there are weird random errors popping up.

SimonMeskens:
Sorry for the linting commits, I figured out how to run the linter now (didn't know there was one). It seemed to have done a weird nonsensical merge too. If it bothers you, I'll do a rebase at the end?

SimonMeskens:
With the current method, you can't actually assign anything to the interface HKT<U, A> it seems, which might be bad. Then again, there really is no reason to have a HKT<U, A> anywhere in the code, as it's possible to convert to a concrete at compile time, so maybe this behavior is desirable.

gcanti:
@SimonMeskens no problem. It seems that this last version doesn't check wrong HKT2 phantom types anymore though

SimonMeskens:
Yup, it currently strips out the check when compiling, but we could export the check, so that doesn't happen.

gcanti:
I mean that if I change `_L` to `_S` in `State` (as it was before https://github.com/gcanti/fp-ts/pull/155#issuecomment-315545345) it doesn't raise any error

SimonMeskens:
Fixed

gcanti:
This branch also builds with typescript@2.5.0-dev.20170719 without any change.

@SimonMeskens if it's ok for you I'm going to merge this and publish a `fp-ts@next` release

SimonMeskens:
I tried further reducing overloads, but I can't really find a way, so yeah, merge away

PR #158, opened 2017-07-18
Added integrity checking and HKTAs

SimonMeskens:
Here's the POC for HKTAs. It's a very simple replacement for URI2HKT. It's just a bit more typesafe (forces use of HKTS) and stops you from using wrongly defined URI2HKT mappings. It also means users don't have to use indexers anymore, but just plain generics, to access the URI2HKT map.

### DIFF
--- a/src/Applicative.ts
+++ b/src/Applicative.ts
@@ -1,4 +1,4 @@
-import { HKT, HKTS, HKT2S, URI2HKT, URI2HKT2 } from './HKT'
+import { HKT, HKTS, HKT2S, HKTAs, HKT2As } from './HKT'
 import { Apply, FantasyApply } from './Apply'
 import {
   getFunctorComposition,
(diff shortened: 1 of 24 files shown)


SimonMeskens:
I'm not quite sure why prettier fails btw, I'll need some help with that.

gcnew:
While reading the code, I was wondering whether there is a specific reason to use:

```ts
export type HKTS = keyof {
  [key in URI2HKT<any>[keyof URI2HKT<any>]['_URI']]: any
}
```

instead of simply:
```ts
export type HKTS = URI2HKT<any>[keyof URI2HKT<any>]['_URI']
```

gcnew:
This change looks good from my perspective.

SimonMeskens:
Yup, that simplication works, odd that I didn't think of that, I keep forgetting you can pass disjunct unions an indexer. You want to merge it in? It needs a fix for the "prettier" fail though, I'm not sure what's wrong

gcnew:
How do you run it/how is it supposed to be run? I didn't see an `npm run lint` macro.

gcanti:
you can run fix-prettier

gcnew:
Oops, my bad. I was looking at the bottom of the file and didn't see that the scrips are actually at the top.

gcanti:
In `Identity.ts` should be

```diff
declare module './HKT' {
-  interface URIHKT<A> {
+  interface URI2HKT<A> {
     Identity: Identity<A>
   }
 }
 ```

gcanti:
> It also means users don't have to use indexers anymore, but just plain generics, to access the URI2HKT map

Good idea.

@SimonMeskens I'd love to also keep that small 2-liner (...2 bugs found ;)

gcanti:
Weird, just tried moving that 2-liner to tests but it doesn't work there

SimonMeskens:
Yeah, I did the same thing and moved it to tests, but somehow that doesn't work. I think it has to do with imports and merged declarations. I'll put it in HKT.ts for now

gcanti:
@SimonMeskens to fix travis, you can run `npm run fix-prettier`

SimonMeskens:
Aha, that does seem to work, thanks, now I understand. Prettier just checks, fix-prettier to actually fix, makes sense

gcanti:
(yeah, I use vscode with prettier onsave, so usually I don't run fix-prettier)

Excellent, LGTM

SimonMeskens:
I'm going to let you handle merging. Do you want any rebases done?

PR #162, opened 2017-07-19
use the bottom never type for none, closes #160

### LINKED ISSUE #160: Consolidate option.zero and option.none
I keep getting tripped up using `option.none` forgetting that it widens the value generic type to `any`. I noticed that `option.zero` is provided to allow a generic to be passed through. I wonder if we could consider consolidating these to help ensure type safety. What do you think?

gcanti:


src/Option.ts
```diff
@@ -276,7 +276,7 @@ export function fromPredicate<A>(predicate: Predicate<A>): (a: A) => Option<A> {
 export function partitionMap<A, L, R>(f: (a: A) => Either<L, R>, fa: Option<A>): { left: Option<L>; right: Option<R> } {
   return fa.fold(
     () => ({ left: none, right: none }),
-    a => f(a).fold(l => ({ left: some(l), right: none }), a => ({ left: none, right: some(a) }))
+    a => f(a).fold(l => ({ left: some(l), right: none }), a => ({ left: none as Option<L>, right: some(a) }))
```
OliverJAsh: How come you need the type assertion here?

gcanti: because ts is using the left branch (`l => ({ left: some(l), right: none })`) to infer the types, I pushed a cleaner solution

OliverJAsh:
I'm curious of we could also use `never` in the `Left`/`Right` for the generic of the opposing value?

PR #163, opened 2017-07-20
add Task.tryCatch, closes #159

### LINKED ISSUE #159: async either.tryCatch (for promises/tasks)

gcanti:


src/Task.ts
```diff
@@ -82,4 +83,8 @@ export function empty<A>(): Task<A> {
   return never as Task<A>
 }
 
+export function fromPromise<L, A>(f: Lazy<Promise<A>>, onrejected: (reason: any) => L): Task<Either<L, A>> {
+  return new Task(() => f().then(a => right<L, A>(a), reason => left<L, A>(onrejected(reason))))
+}
+
```
OliverJAsh: It does feel slightly inconsistent to have `Task.fromPromise` and `either.tryCatch`, which are async and sync versions of the same idea, but with different naming. 🤔

gcanti: Do you prefer `tryCatchAsync`? Or `asyncTryCatch`? Or even simply `tryCatch`

OliverJAsh: Perhaps `tryCatch`, as the `async` part is implicit with the `Task` module?

So then we'll have `either.tryCatch => Either` and `task.tryCatch => Task<Either>>`?

gcanti: > Perhaps tryCatch, as the async part is implicit with the Task module?

👍

PR #167, opened 2017-07-22
add Distributive functor

gcanti:
This PR provides the notion that is categorically dual to `Traversable`.

A `Distributive` Functor is one that you can push any functor inside of.

```purescript
distribute :: (Functor f, Distributive g) => f (g a) -> g (f a)
```

Compare this with the corresponding `Traversable` notion, `sequence`.

```purescript
sequence :: (Applicative f, Traversable g) => g (f a) -> f (g a)
```

### DIFF
--- a/src/Distributive.ts
+++ b/src/Distributive.ts
@@ -1,0 +1,56 @@
+import { HKT, HKTS, HKT2S, HKTAs, HKT2As } from './HKT'
+import { Functor } from './Functor'
+
+// adapted from https://github.com/purescript/purescript-distributive
+
+/** Categorical dual of `Traversable`:
+ *
+ * - `distribute` is the dual of `sequence` - it zips an arbitrary collection of containers.
+ * - `collect` is the dual of `traverse` - it traverses an arbitrary collection of values.
+ *
+ * Laws:
+ *
+ * - `distribute = collect id`
+ * - `distribute <<< distribute = id`
+ * - `collect f = distribute <<< map f`
+ * - `map f = unwrap <<< collect (Identity <<< f)`
+ * - `map distribute <<< collect f = unwrap <<< collect (Compose <<< f)`
+ */
+export interface Distributive<F> extends Functor<F> {
+  distribute<G>(G: Functor<G>): <A>(gfa: HKT<G, HKT<F, A>>) => HKT<F, HKT<G, A>>
+}
+
+export class Ops {
+  /** A default implementation of `collect`, based on `distribute` */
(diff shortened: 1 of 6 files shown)


SimonMeskens:
Is there a reason why Distributive and Traversable don't have corresponding fantasy interfaces?

gcanti:
AFAIK `FantasyTraversable` does exist. `FantasyDistributive` seems not possible since the outer type is `HKT<G, _>`. What do you propose?

SimonMeskens:
The spec for fantasy-land Traversable is here:
https://github.com/fantasyland/fantasy-land#traversable

If I understand correctly, Distributive would just be a reversion of the arrows?

gcanti:
> AFAIK FantasyTraversable does exist

I mean that is already defined in fp-ts https://github.com/gcanti/fp-ts/blob/ee67de4b897cb9a5b87be9d39cee6751a1353251/src/Traversable.ts#L11

The signature

```purescript
distribute :: (Functor f, Distributive g) => f (g a) -> g (f a)
```

suggests that adding a method to the prototype of `Distributive g` is useless since the value at hand is a `f (g a)`

SimonMeskens:
Oh my bad, right, it's FantasyFilterable that doesn't exist, I looked at the wrong file.

I'm not sure I understand how your version of Distributive works yet, I'll play around with it a bit tomorrow and look if I can come up with a sensible Fantasy version.

gcanti:
> it's FantasyFilterable that doesn't exist

Ah you are right (and `Witherable` as well). Would you like to send a PR for them? Writing the `Fantasy*` instances for `Filterable` and `Witherable` should make clear why I think is not possible to do the same for `Distributive`: both the missing definitions involve a `HKT<F, A>` value where `F` is the "main" type parameter of the interfaces while in `Filterable` the main type parameter is `F` but the involved value is `G`-parametrised. Does it make sense?

SimonMeskens:
Ah, cool, yeah, I was finishing up one of the Immutable collections yesterday and ran into this part of the library. I'll have a look if I can solve the conundrum.

SimonMeskens:
You are correct.

I looked up all the theory behind it (and why a monadic operation such as distribute or traverse only needs respectively a Functor or an Applicative) and I get it now. Without extension methods in TypeScript, you cannot attach the Distributive to anything. You can always write a Fantasy type of a dual, unless that dual forces the focus type into the output slot. Makes total sense. There's no hidden way to express it that would give rise a Fantasy type.

PR #174, opened 2017-07-24
FantasyFilterable and FantasyWitherable

SimonMeskens:
As promised. Array doesn't have Fantasy* implementations and Either needs a monoid to be available to do them, so only Option.ts is changed to reflect the new interfaces.


gcanti:
Thanks!

PR #177, opened 2017-07-25
add generic function

amir-arad:
I've added a generic function with variable amount of arguments. It's something I need in my work for function chaining, adaptors and such.

Issues:
1. type `NumberToString` should move to `typelevel-ts` project IMO
2. I couldn't find a way to get rid of the first generic param `I`. It looks like it's redundant, as the size of the tuple is already in the `A` param. 
I've tried these things:
```ts
type TupleLength<T, I = 0> = {
    'true': TupleLength<T, Increment[I]>
    'false': I
}[ObjectHasKey<T, I>]
type GFunction2<A, R=void> = GFunction<TupleLength<A>, A, number>;
type GFunction3<A, R=void, I=0> = GFunction<TupleLength<A, I>, A, number>;
```
and also
```ts
export type GFunction<A extends Array<any>, R=void> =
    If<ObjectHasKey<A, '1'>,
        Function2<A[0], A[1], R>,
        If<ObjectHasKey<A, '0'>,
            Function1<A[0], R>,
            Lazy<R>>>
```
would welcome any ideas...

### DIFF
--- a/src/function.ts
+++ b/src/function.ts
@@ -228,3 +228,18 @@ export function toString(x: any): string {
 export function tuple<A, B>(a: A, b: B): [A, B] {
   return [a, b]
 }
+
+export type NumberToString = ['0', '1', '2', '3', '4', '5', '6', '7', '8', '9', '10'];
+export type GFunction<I extends number, A, R=void> = {
+    0: Lazy<R>;
+    1: Function1<A[0], R>
+    2: Function2<A[0], A[1], R>
+    3: Function3<A[0], A[1], A[2], R>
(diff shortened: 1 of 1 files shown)


SimonMeskens:
I did some initial tests and it's perfectly possible to get rid of the first generic parameter. The code is kinda complex though, it seems. I'll report back if I find a solution I like.

gcanti:
Interesting topic, thanks for this PR @amir-arad, though it seems more related to typelevel-ts than fp-ts, please open an issue there if you find a good solution

PR #178, opened 2017-07-27
Doc for `fromNullable`

davetayls:
Not sure if this is the right style of description, but I wanted to start from somewhere to help me get the right direction so that i can be useful 😄

### DIFF
--- a/src/Option.ts
+++ b/src/Option.ts
@@ -220,7 +220,7 @@ export function fold<A, B>(n: Lazy<B>, s: (a: A) => B, fa: Option<A>): B {
   return fa.fold(n, s)
 }
 
-/** 
+/**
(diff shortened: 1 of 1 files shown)


davetayls:
@gcanti @SimonMeskens ... any thoughts on this so i can go ahead and add further docs to what I understand?

gcanti:
As a possible source of documentation: many modules are porting of PureScript or Scala libs. For example the `Option` module is adapted from

- [Scala's Option](https://www.scala-lang.org/api/current/scala/Option.html)
- [purescript-maybe](https://github.com/purescript/purescript-maybe/blob/435566d37188a574276b9b98737bfcb4d5b89a9f/src/Data/Maybe.purs)

davetayls:
Brill, that's really helpful. I can imitate that to add more info into the doc

Issue #183, opened 2017-07-31
contains, isDefined, isEmpty, ... for Option


alexandervanhecke:
Hi,
I'm integrating fp-ts into a project, I noticed some member functions I would expect to be present in the Some / None classes were missing, like contains, isDefined, isEmpty (and possibly others).  

I realize it's possible to get this functionality using other functions, like e.g. isNone, isSome, etc.  

Is this a design choice (ie using global functions in which you pass Option instances instead of member functions on Some / None)?  

I'm quite willing to create a pull request for these methods if you want me to. 

wkr,

gcanti:
> Is this a design choice

Yes, namely the members of a type-class instance, for example `{ map, of, ap, chain }` for `Monad`. In addition I've just added the methods I use the most (in particular `fold` and `getOrElse`), feel free to add some useful member function though.

Another option, in case you judge a method just a nice-to-have, is augmenting the base module in your own code

```ts
import * as option from 'fp-ts/lib/Option'

declare module 'fp-ts/lib/Option' {
  interface None<A> {
    someMethod(a: A): boolean
  }
  interface Some<A> {
    someMethod(a: A): boolean
  }
}

option.None.prototype.someMethod = function<A>(a: A) {
  ...
}

option.Some.prototype.someMethod = function<A>(a: A) {
  ...
}
```

alexandervanhecke:
Thanks,
understood.  I just created a pull request for this [#184](https://github.com/gcanti/fp-ts/pull/184), added methods and unit tests for : 
* contains
* isEmpty
* isDefined
* nonEmpty
* forEach
* exists
* orEmpty

PR #184, opened 2017-07-31
add the following member methods in Some / None classes :

alexandervanhecke:
- contains
- isEmpty
- isDefined
- nonEmpty
- forEach
- exists
- orEmpty

see also issue https://github.com/gcanti/fp-ts/issues/183

### DIFF
--- a/src/Option.ts
+++ b/src/Option.ts
@@ -104,6 +104,27 @@ export class None<A>
   toString() {
     return 'none'
   }
+  contains(setoid: Setoid<A>, a: A): boolean {
+    return false
+  }
+  isEmpty(): boolean {
+    return true
+  }
+  isDefined(): boolean {
+    return !this.isEmpty()
+  }
+  nonEmpty(): boolean {
+    return this.isDefined()
+  }
+  forEach<B>(f: (a: A) => B) {
+    // empty
+  }
+  exists(p: (a: A) => Boolean): boolean {
+    return false
+  }
+  orEmpty(monoid: Monoid<A>): A {
+    return monoid.empty()
+  }
 }
 
 export const none = None.value
@@ -203,6 +224,27 @@ export class Some<A>
   }
   toString() {
     return `some(${toString(this.value)})`
+  }
+  contains(setoid: Setoid<A>, a: A): boolean {
+    return setoid.equals(this.value, a)
+  }
+  isEmpty(): boolean {
+    return false
+  }
+  isDefined(): boolean {
+    return !this.isEmpty()
+  }
+  nonEmpty(): boolean {
+    return this.isDefined()
+  }
+  forEach<B>(f: (a: A) => B) {
+    f(this.value)
+  }
+  exists(p: (a: A) => Boolean) {
+    return !this.isEmpty() && p(this.value)
+  }
+  orEmpty(monoid: Monoid<A>): A {
+    return this.value
   }
 }
 
--- a/src/Option.ts
+++ b/src/Option.ts
@@ -107,23 +107,14 @@ export class None<A>
   contains(setoid: Setoid<A>, a: A): boolean {
     return false
   }
-  isEmpty(): boolean {
+  isNone(): boolean {
     return true
   }
-  isDefined(): boolean {
-    return !this.isEmpty()
-  }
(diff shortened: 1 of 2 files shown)


alexandervanhecke:
@SimonMeskens no other reason than I got my inspiration from the scala Option type which defines these methods and aliases.  But it's definitely debatable.

wkr,

SimonMeskens:
I figured that was the case, just wanted to check. Probably better if you didn't waste a full closure on it though, maybe just:

```TypeScript
nonEmpty: () => boolean = this.isDefined
```

It's not a big deal, since this is not performance critical code and the latest V8 optimizes specially for this case, but it seems a bit wasteful to create a full closure just for an alias. Added benefit is that without the closure, nonEmpty is strictly equal to isDefined.

alexandervanhecke:
@SimonMeskens thanks for your feedback, I wasn't even aware this could have a performance impact.

src/Option.ts
```diff
@@ -204,6 +225,27 @@ export class Some<A>
   toString() {
     return `some(${toString(this.value)})`
   }
+  contains(setoid: Setoid<A>, a: A): boolean {
```
SimonMeskens: Would it make more sense to make `setoid` optional and fall back to `this.value.equals(a)`?

SimonMeskens:
> thanks for your feedback, I wasn't even aware this could have a performance impact.

If you wrap a function call inside another function, you always create a new function with a new callstack. Arrow functions don't have a full context and are a bit lighter, but just using equality like I suggested doesn't have any overhead.

As I said, it doesn't matter much for performance, but I think it's just good habit not to create closures when you don't need to. In older versions of V8, creating a new closure is about 2x slower, in newer versions, it's almost as fast in this specific case, because V8 will inline the call for you. Unless you'll be calling nonEmpty a couple thousand times per second, you won't even notice a linear increase like that.

src/Option.ts
alexandervanhecke: Good point, I've tried to mirror the signature of the `equals` method, but an optional `setoid` would work as well of course.  Let me know what you prefer, and I'll update if needed.

SimonMeskens: Ah yes, I see how it mirrors `equals`. I don't like how that breaks compatibility with most other libraries. Is there a reason for that @gcanti?

gcanti:
@alexandervanhecke `.isEmpty`, `.isDefined` and `.nonEmpty` do the same thing, can we keep just one? Or maybe two of them, with the following renaming (for consistency with the top-level functions)

`isEmpty` -> `isNone`
`isDefined` -> `isSome`

> Is there a reason for that?

@SimonMeskens yes, you have to provide an explicit `Setoid` instance, corresponding to the type constraint `Eq a =>` in the following signature

```purescript
contains :: Eq a => Maybe a -> a -> Boolean
//           ^       ^
//        setoid    this
```

This encoding, i.e. translating a type constraint to an additional required dictionary, usually in first position, is consistent across the library.

> `exists(p: (a: A) => Boolean): boolean`

I guess should be **b**oolean

> `.forEach<B>(f: (a: A) => B): void`

As is this signature means that some side effect will be executed, should be removed.

`.orEmpty` seems too specialised and a bit weird, it requires a `Monoid` instance, but if you have such an instance in scope can't you just call `.getOrElse(monoid.empty())`?

SimonMeskens:
Yes, I do understand the consistency, I just don't like the fact that fp-ts is incompatible with fantasy-land in so many places, and thus incompatible with libraries that consume fantasy types. Maybe we should write an overload for every compatibility break?

gcanti:
I see, we could discuss this topic in a new issue, though being compatible with fantasy-land and static-land is not a goal and AFAIK there are already many unavoidable incompatibilities

alexandervanhecke:
@gcanti thanks for your feedback, I've removed these functions and did the renames as you suggested.  

as for the `isSome: () => boolean = () => true` syntax, I ended up not using it because it seemed to trigger errors in some of the other mocha tests

let me know if there are other things you want me to change

gcanti:
Thanks @alexandervanhecke, let me merge some other PRs and then I'll publish a new release with this

PR #199, opened 2017-08-12
Fantas, Eel, and Specification

gcanti:
This PR adds a bunch of files, one for each post contained in this nice [blog series](http://www.tomharding.me) by @i-am-tom (thanks Tom!) on Fantasy Land 

The idea is that people could read a post and then find out how the examples can be translated to "idiomatic" fp-ts.

Comments, corrections and suggestions are welcomed


SimonMeskens:
I think this is a really good idea

SimonMeskens:
Could we ask Tom if he could open source the articles, so we can include them on the documentation website eventually?

i-am-tom:
@SimonMeskens if they aren't all ready, let me know what I need to do to make it happen :)

SimonMeskens:
@i-am-tom 

If you leave a reply here saying: "All my blog articles are licensed MIT" (or whatever license you prefer), I think that takes care of it quite handily.

If you want others to also have the benefit of being able to use them easily, you could edit all your articles (or just the blog main page) and add a line that says: this article is licensed MIT.

If you really want to go above and beyond (for some reason), what I've also seen is people keeping a github repository with their articles and a license.

You're a champ btw :). Love your blog

i-am-tom:
In which case, all my blog articles are licensed MIT :+1: there's a [LICENSE in the site's repo](https://github.com/i-am-tom/i-am-tom.github.io/blob/master/LICENSE), but I'll add a footer or something to make it a bit clearer to people who get to it via the usual channels :) Glad you like it - I hope I've been useful in some way! <3

SimonMeskens:
Nice! Thanks ❤️

PR #201, opened 2017-08-15
Fr/migrate to curried

raveclassic:
I'm opening this PR as a POC to track #200 

Migration list:
- [x] Alt 
- [x] Alternative
- [ ] Applicative
- [ ] Apply
- [ ] Array
- [ ] Bifunctor
- [ ] Chain
- [ ] ChainRec
- [x] Comonad
- [ ] Console
- [ ] Const
- [ ] Contravariant
- [ ] Either
- [ ] EitherT
- [ ] Exception
- [ ] Extend
- [x] Field
- [ ] Filterable
- [ ] Foldable
- [ ] Free
- [ ] Functor
- [ ] HKT
- [ ] IO
- [ ] Identity
- [ ] IxIO
- [ ] IxMonad
- [x] Monad
- [x] Monoid
- [ ] Monoidal
- [ ] NaturalTransformation
- [ ] NonEmptyArray
- [ ] Option
- [ ] OptionT
- [x] Ord
- [x] Ordering
- [ ] Pair
- [x] Plus
- [ ] Profunctor
- [x] Random
- [ ] Reader
- [ ] ReaderT
- [x] Ring
- [x] Semigroup
- [x] Semigroupoid
- [x] Semiring
- [x] Setoid
- [ ] State
- [ ] StateT
- [ ] Store
- [ ] StrMap
- [ ] Task
- [ ] These
- [ ] Trace
- [ ] Traversable
- [ ] Tuple
- [ ] Unfoldable
- [ ] Validation
- [ ] Witherable
- [ ] Writer
- [ ] function

Issue #200, opened 2017-08-14
FR: Provide curried versions of functions


raveclassic:
Hi!

I have just found myself currying by hand yet another function like `insert/remove` from `StrMap` in `fp-ts` and `Lens#modify` in `monocle-ts` to use it with `compose` and `pipe`. 

Here's the question therefore - are you going to include curried versions of these helpers? Or maybe replace current with curried and export uncurried with a postfix like it's done in `sanctuary`: 
```
export const remove_ = <A>(k: string, d: StrMap<A>): StrMap<A> => { ... }
export const remove: (k: string) => <A>(d: StrMap<A>) => StrMap<A> = curry(remove_);
```

gcanti:
My policy so far is simple

- do not curry...
- ...unless type inference and/or correctness can be improved

Example

```ts
import { Applicative } from '../src/Applicative'
import { HKT } from '../src/HKT'
import { option, some } from '../src/Option'
import { right } from '../src/Either'

// curried version
declare function traverse1<F>(F: Applicative<F>): <A, B>(f: (a: A) => HKT<F, B>, ta: Array<A>) => HKT<F, Array<B>>

// x: HKT<"Option", number[]>
export const x1 = traverse1(option)(a => some(a), [1, 2, 3])
// export const x2 = traverse1(option)(a => right(a), [1, 2, 3]) // error :)

// uncurried version
declare function traverse2<F, A, B>(F: Applicative<F>, f: (a: A) => HKT<F, B>, ta: Array<A>): HKT<F, Array<B>>

// x3: HKT<"Option" | "Either", number[]>
export const x3 = traverse2(option, a => right(a), [1, 2, 3]) // no error :(
```

I'm definitely open to suggestions in order to change the policy (in general I would avoid API duplications though)

raveclassic:
Well, at the beginning I was trying to change an object through lens multiple times and noticed that all modifying functions on every structure (Array, StrMap etc.) are uncurried as well as `Lens#modify` so at first the code looked like:
```typescript
type Entity = {
  id: string,
  foo: string
};
type State = {
  ids: string[],
  entities: StrMap<Entity>
};
const ids = Lens.fromProp<State, 'ids'>('ids');
const entities = Lens.fromProp<State, 'entities'>('entities');

//signature is aligned with redux reducers (that's why state goes first)
const removeFromState = (state: State, id: string): State => compose(
  state => ids.modify(ids => filter(id_ => id_ !== id, ids), state),
[...]

PR #202, opened 2017-08-16
Fr/migrate to curried

raveclassic:
This PR is a copy of #201 which tracks #200 

Migration list:
- [x] Alt 
- [x] Alternative
- [x] Applicative
  - [ ] getApplicativeComposition
- [x] Apply
  - [ ] ap
- [x] Array
  - ~fold~
  - ~slice~
- [x] Bifunctor
  - [ ] bimap (partial?)
- [x] Chain
  - [ ] chain
- [x] ChainRec
  - [ ] chainRec
  - [ ] tailRec
- [x] Comonad
- [x] Console
- [x] Const
- [x] Contravariant
  - [x] tests for Ops#lift
- [x] Either
  - [ ] fromPredicate
- [x] EitherT
- [x] Exception
- [x] Extend
  - [ ] extend
- [x] Field
- [x] Filterable
  - [ ] partitionMap
- [x] Foldable
  - [ ] reduce
  - [ ] getFoldableComposition
  - [ ] foldMap
  - [ ] foldr
  - [ ] intercalate
  - [ ] Ops#traverse_
  - [ ] Ops#sequence_
- [x] Free
- [x] Functor
  - [ ] map
  - [ ] Ops#flap
- [x] HKT
- [x] IO
- [x] Identity
- [x] IxIO
- [x] IxMonad
  - [ ] ichain
- [x] Monad
- [x] Monoid
- [x] Monoidal
- [x] NaturalTransformation
- [x] NonEmptyArray
- [x] Option
- [x] OptionT
- [x] Ord
- [x] Ordering
- [x] Pair
- [x] Plus
- [x] Profunctor
  - [ ] promap
- [x] Random
- [x] Reader
- [x] ReaderT
- [x] Ring
- [x] Semigroup
- [x] Semigroupoid
- [x] Semiring
- [x] Setoid
- [x] State
- [x] StateT
- [x] Store
- [x] StrMap
  - [ ] mapWithKey
- [x] Task
- [x] These
  - [ ] concat
- [x] Trace
  - ~trace~
- [x] Traversable
  - [ ] traverse
  - [ ] sequence
- [x] Tuple
  - [ ] getSetoid 
  - [ ] getOrd
  - [ ] getSemigroup
- [x] Unfoldable
  - [ ] replicateA
  - [ ] singleton
- [x] Validation
  - [ ] fromPredicate 
- [x] Witherable
  - [ ] Ops#wither
  - [ ] Ops#wilted
  - [ ] Ops#withered
- [x] Writer
- [ ] function
  - ~tuple~
  - [ ] or
  - [ ] and
  - [ ] on

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,3 @@
+[*]
+indent_style = space
+indent_size = 2
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@
 node_modules
 lib
 dev
+.idea
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,4 +1,7 @@
 {
-    "typescript.tsdk": "./node_modules/typescript/lib",
-    "prettier.printWidth": 120
+  "typescript.tsdk": "./node_modules/typescript/lib",
+  "prettier.printWidth": 120,
+  "prettier.semi": false,
+  "prettier.singleQuote": true,
+  "tslint.enable": true
 }
--- a/examples/ArrayOption.ts
+++ b/examples/ArrayOption.ts
@@ -17,15 +17,11 @@ export const URI = 'ArrayOption'
 export type URI = typeof URI
 
 export class ArrayOption<A> implements FantasyMonad<URI, A> {
-  static of = of
   readonly _A: A
   readonly _URI = URI
-  constructor(public readonly value: Array<Option<A>>) {}
+  constructor(readonly value: Array<Option<A>>) {}
   map<B>(f: (a: A) => B): ArrayOption<B> {
     return new ArrayOption(optionTArray.map(f, this.value))
-  }
-  of<B>(b: B): ArrayOption<B> {
-    return of(b)
   }
   ap<B>(fab: ArrayOption<(a: A) => B>): ArrayOption<B> {
     return new ArrayOption(optionTArray.ap(fab.value, this.value))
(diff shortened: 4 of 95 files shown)


gcanti:
@raveclassic With the current order of arguments I think we should keep Functor.map uncurried otherwise type inference will suffer

```ts
export interface Functor<F> {
  readonly URI: F
  map: <A, B>(f: (a: A) => B, fa: HKT<F, A>) => HKT<F, B>
}
```

the `fa` argument here helps ts to infer the type `A`, otherwise you have to always add type annotations to the argument `f`.

If you want to curry `map`, we should swap the arguments (as in Scala)

```ts
export interface Functor<F> {
  readonly URI: F
  map: <A>(fa: HKT<F, A>) => <B>(f: (a: A) => B) => HKT<F, B>
}
```

gcanti:
The problem is that in a real world scenario you likely want to apply the `f` argument. Now if we swap the arguments, you can't still do that, hence we end up with a breaking change with no apparent gain

raveclassic:
@gcanti Agree. What about leaving such cases as is and currying everything else where possible?

gcanti:
Yep, we can carry on for now and come back to this topic in a second phase

raveclassic:
@gcanti Ok I'll try to mark all problematic places

gcanti:
Note (to myself): do not curry `function.tuple` :)

raveclassic:
> Note (to myself): do not curry function.tuple :)

Just noticed there are actually two `tuple` functions - uncurried in `function.ts` and curried in `Monoidal.ts` :)

gcanti:
The `function.tuple` is an handy helper because ts infers `[1, 'a']` as `(string | number)[]
` instead of `[number, string]` (I should add a comment)

p.s. working on the `Array` module

gcanti:
@raveclassic Working on Chain, ChainRec, Either, EitherT

gcanti:
Working on Exception, Extend, Filterable, Foldable, Functor

raveclassic:
@gcanti Noticed a cast `eq(alt(none as Option<number>)(some(2)), some(2))`. We need to add some overloads to Option's `alt`:
```typescript
// overload here to support 'none' as first argument
export function alt(fx: Option<never>): <A>(fy: Option<A>) => Option<A>
export function alt<A>(fx: Option<A>): (fy: Option<A>) => Option<A>
export function alt<A>(fx: Option<A>): (fy: Option<A>) => Option<A> {
  return fy => fx.alt(fy)
}
```

Fixed it here

gcanti:
Ah.. I thought they were vestigial. Maybe there's no need for special casing though, usually you get an `Option<A>` rather than a `None`:

```ts
import { alt, none, some, Option } from '../src/Option'

function f(): Option<number> {
  return none
}

alt(f())(some(1))
// or even
eq(alt<number>(none)(some(2)), some(2))
```

raveclassic:
Ok I think we're done here!

gcanti:
@raveclassic I left the Trace module uncurried intentionally, that's because it's supposed to be a quick and dirty way to debug the code, what's your opinion?

raveclassic:
Ok I reverted that change

gcanti:
For what concerns the left bits, this is my proposal

Legend:

- TCR: type constraint rule: all type constraints are grouped together
- TI: currying will cause bad type inference
- TU: is not curried because of the typical usage

Functions

- `Applicative.getApplicativeComposition` Won't fix: TCR
- `Apply.ap` Won't fix: TI
- `Bifunctor.bimap` Won't fix: TU
- `Chain.chain` Won't fix: TI
- `ChainRec.chainRec` Won't fix: TI
- `ChainRec.tailRec` Won't fix: TI
- `Either.fromPredicate` Won't fix: TU
- `Extend.extend` Won't fix: TI
- `Filterable.partitionMap` Won't fix: TI
- `Foldable.reduce` Won't fix: TI
- `Foldable.getFoldableComposition` Won't fix: TCR
- `Foldable.foldMap` Won't fix: TCR
- `Foldable.foldr` **will fix**
- `Foldable.intercalate` Won't fix: TCR
- `Foldable.traverse_` Won't fix: TCR
- `Foldable.sequence_` Won't fix: TCR
- `Functor.map` Won't fix: TI
- `Functor.flap` *is already curried*
- `IxMonad.ichain` Won't fix: TI
- `Profunctor.promap` **will fix** (only the `fbc` argument)
- `StrMap.mapWithKey` Won't fix: TU
- `These.concat` Won't fix: TCR
- `Traversable.traverse` Won't fix: TI
- `Traversable.sequence` Won't fix: TCR
- `Tuple.getSetoid` Won't fix: TCR
- `Tuple.getOrd` Won't fix: TCR
- `Tuple.getSemigroup` Won't fix: TCR
- `Unfoldable.replicateA` Won't fix: TCR
- `Unfoldable.singleton` **will fix**
- `Validation.fromPredicate` Won't fix: TU
- `Witherable.wither` Won't fix: TCR + TI
- `Witherable.wilted` Won't fix: TCR
- `Witherable.withered` Won't fix: TCR

Let me know if you agree or want to change something

PR #205, opened 2017-08-17
FR: migrate to curried // Array

gcanti:
@raveclassic not sure about `fold` and `slice` though, I feel like they shouldn't be curried

### DIFF
--- a/src/Array.ts
+++ b/src/Array.ts
@@ -3,6 +3,7 @@ import { Monoid } from './Monoid'
 import { Applicative } from './Applicative'
 import { Monad } from './Monad'
 import { Foldable } from './Foldable'
+import { Unfoldable } from './Unfoldable'
 import { Traversable } from './Traversable'
 import { Alternative } from './Alternative'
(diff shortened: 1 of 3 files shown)


raveclassic:
@gcanti Why not?

gcanti:
@raveclassic because of their typical usage

```ts
/** Break an array into its first element and remaining elements */
export const fold = <A, B>(nil: Lazy<B>, cons: (head: A, tail: Array<A>) => B, as: Array<A>): B =>
  as.length === 0 ? nil() : cons(as[0], as.slice(1))

/** Extract a subarray by a start and end index */
export const slice = (start: number, end: number) => <A>(as: Array<A>): Array<A> => as.slice(start, end)  
```

what's the point of currying `fold` (nil, cons arguments) or `slice` (start, end arguments)? I mean, I never had a use case for partial application here

The pairs `(nil, cons)` and `(start, end)` feel like tangled arguments

raveclassic:
It's ok then, I don't disagree

PR #210, opened 2017-08-18
Bifunctor, Free, IO, Identity, IxMonad, IxIO, Monoidal

gcanti:


gcanti:
Next PR: NonEmptyArray, Option, OptionT, Pair, Profunctor

PR #212, opened 2017-08-18
Reader, State, StateT, Store, StrMap (some changes), Task, These, Trace, Traversable, Tuple, Unfoldable, Validation, Writer

gcanti:


gcanti:
@raveclassic I think I'm done with this

PR #213, opened 2017-08-19
add Moore machine

gcanti:
This is not meant to be merged, it's just as exercise, a translation of this tweet by drboolean https://twitter.com/drboolean/status/898634041129279488

### DIFF
--- a/examples/Moore.ts
+++ b/examples/Moore.ts
@@ -1,0 +1,91 @@
+import { Comonad } from '../src/Comonad'
+import { Profunctor } from '../src/Profunctor'
+import { compose } from '../src/function'
+import { Monad } from '../src/Monad'
+
+// Adapted from https://hackage.haskell.org/package/machines-0.6.3/docs/src/Data-Machine-Moore.html
+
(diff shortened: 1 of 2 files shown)


SimonMeskens:
Should we add this to docs? Looks cool.

gcanti:
Ah sure, or maybe in the examples folder.

As a further interesting exercise, I wonder if is possible to constraint the transitions to make sure statically that you can login twice before you logout

```ts
.step({ type: 'didLogIn', user: { name: 'bob' } })
.step({ type: 'didLogIn', user: { name: 'jan' } }) // <= should raise an error
.step({ type: 'didLogOut' })
```

Perhaps something along the lines of indexed monads https://github.com/gcanti/fp-ts/blob/016b40de5d562a6f0871d648ff17f81b4144d9d7/examples/ixIO.ts

SimonMeskens:
I'd probably just model it as capabilities using Either somehow. I haven't worked with indexed monads though, that looks like a cool solution indeed.

PR #216, opened 2017-08-23
FR: migrate to curried // function

gcanti:
@raveclassic apparently I forgot this module

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "fp-ts",
-  "version": "0.5.0",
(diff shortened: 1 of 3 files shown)


raveclassic:
Oh, I though it was left untouched intentionally as well.

raveclassic:
The list in #202 reflects what's left

gcanti:
> The list in #202 reflects what's left

This is super useful, I didn't noticed! Let me scan that list...

PR #222, opened 2017-09-07
add getRecordSemigroup and getRecordMonoid

gcanti:


src/Monoid.ts
```diff
@@ -1,4 +1,10 @@
-import { Semigroup, getProductSemigroup, getDualSemigroup, fold as foldSemigroup } from './Semigroup'
+import {
+  Semigroup,
+  getProductSemigroup,
+  getDualSemigroup,
+  fold as foldSemigroup,
```
raveclassic: Just thinking out loud - what do you think about moving all such "global" functions here and in similar modules to TypeScript namespaces? 
There're plenty of functions with the same name scattered across different modules and it's a big deal to manually name them when importing with `as`.
A simple solution would be moving these functions to a namespace with the name of a module and export it instead of functions. Still autoimport/autocomplete will catch names of the functions as well as the name of namespace. Here's a basic example:

```typescript
//Option.ts
export class Option<A> {
}
export namespace Option {
  export declare function map<A, B>(fa: Option<A>, mapper: (a: A) => B): Option<B>
}
```
Such functions then could be referenced as
```typescript
import { Option } from 'option';
//Option.map()
//or
import map = Option.map; //typescript namespace import (can be autoimported)
//map()
```

gcanti: I'm not acquainted with namespaces, is there a difference from

```ts
import * as option from 'fp-ts/lib/Option'

option.map
// or
const map = option.map
```

raveclassic: Note quite the same, it'd actually be 
```
import * as option from 'fp-ts/lib/Option';
const Option = option.Option; //namespace as well as class 
//or
import { Option } from 'fp-ts/lib/Option'; //also namespace as well as class
const map = Option.map;
//or
import map = Option.map; //it's just an old syntax, but it's absolutely valid
```

The difference is that you don't need to manually write `import * as ... from ...`. Another benefit is tree-shaking optimization in webpack 2 which is possible only with named imports. Though I'm not exactly sure whether it's possible with namespaces.

Issue #225, opened 2017-09-14
Nice function to get the Monad value


afharo:
I have this use case:
```typescript
import {Option, some, none} from "fp-ts/lib/Option";
interface TestObj {
    a: number,
    b: Option<number>
}

const pairs: Array<TestObj> = [
  {a: 1, b: some(1)}, 
  {a: 1, b: none}, 
];

function add(a: number, b: number) {
   return a+b;
}

// I want to add only the ones when b is some, just discard the rest
pairs
   .filter((p) => p.b.isSome())
   .map((p) => add(p.a, p.b.getOrElse(() => 0));
```
`getOrElse` function will never be hit, but I need it to be there if I want to get the value from b out of the Option. I know it exists b._A but the underscore kind of tells it's an internal value and shouldn't be used, doesn't it?
Is there any other way to solve this or can it be added a public function/property to get the value?

gcanti:
> I know it exists b._A

`_A` is a "phantom type", i.e. a field that doesn't contain any value (it's there for type checking purposes, also it helps to implement HKTs)

You can use `Option#fold` I guess

```ts
pairs.reduce((acc, obj) => acc + obj.b.fold(() => 0, n => n + obj.a), 0)
```

raveclassic:
or even shorter:
```
pairs.reduce((acc, obj) => acc + obj.b.isSome() ? obj.a : 0, 0)
```

raveclassic:
@afharo I think it's the only way because `filter` does not tell type-system that all filtered values are now `Some<number>`s, they are still `Option<number>`. So if you you need to add `a` to `b` you still need to check wether `b` is a `Just` when extracting its actual value. You can do it only with `getOrElse`.

afharo:
Yeah, I can use those, but, at the end either `.fold` or `.getOrElse` is all about providing a value when I already know that value is enforced to exist, making the code cumbersome and providing a branch that will never be covered.
The adding example was a very simple one to illustrate the example, but there are some other use cases like (if all should have a some() element in b)

```typescript
// If at least 1 is none
if (pairs.find((p) => p.b.isNone()) {
   return doSomething()
} else {
  // All values in here are some, fold or getOrElse are redundant.
  return doSomethingUsingB()
}
```

Another use case where `fold` or `getOrElse` are not the best option:
```typescript
const pairs: Array<TestObj> = [
  {a: 1, b: some({c: "test", d:3})}, 
  {a: 1, b: none}, 
];

[...]

PR #228, opened 2017-09-20
Add `toUndefined` to the Option type

vegansk:
as it discussed here: #225

### DIFF
--- a/src/Option.ts
+++ b/src/Option.ts
@@ -89,6 +89,9 @@ export class None<A>
   toNullable(): A | null {
     return null
   }
+  toUndefined(): A | undefined {
+    return undefined
+  }
   inspect() {
     return this.toString()
   }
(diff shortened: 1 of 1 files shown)


gcanti:
@vegansk btw does fp-ts play well with React?

raveclassic:
@gcanti Perfectly :)

vegansk:
@gcanti , yep. The only really missing thing is the do notation, but how we can have it without ts support in sweetjs, I don't know :-)

raveclassic:
@vegansk There [is a way](https://github.com/fluture-js/Fluture#do) but with a loss of type safety, because TS type system doesn't play well with generators. Actually is does not at all =(

vegansk:
So let's wait when macros will arrived. I already have an experience of adding do notation to the language that doesn't support it using macros :-) https://github.com/vegansk/nimfp/blob/master/tests/fp/test_forcomp.nim#L48

PR #232, opened 2017-09-24
These: add getMonad

gcanti:


SimonMeskens:
Nice, this'll come in handy :)

PR #234, opened 2017-09-29
Extra array functions

chasent:
Thank you for this very helpful library!
I have added some functions I thought would have been useful when I first picked it up.

### DIFF
--- a/src/Array.ts
+++ b/src/Array.ts
@@ -8,7 +8,7 @@ import { Traversable } from './Traversable'
 import { Alternative } from './Alternative'
 import { Plus } from './Plus'
 import { liftA2 } from './Apply'
-import { Option } from './Option'
+import { Option, fromNullable } from './Option'
 import * as option from './Option'
 import { Ord, toNativeComparator } from './Ord'
 import { Extend } from './Extend'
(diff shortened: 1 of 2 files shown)


gcanti:
Thanks @chasent, `flatMap` is already there (is named `chain` though)

raveclassic:
@gcanti Awesome! Thanks!
Do you think it would be better to use native `find` and `fromNullable` for `find` (`findFirst`) and `for cycle`  for `findLast` for the sake of performance?

src/Array.ts
```diff
@@ -42,6 +42,13 @@ export const concat = <A>(x: Array<A>) => (y: Array<A>): Array<A> => x.concat(y)
 
 export const map = <A, B>(f: (a: A) => B, fa: Array<A>): Array<B> => fa.map(f)
 
+/**
+ * Apply a function that returns an array to each element in an array,
+ * flattening the results into a new array.
+ */
+export const flatMap = <A, B>(f: (a: A) => Array<B>) => (fa: Array<A>): Array<B> =>
```
raveclassic: There is already a `chain` function

src/Array.ts
```diff
@@ -182,6 +189,12 @@ export const findIndex = <A>(predicate: Predicate<A>) => (as: Array<A>): Option<
   return option.none
 }
 
+/** Find the first element which satisfies a predicate function */
+export const findFirst = <A>(predicate: Predicate<A>) => (as: Array<A>): Option<A> => head(filter(predicate)(as))
```
raveclassic: `findFirst` can be replaced with more efficient native `find` wrapped with `fromNullable`

src/Array.ts
```diff
@@ -182,6 +189,12 @@ export const findIndex = <A>(predicate: Predicate<A>) => (as: Array<A>): Option<
...
+
+/** Find the last element which satisfies a predicate function */
+export const findLast = <A>(predicate: Predicate<A>) => (as: Array<A>): Option<A> => last(filter(predicate)(as))
```
raveclassic: `findLast` can be replaced with more efficient reversed `for`-cycle also wrapped with `fromNullable`

chasent:
Welp, I feel stupid for not noticing `chain.flatten(array)`!
I've addressed the feedback given.

raveclassic:
@chasent Great, thanks!

PR #240, opened 2017-10-02
Either: pass value to `getOrElse`

jiayihu:
This PR changes `getOrElse` signature and implementation of `Either`, allowing the callback to receive the `Left` value but nevertheless keeping backward compatibility.

Note: this will diverge from Scala typings for `Either.getOrElse` but it's useful for behaviours depending on the `Left` value

### DIFF
--- a/src/Either.ts
+++ b/src/Either.ts
@@ -71,8 +71,8 @@ export class Left<L, A>
   fold<B>(left: (l: L) => B, right: (a: A) => B): B {
     return left(this.value)
   }
-  getOrElse(f: Lazy<A>): A {
-    return f()
+  getOrElse(f: (l: L) => A): A {
+    return f(this.value)
   }
   equals(S: Setoid<A>): (fy: Either<L, A>) => boolean {
     return fy => fy.fold(constTrue, constFalse)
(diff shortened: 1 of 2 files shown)


vegansk:
The `Either` type is right biased, so `getOrElse` parameter should be `Lazy<A> => A`. See:
https://www.scala-lang.org/api/current/scala/util/Either.html#getOrElse[B1>:B](or:=>B1):B1

In cats this operation named `valueOr`: http://typelevel.github.io/cats/api/cats/syntax/EitherOps.html#valueOr[BB>:B](f:A=>BB):BB

gcanti:
The proposed `getOrElse` looks more general than the current one and the change is backward compatible (just pass a thunk if you are not interested in the left value). Any drawback?

raveclassic:
I'm for `valueOr` because now it's tiresome to wrap quick default value with `constant` each time when passing to `getOrElse`. So there could be two methods to pass either lazy value (with left value passed to thunk) or non-lazy. Yep, this would change `getOrElse` signature and would be a breaking change but still it would be more aligned with Scala (afaics).

gcanti:
@raveclassic looks like material for another issue, please let's keep the discussion on this PR, which is backward compatible

gcanti:
if no one has any objections, I'm going to merge this

sledorze:
no objections here, it's handy

PR #245, opened 2017-10-07
Either: add fromNullable

gcanti:


src/Either.ts
```diff
@@ -227,6 +227,13 @@ export const mapLeft = <L, M>(f: (l: L) => M) => <A>(fa: Either<L, A>): Either<M
 export const fromOption = <L>(defaultValue: L) => <A>(fa: Option<A>): Either<L, A> =>
   fa.fold(() => left(defaultValue), a => right(a))
 
+/**
+ * Takes a default and a nullable value, if the value is not nully, turn it into
+ * a `Right`, if the value is nully use the provided default as a `Left`
+ */
+export const fromNullable = <L>(defaultValue: L) => <A>(a: A | null | undefined): Either<L, A> =>
+  a == null ? left(defaultValue) : right(a)
```
raveclassic: Nice addition!
Non-strict comparisons are considered a bad practice in JS though even if it doesn't really matter in this particular case.

PR #248, opened 2017-10-09
add type constraint to Either and Validation

gcanti:


raveclassic:
- should we keep `package-lock.json`? It's useful with npm >= 5 as it locks dependencies.
- how do you get such formatting for `package.json`? I've got current one automatically with just running `npm i` (npm 5)

gcanti:
> should we keep package-lock.json? It's useful with npm >= 5 as it locks dependencies

fixed 

> how do you get such formatting for package.json?

I just save the file, I think is prettier doing its work

raveclassic:
Ahhh I see - it's not NPM but WebStorm's autoformat, it conflicts with prettier. Ok I'll disable it for jsons on my side.

PR #249, opened 2017-10-09
add Option#filter

raveclassic:
Adds Option#filter method


gcanti:
Thanks @raveclassic. To all: are there any new additions which should go in 0.5.4?

raveclassic:
@gcanti Yep, I'm going to open another PR in a minute

PR #250, opened 2017-10-09
add Option#mapNullable

raveclassic:
This PR adds a handy shortcut to `map(f).chain(fromNullable)` which is useful when working with deep optional structures.

### DIFF
--- a/src/Option.ts
+++ b/src/Option.ts
@@ -42,6 +42,12 @@ export class None<A>
   map<B>(f: (a: A) => B): Option<B> {
     return none
   }
+  /**
+   * Maps `f` over this Option's value. If the value returned from `f` is null or undefined, returns none.
+   */
+  mapNullable<B>(f: (a: A) => B | null | undefined): Option<B> {
+    return none
+  }
   ap<B>(fab: Option<(a: A) => B>): Option<B> {
     return none
   }
@@ -135,6 +141,13 @@ export class Some<A>
   constructor(readonly value: A) {}
   map<B>(f: (a: A) => B): Option<B> {
     return new Some(f(this.value))
+  }
+
+  /**
+   * Maps `f` over this Option's value. If the value returned from `f` is null or undefined, returns none.
+   */
+  mapNullable<B>(f: (a: A) => B | null | undefined): Option<B> {
+    return this.map(f).chain(fromNullable)
   }
   ap<B>(fab: Option<(a: A) => B>): Option<B> {
     return fab.map(f => f(this.value))
@@ -231,6 +244,8 @@ export const getSetoid = <A>(S: Setoid<A>): Setoid<Option<A>> => ({
 })
 
 export const map = <A, B>(f: (a: A) => B, fa: Option<A>): Option<B> => fa.map(f)
+
+export const mapNullable = <A, B>(f: (a: A) => B | null | undefined, fa: Option<A>): Option<B> => fa.mapNullable(f)
 
 export const of = <A>(a: A): Option<A> => new Some(a)
 
--- a/test/Option.ts
+++ b/test/Option.ts
@@ -17,6 +17,7 @@ import {
 import * as array from '../src/Array'
 import { setoidNumber } from '../src/Setoid'
 import { eqOptions as eq } from './helpers'
+import { identity } from '../src/function'
 
 describe('Option', () => {
   it('fold', () => {
@@ -51,6 +52,23 @@ describe('Option', () => {
   it('map', () => {
     const f = (n: number) => n * 2
     eq(map(f, some(2)), some(4))
+  })
+
+  it('mapNullable', () => {
+    type Nested = {
+      foo?: number
+      foo2: {
+        bar2?: string
+      }
+    }
+    const nested: Nested = {
+      foo2: {}
+    }
+    const nestedOption = some(nested)
+    assert.deepEqual(nestedOption.mapNullable(value => value.foo), none)
+    assert.deepEqual(nestedOption.mapNullable(value => value.foo2), some(nested.foo2))
+    assert.deepEqual(nestedOption.mapNullable(value => value.foo2.bar2), none)
+    assert.deepEqual(none.mapNullable(identity), none)
   })
 
   it('ap', () => {


gcanti:
This sounds like too specialized.

> useful when working with deep optional structures

Have you got a real world example? Looks like a job for functional optics

raveclassic:
@gcanti The usecase is pretty-common - an API responds with a deeply-nested structure with almost every field optional (`?:`). So of you need to access deep value through several optional layers you end up with chaining `fromNullable` after each step. The code quickly becomes too verbose:

```typescript

type Phone = {
	number: string
};
type Employment = {
	phone?: Phone
};
type ClientInfo = {
	employment?: Employment
};
type ClientResponse = {
	info?: ClientInfo
};

const getPhoneNumberFromResponse = (response: ClientResponse): Option<string> => {
	return fromNullable(response.info)
		.map(info => info.employment)
		.chain(fromNullable)
		.map(empoyment => empoyment.phone)
		.chain(fromNullable)
		.map(phone => phone.number);
};
```

How would you do this with `monocle-ts`'s `Optional`?

gcanti:
```ts
import { Option, none, some, fromNullable } from 'fp-ts/lib/Option'
import { Optional } from 'monocle-ts'

interface Phone {
  number: string
}
interface Employment {
  phone?: Phone
}
interface ClientInfo {
  employment?: Employment
}
interface ClientResponse {
  info?: ClientInfo
}

const info = new Optional<ClientResponse, ClientInfo>(s => fromNullable(s.info), a => s => ({ ...s, info: a }))
const employment = new Optional<ClientInfo, Employment>(
  s => fromNullable(s.employment),
  a => s => ({ ...s, employment: a })
)
const phone = new Optional<Employment, Phone>(s => fromNullable(s.phone), a => s => ({ ...s, phone: a }))
const number = new Optional<Phone, string>(s => fromNullable(s.number), a => s => ({ ...s, number: a }))

const getPhoneNumberFromResponse = info
  .compose(employment)
  .compose(phone)
  .compose(number)

const response1: ClientResponse = {
  info: {
    employment: {
      phone: {
        number: '555-1234'
      }
    }
  }
}
const response2: ClientResponse = {
  info: {
    employment: {}
  }
}

console.log(getPhoneNumberFromResponse.getOption(response1)) // some("555-1234")
console.log(getPhoneNumberFromResponse.getOption(response2)) // none
```

I'd define an helper function to create those optionals though

```ts
const fromOptionalProp = <S, A extends S[K], K extends keyof S>(k: K): Optional<S, A> =>
  new Optional((s: any) => fromNullable(s[k]), a => s => ({ ...s, [k as any]: a }))

const info = fromOptionalProp<ClientResponse, ClientInfo, 'info'>('info')
const employment = fromOptionalProp<ClientInfo, Employment, 'employment'>('employment')
const phone = fromOptionalProp<Employment, Phone, 'phone'>('phone')
const number = fromOptionalProp<Phone, string, 'number'>('number')
```

Another option (which I like the most) is changing the models

```ts
interface Employment2 {
  phone: Option<Phone>
}
interface ClientInfo2 {
  employment: Option<Employment2>
}
interface ClientResponse2 {
  info: Option<ClientInfo2>
}

const getPhoneNumberFromResponse2 = (response: ClientResponse2): Option<string> =>
  response.info
    .chain(info => info.employment)
    .chain(employment => employment.phone)
    .chain(phone => some(phone.number))

const response3: ClientResponse2 = {
  info: some({
    employment: some({
      phone: some({
        number: '555-1234'
      })
    })
  })
}
const response4: ClientResponse2 = {
  info: some({
    employment: none
  })
}

console.log(getPhoneNumberFromResponse2(response3)) // some("555-1234")
console.log(getPhoneNumberFromResponse2(response4)) // none
```

raveclassic:
@gcanti Thanks for the examples!

However introducing optics seems like an unnecessary complication of the code and it looks like an overkill to provide both getter and setter for each optional field, even with a helper.

The second one with changing `?:` to `Option<>` might though be a nice solution if I'm able to change swagger-codegen template to generate both typings and serialization layer. It requires some ivestigation so I'll try later. But still it would produce unnecessary runtime overhead.

For now, doesn't `mapNullable` seem like a similar helper as `fromNullable` is only on an instance of an `Option`?

gcanti:
> For now, doesn't mapNullable seem like a similar helper as fromNullable is only on an instance of an Option?

Yes you are right, I'm changing my mind. This method looks pretty handy for interop. 

A nitpick: shouldn't be named `chainNullable`?

raveclassic:
> A nitpick: shouldn't be named chainNullable?

I think no, because there's no flattening after mapping. The signature is `Option<A>#mapNullable<B>(f: (a: A) => B | null | undefined): Option<B>`. And with chain it would rather be something like `Option<A>#chainNullable<B>(f: (a: A) => Option<B | null | undefined>): Option<B>` or even `Option<A>#chainNullable<B>(f: (a: A) => Option<B> | null | undefined): Option<B>`.

I can't think of a use case for such operation and don't think such method would be as useful as `mapNullable`. Still it can be added now. Or later when there's a proper use case.

gcanti:
Actually there's a kind of flattening, a Nullable is a kind of raw Option

```ts
// a "raw" Option (i.e. not wrapped)
// where Some<A> = A
// and None = null | undefined

type Option_<A> = A | null | undefined
```

so you get the following signature

```ts
mapNullable<B>(f: (a: A) => Option_<B>): Option<B>
```

which looks like a `chain`. 

---

In fact if you happen to use `Option`s as a representation of an optional field

```ts
interface Phone {
  number: Option<string>
}
interface Employment {
  phone: Option<Phone>
}
interface ClientInfo {
  employment: Option<Employment>
}
interface ClientResponse {
  info: Option<ClientInfo>
}
```

you must change `getPhoneNumberFromResponse` from

```ts
const getPhoneNumberFromResponse = (response: ClientResponse): Option<string> => {
  return fromNullable(response.info)
    .mapNullable(info => info.employment)
    .mapNullable(employment => employment.phone)
    .mapNullable(phone => phone.number)
```

to

```ts
const getPhoneNumberFromResponse = (response: ClientResponse): Option<string> => {
  return fromNullable(response.info)
    .chain(info => info.employment)
    .chain(employment => employment.phone)
    .chain(phone => phone.number)
```

which seems weird (from `map` to `chain`?) while with `chainNullable` you can just remove the `Nullable` part in the name, i.e. from

```ts
const getPhoneNumberFromResponse = (response: ClientResponse): Option<string> => {
  return fromNullable(response.info)
    .chainNullable(info => info.employment)
    .chainNullable(employment => employment.phone)
    .chainNullable(phone => phone.number)
```

to

```ts
const getPhoneNumberFromResponse = (response: ClientResponse): Option<string> => {
  return fromNullable(response.info)
    .chain(info => info.employment)
    .chain(employment => employment.phone)
    .chain(phone => phone.number)
```

raveclassic:
> which looks like a chain

Yep, it only _looks_ like a chain but it's not. The type returned from mapper differs from the container type Option.

> which seems weird (from map to chain?)

It is absolutely correct because I've changed the type. And to enable flattening I should explicitly change working method from plain mapping to flattening chaining. Option is not a Nullable (Option_).

I'm only trying to say that `chain` on the type `Option` should require its callback to return exactly the same type, then it will be a proper chain (correctly if I'm wrong). `chainNullable` as a particular case should also behave the same, but only stripping out `null | undefined` from type inner type which seems weird because it's inner type is already an `Option`. `mapNullable` on the other hand doesn't have such requirements and therefore doesn't make any flattening. It only maps over the inner value stripping out `null | undefined` from it which doesn't actually relate to `Option` type. If it relates then we can just strip in `map` method.

I'm ok with any naming, just trying to sort things out.

gcanti:
Ah now I understand your point of view, thanks for explaining. Both interpretations are somehow acceptable to me for different reasons, so let's stick with the current name (sorry for the lengthy nitpick).

PR #255, opened 2017-10-19
(Quick and dirty) automate documentation

gcanti:
@raveclassic this is another attempt to generate documentation from source files (I developed a custom script).


docs/api/parser.ts
```diff
@@ -0,0 +1,202 @@
...
+const parseTypeAliasDeclarationData = (sf: SourceFile, tad: TypeAliasDeclaration): Option<Data> => {
+  const annotation = getAnnotation(tad.getDocumentationCommentNodes())
+  if (isData(annotation)) {
+    const dataName = tad.getName()
+    const signature = tad.getText().substring('export '.length)
+    const description = fromJSDocDescription(annotation.description)
+    const constructors: Array<Constructor> = annotation.tags.filter(tag => isConstructor(tag)).map(tag => {
+      const name = tag.name
+      if (typeof name === 'undefined') {
+        throw new Error(`Missing constructor name in ${dataName}`)
```
raveclassic: Should pure fp library use `throw` in it's sources? Isn't it a poor example? ;)

gcanti: eh... I added "Quick and dirty" in the title specifically for that 😛 (and some poor string parsing..)

More seriously, if the result feels ok we can improve the code at a later stage

gcanti:
(ominous throws removed)

@raveclassic please take a look at the general result here and let me know what you think

[API index](https://github.com/gcanti/fp-ts/blob/776aa1d7162261e0fde19fbc6a2c871abcda03ae/docs/api/md/index.md)

The relevant info should be there (and we can change the template anytime)

docs/api/parser.ts
raveclassic: Sounds good :)

raveclassic:
Everything looks fine

gcanti:
@raveclassic thanks for reviewing 👍

PR #269, opened 2017-11-14
Fix/move semigroup methods

valery-paschenkov:
Hi,

This PR moves semigroup methods from Monoid to Semigroup module, like it's done in Ord and Setoid modules.


gcanti:
Thanks @valery-paschenkov

PR #270, opened 2017-11-15
docs: explain how HKT, URI2HKT, HKTS, HKTAs work

gcanti:


docs/HKT.md
```diff
@@ -0,0 +1,168 @@
...
+}
+```
+
+and `Functor`
+
+```ts
+// Functor.ts
+
+export interface Functor<F> {
+  readonly URI: F
```
raveclassic: Does Functor need a URI field?

docs/HKT.md
```diff
@@ -0,0 +1,168 @@
...
+
+`x` is not usable
+
+```ts
+x.value // static error: Property 'value' does not exist on type 'HKT<"Identity", number>'
+```
+
+## The solution
+
+We must somehow teach TypeScript that `HKT<"Identity", number>` is really `Identity<number>`, or more generally that `HKT<"Identity", A>` is `Identity<number>` for all `A`.
```
raveclassic: It should be `Identity<A>` for all `A`

docs/HKT.md
```diff
@@ -0,0 +1,168 @@
...
+```
+
+Adding an entry means to leverage the module augmentation feature
+
+```ts
+// Identity.ts
+
+declare module './HKT' {
+  interface URI2HKT<A> {
+    Identity: Identity<A> // maps the URI "Identity" to the type constructor `Identity`
```
raveclassic: I think it should be stressed that the key "Identity" is actually the same value as `URI` constant and type defined with a HKT.

docs/HKT.md
```diff
@@ -0,0 +1,168 @@
...
+
+x.value // ok
+```
+
+In order to make easier to define an overloading, we can write some type aliases
+
+```ts
+// HKT.ts
+
+export type HKTS = keyof URI2HKT<any>
```
raveclassic: `HKTS` stands for all registered with module augmentation type constructor URIs

docs/HKT.md
```diff
@@ -0,0 +1,168 @@
...
+
+export type HKTAs<F extends HKTS, A> = URI2HKT<A>[F]
```
raveclassic: And `HKTAs` extracts actual registered type constructor from the storage by its URI and also for passed A type argument. So it's a final usable type.

docs/HKT.md
gcanti: With recent versions of TypeScript? Not sure

raveclassic: I got it - this `URI` is used when creating an _instance_ of `Functor`. And `_URI` is used to satisfy `HKT` interface when implementing it with custom type class. I think it should be also in the docs otherwise it's not clear what's the difference between `URI` and `_URI`.

docs/HKT.md
```diff
@@ -0,0 +1,173 @@
...
+  map<B>(f: (a: A) => B): Identity<B> {
+    return new Identity(f(this.value))
+  }
+}
+
+export const map = <A, B>(f: (a: A) => B, fa: Identity<A>): Identity<B> => {
+  return fa.map(f)
+}
+
+export const identity: Functor<URI> = {
```
raveclassic: Here is this place with an _instance_

raveclassic:
Awesome!

gcanti:
Thanks @raveclassic for reviewing, going to squash and merge

PR #279, opened 2017-12-05
Removed duplicate fold definition from Either

pnicolli:
Found a duplicate definition for the fold signature of Either (Flow). This microfix just removes it.


gcanti:
Good catch, thanks @pnicolli

PR #286, opened 2017-12-13
refactor concat, snoc, cons, closes #284

### LINKED ISSUE #284: Optimize array concat
Should bench the solution as this comment may lead us to believe we can significantly benefit from a custom solution.

https://github.com/gcanti/io-ts/commit/1e3c14fa5b21d0e8055c582d3f345616ff1ef5d1#diff-66016d1aae94ecee8cc1138091ae8d18R28

gcanti:


perf/cons.js
```diff
@@ -0,0 +1,48 @@
...
+  return [a].concat(as)
+}
+
+const cons2 = a => as => {
+  const len = as.length
+  const r = Array(len + 1)
+  for (let i = 0; i < len; i++) {
+    r[i] = as[i]
+  }
+  r.unshift(a)
```
sledorze: @gcanti if you unshift, no need to allocate the array with 'len+1'.

sledorze:
@gcanti are you ok with the chain / flatten approach?

sledorze:
@gcanti here are the perf diff I'm proposing (low hanging fruit)

map
```
map1 x 7,609,379 ops/sec ±0.49% (89 runs sampled)
map2 x 28,970,550 ops/sec ±1.13% (84 runs sampled)
```

chain
```
chain1 x 4,413,103 ops/sec ±0.74% (87 runs sampled)
chain2 x 12,084,771 ops/sec ±1.00% (87 runs sampled)
```

flatten
```
flatten1 x 7,799,193 ops/sec ±1.78% (84 runs sampled)
flatten2 x 17,044,084 ops/sec ±1.79% (82 runs sampled)
```

sledorze:
@gcanti basic rule that starts to appear: NEVER use native functions, period.

gcanti:
@sledorze Awesome. Added `chain3`, `filter3` to perf, they look faster. WDYT?

sledorze:
AFK do they or do they are? (Small / big arrays) if so LGTM!

sledorze:
Puzzled by reverse1 variation

gcanti:
Ah it depends on the array length. After more tests:

filter3: yes, it's faster
chain3: nope

Note that for length > 15 chain1 is faster then chain2, so I would keep chain1

sledorze:
@gcanti Ok good! Thanks I had not enough time to do all the small/big collection comparison
May we go the path were we switch the implementation based on the length? I'm hesitating because:

- It can change over time and would need to be backed by comparison perf test (which take some time to run)
- It may change based on the actual VM (or not).

On the other side, getting a *3 / *4 boost on some operations is a lot.

It looks like a whole subject and may be should be part of another lib on which fp-ts could rely? (maybe there's an existing one)

gcanti:
> May we go the path were we switch the implementation based on the length?

IMO no. I'm fine with some changes if the resulting code is not too messy and the gain is big (3x-4x) but I wouldn't go that route

PR #289, opened 2017-12-20
Ordering: add fromNumber, toNumber

gcanti:


src/Ordering.ts
```diff
@@ -3,6 +3,16 @@ import { Semigroup } from './Semigroup'
 
 export type Ordering = 'LT' | 'EQ' | 'GT'
 
+/** @function */
+export const fromNumber = (n: number): Ordering => {
```
raveclassic: I was also thinking about something like that. What about changing Ordering from string literal union to numeric enum?
```typescript
enum Ordering {
  LT = -1,
  EQ = 0,
  GT = 1
}
```

gcanti: @raveclassic that would be a breaking change (perhaps we could start to gather such proposed changes in an issue for discussion)

raveclassic: I've opened https://github.com/gcanti/fp-ts/issues/290 to track them

PR #310, opened 2018-01-31
upgrade to TypeScript 2.7.1, closes #306

### LINKED ISSUE #306: TypeScript 2.7
This is not an issue, I just want to say that the current version of `fp-ts` (0.6.8) is `typescript@2.7.0-rc` compliant.

With custom data types, you may want to quote the phantom fields in order to silent the new `--strictPropertyInitialization` flag

```diff
export class Task<A> implements FantasyMonad<URI, A> {
- readonly _A: A
+ readonly '_A': A
- readonly _URI: URI
+ readonly '_URI': URI
  constructor(readonly run: Lazy<Promise<A>>) {}
}
```

(or use the new `!` syntax)

Reference [Announcing TypeScript 2.7 RC](https://blogs.msdn.microsoft.com/typescript/2018/01/17/announcing-typescript-2-7-rc/)

gcanti:


package.json
```diff
@@ -47,7 +47,7 @@
     "ts-simple-ast": "0.74.0",
     "tslint": "4.4.2",
     "tslint-config-standard": "4.0.0",
-    "typescript": "2.6.2"
+    "typescript": "^2.7.1"
```
raveclassic: It would be better to lock the version with `~` instead of `^`

gcanti: isn't pinned to `2.7.1` by `package-lock.json`?

raveclassic: oh my, forgot about that :D

gcanti: eheh no problem (and thanks for reviewing btw)

PR #312, opened 2018-02-06
1.0.0

gcanti:
@raveclassic @sledorze

Changelog: see commits

Try: `npm i gcanti/fp-ts#next`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,11 @@
 **Note**: Gaps between patch versions are faulty/broken releases. **Note**: A feature tagged as Experimental is in a
 high state of flux, you're at risk of it changing without notice.
 
+# 1.0.0
+
+* **Breaking Change**
+  * see https://github.com/gcanti/fp-ts/pull/312 (@gcanti)
+
 # 0.6.8
 
 * **New Feature**
--- a/LICENSE
+++ b/LICENSE
@@ -1,6 +1,6 @@
 MIT License
 
-Copyright (c) 2017 Giulio Canti
+Copyright (c) 2017-2018 Giulio Canti
 
 Permission is hereby granted, free of charge, to any person obtaining a copy
 of this software and associated documentation files (the "Software"), to deal
--- a/docs/HKT.md
+++ b/docs/HKT.md
@@ -1,4 +1,4 @@
-# How `HKT`, `URI2HKT`, `HKTS` and `HKTAs` work
+# How `HKT`, `URI2HKT`, `URIS` and `Type` work
 
 ## Introduction
 
@@ -17,14 +17,15 @@ and `Functor`
 
 ```ts
 // Functor.ts
+import { HKT } from 'fp-ts/lib/HKT'
 
 export interface Functor<F> {
   readonly URI: F
-  map<A, B>(f: (a: A) => B, fa: HKT<F, A>): HKT<F, B>
+  map: <A, B>(fa: HKT<F, A>, f: (a: A) => B) => HKT<F, B>
 }
 
 export function lift<F>(F: Functor<F>): <A, B>(f: (a: A) => B) => (fa: HKT<F, A>) => HKT<F, B> {
-  return f => fa => F.map(f, fa)
+  return f => fa => F.map(fa, f)
 }
 ```
 
@@ -32,6 +33,7 @@ We can define an instance of `Functor` for `Identity`
 
 ```ts
 // Identity.ts
+import { Functor } from 'fp-ts/lib/Functor'
 
 export const URI = 'Identity'
 
@@ -46,7 +48,7 @@ export class Identity<A> {
   }
 }
 
-export const map = <A, B>(f: (a: A) => B, fa: Identity<A>): Identity<B> => {
+const map = <A, B>(fa: Identity<A>, f: (a: A) => B): Identity<B> => {
   return fa.map(f)
 }
 
@@ -64,7 +66,7 @@ export const identity: Functor<URI> = {
 const double = (n: number): number => n * 2
 
 // x: Identity<number>
-const x = map(double, new Identity(1))
+const x = Identity(1).map(double)
 ```
 
 However there's a problem with functions which abstracts over `Functor` like `lift`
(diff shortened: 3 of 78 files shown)


gcanti:
@raveclassic @sledorze (and everybody else) 

Next steps? What can I do to help you testing this version, maybe in a real scenario?

One option would be

- publish this version in the `@next` channel
- publish a `io-ts` version pointing to ^ in the `@next` channel (just verified: no breaking changes)
- maybe the same thing for `monocle-ts` (?) (just verified: no breaking changes)

Are there other libraries around fp-ts that are you using?

raveclassic:
@gcanti I'm using `typelevel-ts` but it seems like it shouldn't be affected. Unfortunately our project is being stabilized now before its release so I can't update fp-ts to test it on large codebase.

sledorze:
@gcanti we're using `monocle-ts` too.
I'm a few days before a 1 week holidays so will *really* try to give it a go but no promises.

gcanti:
Ok, published on npm

- `fp-ts@next` (`fp-ts@1.0.0-rc`)
- `io-ts@next` (`io-ts@0.9.8-rc`)
- `monocle-ts@next` (`monocle-ts@0.5.3-rc`)
- `io-ts-types@next` (`io-ts-types@0.2.5-rc`)
- `newtype-ts@next` (`newtype-ts@0.1.3-rc`)

Except fp-ts, the other packages should **NOT** have API breaking changes

gcanti:
Those releases are mainly for early feedback, I'm waiting to update the documentation when things settle down and we are satisfied by the result. 

Namely

- changelog (will be huge)
- docs/HKT.md

Conversely `examples` and `exercises` should be up to date

sledorze:
@gcanti Here I will consolidate the issues I encounter:

Problem with Either predicates.

Previous version:
```typescript
        import { either } from 'fp-ts'
...
        arrayOfEithers
        .filter(either.isRight) // predicate
        .map(x => ...) // x is Right<..> :)
```

New version:
```typescript
        arrayOfEithers
        .filter(x => x.isRight()) // predicate refines x 
        .map(x => ...) // x is Either<..> :(
```
We're missing that kind of support:

```typescript
export const isRight = <A, B>(x: either.Either<A, B>): x is either.Right<A, B> =>
  x.isRight()
```

gcanti:
@sledorze I would argue that using `Right` as a type is wrong: `Right` is a constructor not a type (well in TypeScript _is_ also a type because of the implementation and because I'm forced to export `Left` and `Right` by TypeScript itself but should be considered as an implementation detail)

In PureScript for example you can't use `Right` as a type

```purescript
data Either l r = Left l | Right r

x :: Right String -- doesn't compile
x = Right "foo"

x :: Either Number String -- ok
x = Right "foo"
```

p.s.

Also in your particular case there's [`Array.rights`](https://github.com/gcanti/fp-ts/blob/da8091ef4c2f7908e71149ed4ad956fdc6f613cf/src/Array.ts#L510)

sledorze:
@gcanti this is not on array but on (rxjs) observables the use case is not easy without those refinements.
The only solution is to recreate that top level predicate (idiomatic) OR to define another Observable combinator (which a end user would see as the least solution).
Also `Right`, as implemented, is a type in `typescript` and its usage is very useful and idiomatic; why restricting it to the possibilities of Purescript.
That's why I think we should (re)add it (based on a real world code base).

sledorze:
@gcanti 
Note on the order of params.
I see that Functor and Chain params are now inverted like so:
```
chain: <A, B>(fa: HKT<F, A>, f: (a: A) => HKT<F, B>) => HKT<F, B>;
map: <A, B>(fa: HKT<F, A>, f: (a: A) => B) => HKT<F, B>;
```

But on Apply that's not the case:
```
ap: <A, B>(fab: HKT<F, (a: A) => B>, fa: HKT<F, A>) => HKT<F, B>;
```
is it deliberate? (if so, why?)

gcanti:
> why restricting it to the possibilities of Purescript

PureScript is my reference when it comes to establish what's good and what's bad, in particular with respect to functional programming.

However if there is consensus I'm ok to re-add those custom guards because

- clearly you find them useful
- one less breaking change

sledorze:
Ok anyway i m adding the missing bits into a file so that we can review it later.

gcanti:
> is it deliberate? (if so, why?)

Based on my early experiments there's no gain in terms of type inference / suggestions by swapping the `Apply` arguments. I'll investgate further though

sledorze:
@gcanti i was specifically thinking about approachability by homogeneity (if not, wondering what s the default/preferred behavior)

sledorze:
@gcanti I find we lack meaningful definition for Ordering (I miss LT, EQ, GT) even if they map to -1, 0, 1 in the end.

sledorze:
@gcanti Ok so I've ported our code base on the new version.

Notable:

- A lot of foldL usage (63).

- new order of args leads to better inference AND looks better also with prettier; often we can see everything in the first line as the biggest param (the function) is the last one.

- Took me a bit of time to understand there was new Monad1, Monad2, etc.. while porting some monadTransformers.

- miss the isRight / isSome a lot in the front-end side (observables) by our frontEnd guy; for that kind of two stage usages: ```opt$.filter(isSome).map(x => x.value)```.

- no array.concat; I'm using array.flatten, I don't know if it's better or not performance wise.

- Less type params to write explicitly.

Overall, I would consider this version as ready for usage.

sledorze:
@gcanti oh I forgot to say that io-ts-types are not referencing the new libs explicitly.

gcanti:
@sledorze great feedback, thanks

> miss the isRight / isSome

ok I'm going to re-add them (*)

> no array.concat

you can import `concat` either from the `function` module or calling `Array.getSemigroup()`

> io-ts-types are not referencing the new libs explicitly

My plan would be

- release `fp-ts@1.0.0` (after (*))
- merge https://github.com/gcanti/io-ts/pull/112 and release `io-ts@1.0.0`
- update all remaining libraries (inluded a cleaned up `io-ts-types`)

gcanti:
UPDATE

I'm working on a bunch of PRs, the goal would be to coordinate all releases in one shot. Here's the dependencies

```
fp-ts (v1.0.0) -- io-ts (v1.0.0)
              \-- monocle-ts (v1.0.0)
                                    \-- newtype-ts (v0.2.0)
                                                           \-- io-ts-types (v0.3.0)
```

You can install the current candidates by running

- `npm i gcanti/fp-ts#next` 
- `npm i gcanti/io-ts#next` (related PR [here](https://github.com/gcanti/io-ts/pull/112))
- `npm i gcanti/monocle-ts#next` (related PR [here](https://github.com/gcanti/monocle-ts/pull/41))
- `npm i gcanti/newtype-ts#next` (related PR [here](https://github.com/gcanti/newtype-ts/pull/6))
- `npm i gcanti/io-ts-types#next` (related PR [here](https://github.com/gcanti/io-ts-types/pull/17))

(for these kind of tests / quick feedback installing from github looks more practical than publishing on npm)

Let me know if you hit any issue

sledorze:
Ok, got no issue so far (but was not able to test with your latest changes - aka type guards - as they've not been published yet).

PR #320, opened 2018-02-15
Add scan left/right functions

PaNaVTEC:
Hi, 

I've observed that in haskell/scala they have a scan function which allows to do a fold carrying intermediate steps and I've translated the haskell implementation to TS.

http://hackage.haskell.org/package/base-4.10.1.0/docs/src/GHC.List.html#scanl
https://github.com/scala/collection-strawman/blob/57340b9fa6732a0bae37845f16dfb272fbb2173a/collections/src/main/scala/strawman/collection/immutable/LazyList.scala#L271

### DIFF
--- a/docs/api/md/Array.md
+++ b/docs/api/md/Array.md
@@ -129,7 +129,7 @@ _function_
 ```ts
 <A, B>(f: ((b: B, a: A) => B), b: B, as: Array<A>): Array<B>
 ```
-Same as `fold` but it carries over the intermediate steps
+Same as `reduce` but it carries over the intermediate steps
 
 # getMonoid
 
--- a/docs/api/md/Array.md
+++ b/docs/api/md/Array.md
@@ -122,6 +122,15 @@ _function_
 
 Lazy version of `fold`
 
+# scan
+
+_function_
+
+```ts
+<A, B>(f: ((b: B, a: A) => B), b: B, as: Array<A>): Array<B>
+```
+Same as `fold` but it carries over the intermediate steps
+
 # getMonoid
 
 _function_
--- a/src/Array.ts
+++ b/src/Array.ts
@@ -205,6 +205,16 @@ export const scanLeft = <A, B>(as: Array<A>, b: B, f: ((b: B, a: A) => B)): Arra
   return r
 }
 
+export const scanRight = <A, B>(as: Array<A>, b: B, f: ((a: A, b: B) => B)): Array<B> => {
+  const l = as.length
+  let r = new Array(l + 1)
+  r[l] = b
+  for (let i = l - 1; i >= 0; i--) {
+    r[i] = f(as[i], r[i + 1])
+  }
+  return r
+}
+
 /**
  * Test whether an array is empty
  * @function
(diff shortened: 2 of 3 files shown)


sledorze:
I think we should warn for stack safety.

PaNaVTEC:
@sledorze Thanks for your comment, you are right. 
What about this implementation: 

```export const scan = <A, B>(f: ((b: B, a: A) => B), b: B, as: Array<A>): Array<B> => {
  const l = as.length
  let r = [b]
  for (let i = 0; i < l; i++) {
    r.push(f(r[i], as[i]))
  }
  return r
}
```
It does the same and it is iterative, so we don't have to worry about that

sledorze:
@PaNaVTEC thanks!
Sorry for being picky but performance wise, it would be nice to follow that preallocation scheme (which was backed by actual measurement):
https://github.com/gcanti/fp-ts/blob/master/src/Array.ts#L45-L52

I think we're close to it :)

PaNaVTEC:
@sledorze true again, I didn't even noticed. Thanks for your review :)

sledorze:
@PaNaVTEC I'm sorry, shouldn't it be more like:

```typescript
export const scan = <A, B>(f: ((b: B, a: A) => B), b: B, as: Array<A>): Array<B> => {
  const l = as.length
  let r = new Array(l+1)
  r[0] = b;
  for (let i = 0; i < l; i++) {
    r[i+1] = f(r[i], as[i]);
  }
  return r
}
```

sledorze:
@PaNaVTEC also I looks to briefly but to fit with the rest of the lib you may reverse the params.

gcanti:
About the name, there's `scanr` too, so maybe `scanl` (or `scanLeft`)?

> you may reverse the params

Yes, type inference for `f` will be much better with the following signature

```ts
<A, B>(as: Array<A>, b: B, f: (b: B, a: A) => B): Array<B>
```

sledorze:
@gcanti Oh yes, `scan` may be ambiguous `scanl`.

PaNaVTEC:
@gcanti @sledorze I didn't want to add that l in the name because I've seen that you guys have a `reduce` function but there is not `reduceRight` and `fold` has a `foldL` but that L means Lazy, I can add if it is clearer tho

sledorze:
@PaNaVTEC good point, @gcanti let's discuss it!

sledorze:
One possibility is to name the combinators 'xxxLeft' and 'xxxRight' and keep an extra 'L' for their lazy variant.
Have a default alias without 'Left' or 'Rigth' but make that clear in the comment.

gcanti:
In `Foldable` there's 

- `foldl`, however is called `reduce`, vestiges of the past (fantasy-land compatibility)
- and `foldr`

I'd go for `scanl` or `scanLeft` (but no aliases)

gcanti:
> the intersection of current input goes in favour of scanLeft

ok for `scanLeft`

@PaNaVTEC there's no need to update the documentation, I'll handle that before releasing next version, thanks

PaNaVTEC:
Thank you for your help guys. I've added `scanRight` and renamed the other to `scanLeft`

sledorze:
@PaNaVTEC that's great!

PaNaVTEC:
Since @gcanti said there is not need to update the documentation, is there anything else I can do to merge this?

gcanti:
Thanks @PaNaVTEC

PR #324, opened 2018-02-19
Doc: Option, alt method

piq9117:


src/Option.ts
```diff
@@ -70,6 +70,18 @@ export class None<A> {
   reduce<B>(b: B, f: (b: B, a: A) => B): B {
     return b
   }
+  /**
+   * `alt` short for alternative, takes another `Option`. If this `Option` is a `Some` type
+   * then it will be returned, if it is a `None` then it will return the next `Some` if it exist. If
+   * both are `None` then it will return `none`.
+   *
+   * For example
+   * `const someFn = (o: option.Option<number>) => o.alt(option.some(4))`
```
gcanti: @piq9117 I think we can omit `option.`. `Option`, `some` and `none` should be clear in this context. Also do not include the generated docs (I'll handle that before releasing next version, thanks)

piq9117: gotcha. thanks!

package-lock.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "fp-ts",
-  "version": "1.0.0-rc",
+  "version": "1.0.1",
```
raveclassic: Please, avoid changing `package-lock`

piq9117: 😱 oh no. i didnt mean to change that. I apologize.

src/Option.ts
gcanti: @piq9117 thanks, last thing before merging

> do not include the generated docs (I'll handle that before releasing next version, thanks)

piq9117: totally forgot about that. Hopefully I didn't miss anything. 😅

gcanti: @piq9117 sorry I wasn't clear, I meant to revert the changes to docs/api/md/Option.md, not to remove the file

gcanti: @piq9117 ok never mind, I'll handle this in the next commits. Thanks for this PR

piq9117:
Thanks for all the patience! I'm gonna remember all these on my next PR 😅

Issue #323, opened 2018-02-18
`either.tryCatch`: `onerror` param?


OliverJAsh:
`task.tryCatch` has an `onrejected` param. This is useful because it allows you to define the type of the error:

``` ts
export declare const tryCatch: <L, A>(f: Lazy<Promise<A>>, onrejected: (reason: {}) => L) => Task<Either<L, A>>;
```

However, `either.tryCatch` does not have this. Instead, it assumes the error type will be `Error`—but as we know, that's not always the case in JS!

Would you accept a PR to add an `onerror` param to `either.tryCatch`?

gcanti:
@raveclassic I think @OliverJAsh was referring to, for example, throwing a string

```ts
console.log(
  tryCatch(() => {
    throw 'myerror'
  }).fold(e => e.message.trim(), () => 'ok')
) // runtime error: Cannot read property 'trim' of undefined
```

raveclassic:
@gcanti Yeah, I get it. I've also looked at the generated code when extening Error and it seems like TS correctly emits call to `setPrototypeOf`, so I think it's safe to check `e instanceof Error`

raveclassic:
accidentaly closed, sorry, reopening

raveclassic:
type refinement is done in #325

OliverJAsh:
I like the idea of refining the type generally as in #325, but I don't know how useful this is in practice. Often, only you know how to refine the type when wrapping a specific function in `tryCatch`. For example, if a non-`Error` object is thrown, you might want to do some duck typing and then extract properties from it:

``` js
tryCatch(() => myFn(), thrown => {
  if ('statusCode' in thrown) {
    return new Error(`Bad response: ${thrown.statusCode});
  } else if (thrown instanceof Error) {
    return thrown;
  } else {
    return new Error('Unexpected error');
  }
});
```

Currently, we try to stringify all thrown values, which may result in `[object Object]` :-(

raveclassic:
@OliverJAsh How can type-checker be sure about `thrown` type? You can declare any type there which will lead to runtime error.

OliverJAsh:
It's not completely safe, but you can use user-defined type guards:

``` ts
type ObjectWithStatusCode = { statusCode: number }
const thrownIsObjectWithStatusCode = (thrown: {}): thrown is ObjectWithStatusCode => typeof thrown === 'object' && 'statusCode' in thrown;

(thrown: {}) => {
[...]

PR #325, opened 2018-02-19
Bug Fix: Either.tryCatch now refines the error

gcanti:


gcanti:
@raveclassic thanks!

src/Either.ts
```diff
@@ -281,7 +281,11 @@ export const tryCatch = <A>(f: Lazy<A>): Either<Error, A> => {
   try {
     return right(f())
   } catch (e) {
-    return left(e)
+    if (e instanceof Error) {
+      return left(e)
+    } else {
+      return left(new Error(String(e)))
```
OliverJAsh: What happens if the error thrown is an object, which is so often the case is badly written JS libs? `String({ message: 'foo' }) // => [object Object]` 😒

raveclassic: This exactly what I was pointing to in the issue #323

raveclassic: Anyway, this bug is about incorrect return type Error for things that actually aren't Errors in tryCatch. We still need some proper solution to proccess incorrect thrown entities.

PR #326, opened 2018-02-20
New Feature: add an optional `onerror` argument to `Either.tryCatch`,…

### LINKED ISSUE #323: `either.tryCatch`: `onerror` param?

gcanti:
… fix #323

### DIFF
--- a/src/Either.ts
+++ b/src/Either.ts
@@ -276,16 +276,20 @@ export const fromNullable = <L>(defaultValue: L) => <A>(a: A | null | undefined)
   return a == null ? left(defaultValue) : right(a)
 }
 
+const toError = (e: {}): Error => {
+  if (e instanceof Error) {
+    return e
+  } else {
+    return new Error(String(e))
+  }
+}
+
 /** @function */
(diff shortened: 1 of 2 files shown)


src/Either.ts
```diff
@@ -276,16 +276,20 @@ export const fromNullable = <L>(defaultValue: L) => <A>(a: A | null | undefined)
   return a == null ? left(defaultValue) : right(a)
 }
 
+const toError = (e: {}): Error => {
```
raveclassic: Maybe this could be exported as well? Quite a useful helper

raveclassic:
Nice!

PR #327, opened 2018-02-20
Polish: remove Array.spanIndexUncurry

gcanti:


gcanti:
@sledorze Never mind, after some benchmark this would make performance worse

PR #328, opened 2018-02-22
Make taskEither an instance of BiFunctor

teves-castro:
This PR makes taskEither an instance of BiFunctor.

### DIFF
--- a/src/TaskEither.ts
+++ b/src/TaskEither.ts
@@ -4,6 +4,7 @@ import * as task from './Task'
 import { Task } from './Task'
 import { Monad2 } from './Monad'
 import { Lazy } from './function'
+import { Bifunctor2 } from './Bifunctor'
 
 declare module './HKT' {
   interface URI2HKT2<L, A> {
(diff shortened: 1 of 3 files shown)


src/TaskEither.ts
```diff
@@ -58,6 +59,9 @@ export class TaskEither<L, A> {
   orElse<M>(f: (l: L) => TaskEither<M, A>): TaskEither<M, A> {
     return new TaskEither(this.value.chain(e => e.fold(l => f(l).value, a => eitherTTask.of(a))))
   }
+  bimap<V, B>(f: (l: L) => V, g: (a: A) => B): TaskEither<V, B> {
```
gcanti: @teves-castro what you think about adding `bimap` to `EitherT`

```ts
export function bimap<F extends URIS2>(
  F: Functor2<F>
): <M, L, V, A, B>(fa: Type2<F, M, Either<L, A>>, f: (l: L) => V, g: (a: A) => B) => Type2<F, M, Either<V, B>>
export function bimap<F extends URIS>(
  F: Functor1<F>
): <L, V, A, B>(fa: Type<F, Either<L, A>>, f: (l: L) => V, g: (a: A) => B) => Type<F, Either<V, B>>
export function bimap<F>(
  F: Functor<F>
): <L, V, A, B>(fa: HKT<F, Either<L, A>>, f: (l: L) => V, g: (a: A) => B) => HKT<F, Either<V, B>>
export function bimap<F>(
  F: Functor<F>
): <L, V, A, B>(fa: HKT<F, Either<L, A>>, f: (l: L) => V, g: (a: A) => B) => HKT<F, Either<V, B>> {
  return (fa, f, g) => F.map(fa, e => e.bimap(f, g))
}
```

and then derive `TaskEither`'s `bimap` method from that, like I did with `mapLeft`?

teves-castro: That indeed sounds more interesting.

teves-castro: @gcanti done.

gcanti:
Thanks @teves-castro

PR #329, opened 2018-02-24
EitherT fixes

gcanti:


src/EitherT.ts
```diff
@@ -14,15 +14,15 @@ import { Either, URI as URIEither } from './Either'
 import * as either from './Either'
 
 export interface EitherT<F> extends ApplicativeComposition<F, URIEither> {
-  chain<L, A, B>(f: (a: A) => HKT<F, Either<L, B>>, fa: HKT<F, Either<L, A>>): HKT<F, Either<L, B>>
+  chain: <L, A, B>(f: (a: A) => HKT<F, Either<L, B>>, fa: HKT<F, Either<L, A>>) => HKT<F, Either<L, B>>
```
sledorze: @gcanti for information; what is the difference exactly? (beside the syntax)
Does the inference works differently?

gcanti: @sledorze This way I'm quite sure bivariance doesn't kink in

sledorze: Quite sure or sure? (Have you reproduced it?)

gcanti: My observation is based on early experiments when `strictFunctionTypes` was added, here's a repro

```ts
export interface HKT<URI, A> {
  readonly _URI: URI
  readonly _A: A
}

export interface BadFunctor<F> {
  readonly URI: F
  map<A, B>(fa: HKT<F, A>, f: (a: A) => B): HKT<F, B>
}

export interface GoodFunctor<F> {
  readonly URI: F
  map: <A, B>(fa: HKT<F, A>, f: (a: A) => B) => HKT<F, B>
}

export const URI = 'Identity'

export type URI = typeof URI

export class Identity<A> {
  readonly _A!: A
  readonly _URI!: URI
  constructor(readonly value: A) {}
  map<B>(f: (a: A) => B): Identity<B> {
    return new Identity(f(this.value))
  }
}

const map = <A, B>(fa: Identity<A>, f: (a: A) => B): Identity<B> => {
  return fa.map(f)
}

// no error
export const bad: BadFunctor<URI> = {
  URI,
  map
}

// error: Type 'HKT<"Identity", A>' is not assignable to type 'Identity<A>'
export const good: GoodFunctor<URI> = {
  URI,
  map
}
```

PR #333, opened 2018-02-26
Api Doc(Option): ap & ap_

piq9117:


gcanti:
Thanks! ;)

src/Option.ts
```diff
@@ -53,9 +53,30 @@ export class None<A> {
...
+   *
+   * `const someFn = some(2).ap(some(x => x + 1))` will return `some(3)`.
+   *
+   * `const someFn = none.ap(some(x => x + 1))` will return `none`.
+   */
   ap<B>(fab: Option<(a: A) => B>): Option<B> {
     return none
   }
+  /**
+   * Similar to `ap` but instead of taking a function it takes `some` value or `none`, then applies this `Option`'s
```
raveclassic: You can use `{@link ap}` to link to the method

src/Option.ts
```diff
@@ -53,9 +53,30 @@ export class None<A> {
...
+   * wrapped function to the `some` or `none`. If the `Option` calling `ap_` is `none` it will return `none`.
+   *
+   * For example:
+   *
```
raveclassic: You can use `@example` tag, it's recognized as preformatted code by some editors/IDEs

PR #334, opened 2018-02-27
Latest prettier + use definite assignement assertion for phantom fields

gcanti:


raveclassic:
This drops support for previous versions of TS. Is it ok?

gcanti:
@raveclassic I checked the generated definition files and nothing changed, AFAIK previous versions are still supported (I guess that definite assignment assertions are useless for consumers so they are removed)

/cc @alexandru

alexandru:
Btw, I also used those definite assignment assertions, thanks for the tip.

PR #338, opened 2018-03-06
add fromArray to Set (fixes #336)

### LINKED ISSUE #336: Define 'fromArray' in Set
or maybe `fromFoldable` ?
What do you think @gcanti ?

Something we could have used here:

https://github.com/gcanti/io-ts-types/blob/master/src/fp-ts/createSetFromArray.ts#L36-L43

sledorze:


src/Set.ts
```diff
@@ -165,3 +165,20 @@ export const insert = <A>(S: Setoid<A>): ((a: A, x: Set<A>) => Set<A>) => {
...
+ * Create a set from an array
+ * @function
+ */
+export const fromArray = <A>(S: Setoid<A>) => (as: A[]): Set<A> => {
+  const len = as.length
+  const r = new Set<A>()
+  const equals = S.equals
+  for (let i = 0; i < len; i++) {
+    const a = as[i]
+    if (!some(r, x => equals(x, a))) {
```
gcanti: is there a reason why you are not using `member` here?

sledorze: you're right, changed!

PR #343, opened 2018-03-06
add map for Set (Fixes #342)

### LINKED ISSUE #342: add 'map' for Set

sledorze:


src/Set.ts
```diff
@@ -31,6 +31,21 @@ export const some = <A>(x: Set<A>, predicate: Predicate<A>): boolean => {
   return found
 }
 
+/**
+  * Projects a Set through a function
+  *  @function */
+export const map = <A, B>(x: Set<A>, bset: Setoid<B>, f: (x: A) => B): Set<B> => {
```
gcanti: Following the convention used so far, shouldn't be

```ts
export const map = <B>(S: Setoid<B>) => <A>(x: Set<A>, f: (x: A) => B): Set<B> => {
  ...
}
```

?

src/Set.ts
```diff
@@ -31,6 +31,21 @@ export const some = <A>(x: Set<A>, predicate: Predicate<A>): boolean => {
...
+  const ismember = member(bset)
+  const r = new Set<B>()
+  x.forEach(e => {
+    const v = f(e)
+    if (!ismember(r)(v)) {
```
gcanti: Maybe we can do less work here defining `ismember` as

```ts
const r = new Set<B>()
const ismember = member(bset)(r)
```

sledorze: Indeed, also I'm not sure what to favor between using `forEach`or using an Iterator, do you have any clue?

src/Set.ts
sledorze: ok done

PR #345, opened 2018-03-06
add chain to Set (Fixes #344)

### LINKED ISSUE #344: add 'chain' to Set

sledorze:


src/Set.ts
```diff
@@ -36,6 +36,16 @@ export const every = <A>(x: Set<A>, predicate: Predicate<A>): boolean => {
   return !some(x, not(predicate))
 }
 
+/** @function */
+export const chain = <A, B>(x: Set<A>, bset: Setoid<B>, f: (x: A) => Set<B>): Set<B> => {
```
gcanti: Same as `map`, shouldn't be

```ts
export const chain = <B>(S: Setoid<B>) => <A>(x: Set<A>, f: (x: A) => Set<B>): Set<B>
```

?

sledorze: Ok Done

src/Set.ts
```diff
@@ -36,6 +36,16 @@ export const every = <A>(x: Set<A>, predicate: Predicate<A>): boolean => {
   return !some(x, not(predicate))
 }
 
+/** @function */
+export const chain = <B>(bset: Setoid<B>) => <A>(x: Set<A>, f: (x: A) => Set<B>): Set<B> => {
+  let r = new Set<B>()
+  const setUnion = union(bset)
+  x.forEach(e => {
+    r = setUnion(f(e), r)
```
gcanti: I wonder, shouldn't we use mutability here for performance reasons?

sledorze: Done

PR #347, opened 2018-03-06
add partition to Set

sledorze:


src/Set.ts
```diff
@@ -60,6 +60,24 @@ export const filter = <A>(x: Set<A>, predicate: Predicate<A>): Set<A> => {
   return r
 }
 
+/** @function */
+export const partition = <A>(x: Set<A>, predicate: Predicate<A>): { true: Set<A>; false: Set<A> } => {
```
gcanti: `partition` can be derived from `partitionMap`

https://github.com/gcanti/fp-ts/blob/63c431f6a29c7518ce0c44aefde12500b43cdb56/src/Filterable.ts#L28

Maybe we could even keep the `left` and `right` labels (in the end is just a convention)

```ts
export const partition = <A>(x: Set<A>, predicate: Predicate<A>): { left: Set<A>; right: Set<A> }
```

sledorze: Ok @gcanti may rework it after `partitionMap` is merged.

gcanti: > after partitionMap is merged

Done

sledorze: @gcanti I can do it now thanks.
I still have the reservation that it's clearly less good performance wise to derive partition from partitionMap (we do not have a fusion phase in typescript).
With the change you propose, we would allocate as many additional temporary objects as there's of elements in the Set.
Are you still thinking that tradeoff is necessary?

gcanti: @sledorze right, let's do not derive

PR #351, opened 2018-03-07
Set 'difference' do less work (Fixes #350)

### LINKED ISSUE #350: Make Set 'difference' do less work

sledorze:


src/Set.ts
```diff
@@ -109,7 +109,9 @@ export const intersection = <A>(S: Setoid<A>): ((x: Set<A>, y: Set<A>) => Set<A>
  * @function
  */
 export const difference = <A>(S: Setoid<A>): ((x: Set<A>, y: Set<A>) => Set<A>) => {
-  return (x, y) => filter(y, not(member(S)(x)))
+  const has = member(S)
+  const notHas = (x: Set<A>) => not(has(x))
```
gcanti: `notHas` looks noisy

```ts
export const difference = <A>(S: Setoid<A>): ((x: Set<A>, y: Set<A>) => Set<A>) => {
  const has = member(S)
  return (x, y) => filter(y, not(has(x)))
}
```

sledorze: Ok Done!

PR #356, opened 2018-03-09
add getOrd to Array and relevant tests (Fixes 354)

sledorze:


src/Array.ts
```diff
@@ -42,6 +44,27 @@ export const getMonoid = <A = never>(): Monoid<Array<A>> => {
   }
 }
 
+export const getSetoid: <A>(S: Setoid<A>) => Setoid<A[]> = getArraySetoid
+
+export const getOrd = <T>(O: Ord<T>): Ord<Array<T>> => ({
```
gcanti: Just a nitpick: maybe we can rename the type parameter `T` to `A` (for consistency)

src/Array.ts
```diff
@@ -42,6 +44,27 @@ export const getMonoid = <A = never>(): Monoid<Array<A>> => {
...
+  ...getSetoid(O),
+  compare: (a: Array<T>, b: Array<T>): Ordering => {
```
gcanti: Also here

src/Array.ts
```diff
@@ -42,6 +44,27 @@ export const getMonoid = <A = never>(): Monoid<Array<A>> => {
...
+    const lengthOrder = ordNumber.compare(a.length, b.length)
+    if (lengthOrder !== 0) {
+      return lengthOrder
+    } else {
+      const len = a.length
```
gcanti: This can be moved up since is used in `ordNumber.compare(a.length, b.length)`

gcanti:
> the definition of Ord for foldable leads to the same output for any two arrays

what do you mean?

sledorze:
@gcanti I've checked the purescript implementation - for reference:
https://github.com/purescript/purescript-prelude/blob/master/src/Data/Ord.js

I propose to make the necessary changes to match this behaviour.
The drawback is that it is less efficient in case of array of different length (no shortcut).

@gcanti do you agree on this?

gcanti:
@sledorze I see, that's another possible order for arrays. I'm ok with that, better to add some documentation on how the algorithm works though

src/Array.ts
```diff
@@ -42,6 +44,24 @@ export const getMonoid = <A = never>(): Monoid<Array<A>> => {
   }
 }
 
+export const getSetoid: <A>(S: Setoid<A>) => Setoid<A[]> = getArraySetoid
```
gcanti: `@function` missing

src/Array.ts
```diff
@@ -42,6 +44,24 @@ export const getMonoid = <A = never>(): Monoid<Array<A>> => {
...
+
+export const getOrd = <A>(O: Ord<A>): Ord<Array<A>> => ({
```
gcanti: - `@function` missing
- could you please add some docs?

PR #360, opened 2018-03-10
357

### LINKED ISSUE #357: Flatten all fold usage for performance
@gcanti Ok with it?
To be honest I would have better like to have done that by the/a compiler(/plugin) but to my knowledge, that does not exist (yet?)

sledorze:
Flatten all (flattenable) folds (Used mainly ternary operators).
Fix getAlt in Validation (added tests)
Changed the declaration of `traverse` (Validation, These and Either) from function to values (const) in order to correctly type one of the param and avoid creating another unnecessary closure doing so (`of` function).
Further optimisation (also chain flattening) for mapOption, rights and lefts of Array.

(Fixes #357)

### DIFF
--- a/src/EitherT.ts
+++ b/src/EitherT.ts
@@ -29,7 +29,7 @@ export function chain<F extends URIS2>(F: Monad2<F>): EitherT2<F>['chain']
 export function chain<F extends URIS>(F: Monad1<F>): EitherT1<F>['chain']
 export function chain<F>(F: Monad<F>): EitherT<F>['chain']
 export function chain<F>(F: Monad<F>): EitherT<F>['chain'] {
-  return (f, fa) => F.chain(fa, e => e.fold(l => F.of(either.left(l)), a => f(a)))
+  return (f, fa) => F.chain(fa, e => (e.isLeft() ? F.of(either.left(e.value)) : f(e.value)))
 }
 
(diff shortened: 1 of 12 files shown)


src/EitherT.ts
```diff
@@ -29,7 +29,7 @@ export function chain<F extends URIS2>(F: Monad2<F>): EitherT2<F>['chain']
 export function chain<F extends URIS>(F: Monad1<F>): EitherT1<F>['chain']
 export function chain<F>(F: Monad<F>): EitherT<F>['chain']
 export function chain<F>(F: Monad<F>): EitherT<F>['chain'] {
-  return (f, fa) => F.chain(fa, e => e.fold(l => F.of(either.left(l)), a => f(a)))
+  return (f, fa) => F.chain(fa, e => (e.isLeft() ? F.of(either.left(e.value)) : f(e.value)))
```
sledorze: we may avoid an allocation in the left case by using casting; would you want it @gcanti ?

sledorze:
@gcanti who / when / where do you want we merge that back?

gcanti:
@sledorze I'm working on a PR which adds more tests for each change contained in this PR in order to help you out with this and avoid regressions

sledorze:
@gcanti done!

PR #373, opened 2018-03-23
New Feature: Array.uniq

### LINKED ISSUE #372: Get method uniq on Array
Hello, 

I'm wondering if it's possible to have a method `uniq` on Array like that:

```typescript
const a = [1, 2, 3, 3, 2, 1]

array.uniq(a, setoidNumber) 
// => [1, 2 ,3 ]
```

Thanks a lot

alex-ketch:
Closes #372 

Hi there, thanks for the great library!
Saw an opportunity for an easy contribution and thought I'd pitch in.

I tried to hazard a guess to the [`@since` tag](https://github.com/gcanti/fp-ts/compare/master...alex-ketch:master?expand=1#diff-34929952c8f2d307dfe4421553d5c9ceR741), but let me know if you'd like me to change it (or anything else).

I was going to use `toArray` from `Set`, but that function [orders the elements](https://github.com/gcanti/fp-ts/blob/master/src/Set.ts#L15) which I personally find counter-intuitive.

---

Edit:
Seems using `assert.deepStrictEqual(uniq([NaN, NaN], setoidNumber), [NaN])` fails on Travis despite succeeding locally. I've changed to check for array length instead.


src/Array.ts
```diff
@@ -734,6 +735,17 @@ export const rotate = <A>(n: number, xs: Array<A>): Array<A> => {
   }
 }
 
+/**
+ * Remove duplicate entries from an array, keeping the first found item.
+ * @function
+ * @since 1.3.0
+ */
+export const uniq = <A>(as: Array<A>, S: Setoid<A>): Array<A> => {
```
gcanti: @alex-ketch thanks for this PR

> I tried to hazard a guess to the @since tag

1.3.0 is ok

> I was going to use toArray from Set, but that function orders the elements which I personally find counter-intuitive

The reason why `Set.toArray` requires an `Ord` is that otherwise you might end up with different outputs given the same inputs

```ts
export const toArray = <A>(x: Set<A>): Array<A> => {
  const r: Array<A> = []
  x.forEach(e => r.push(e))
  // without sorting
  return r
}

const x = new Set([1, 2])
const y = new Set([2, 1])
console.log(toArray(x)) // [1, 2]
console.log(toArray(y)) // [2, 1]
```

`x = y` but `toArray(x) != toArray(y)`

**Signature**

I propose the following signature which is more aligned to the convention used so far

```ts
export const uniq = <A>(S: Setoid<A>) => (as: Array<A>): Array<A> => {
  ...
}
```

**Dependency on Set**

I know there will be some code duplication but could we avoid a dependency on the `Set` module?

sledorze: I'm concerned about the behavior regarding preserving the relative order of unique elements.

Specifically Purescript (which is the example we should follow) does preserve relative order:

https://github.com/purescript/purescript-arrays/blob/e502257ce0a89ef7bbf89b20665e6f7a491b781f/test/Test/Data/Array.purs#L313-L318

sledorze: @alex-ketch @gcanti 
I'm concerned about not preserving relative order of elements.

Also, note that purescript (that is a reference of the project) does that specifically: 
https://github.com/purescript/purescript-arrays/blob/e502257ce0a89ef7bbf89b20665e6f7a491b781f/test/Test/Data/Array.purs#L313-L318

alex-ketch: @gcanti The currying and removing dependency on `Set.fromArray` makes sense 👍

---

@sledorze You're talking about `toArray`, correct?
If so I would agree, since as far as I'm aware `Set` makes no guarantees about the order.

> x = y but toArray(x) != toArray(y)

The above is not the case, unless I'm missing something.
https://repl.it/repls/WornSubduedOs

This may be a topic for another issue though

gcanti: @alex-ketch I mean `=` and `!=` in the mathematical sense, i.e. `new Set([1, 2])` and `new Set([2, 1])` represents the same set

```
{ 1, 2 }
```

however `toArray` would output two different arrays: `[1, 2]` is not equal to `[2, 1]` based on the standard equivalence relation defined on `Array<number>` (see its [Setoid instance](https://github.com/gcanti/fp-ts/blob/78917aada5d30f177090141202a3dff7fcd0c77b/src/Array.ts#L58))

alex-ketch:
@gcanti Made the changes. Not sure what the stance is here on using `new` for side-effects, but can assign it to a const if you'd like.

src/Array.ts
```diff
@@ -736,13 +735,13 @@ export const rotate = <A>(n: number, xs: Array<A>): Array<A> => {
...
- * Remove duplicate entries from an array, keeping the first found item.
+ * Remove duplicates from an array, keeping the first occurance of an element.
  * @function
  * @since 1.3.0
  */
-export const uniq = <A>(as: Array<A>, S: Setoid<A>): Array<A> => {
+export const uniq = <A>(S: Setoid<A>) => (as: Array<A>): Array<A> => {
   const r: Array<A> = []
-  fromArray(S)(as).forEach(a => r.push(a))
+  new Set(as).forEach(a => r.push(a))
```
sledorze: @alex-ketch that's not right, you're not using the Setoid in order to check for element equality.
If you try to use non primitive based setoid in your tests, you'll see it does not behave correctly.

on top of my head (not tested)

```typescript
import {
  insert,
  member
  } from 'fp-ts/lib/Set'
import { Setoid } from 'fp-ts/lib/Setoid'



export const unique = <T>(S: Setoid<T>) => {
  const setMember = member(S)
  const insertSet = insert(S)

  return (xs: T[]): T[] => {
    let set = new Set()
    const res = []
    for (const v of xs) {
      if (!setMember(set)(v)) {
        set = insertSet(v, set)
        res.push(v)
      }
    }
    return res
  }
}
```

sledorze: Also it may be very nice to return the original array when there was no duplicates.

sledorze: Oh.. I've given it more love, I think it answers the requirements.

```typescript
export const unique = <T>(S: Setoid<T>) => {
  const setMember = member(S)

  return (xs: T[]): T[] => {
    const set = new Set()
    const l = xs.length
    let i = 0
    for (; i < l; i++) {
      const v = xs[i]
      if (setMember(set)(v)) {
        break
      } else {
        set.add(v)
      }
    }
    if (i === l) {
      return xs
    } else {
      const r = xs.slice(0, i)
      for (; i < l; i++) {
        const v = xs[i]
        if (!setMember(set)(v)) {
          set.add(v)
          r.push(v)
        }
      }
      return r
    }
  }
}
```

tests:

```typescript
    interface Foo {
      a: string
      b: number
    }
    const fooSetoid = contramap((f: Foo) => f.b, ordNumber)
    const fooA: Foo = { a: 'a', b: 1 }
    const fooB: Foo = { a: 'b', b: 1 }
    const fooC: Foo = { a: 'c', b: 2 }
    const fooD: Foo = { a: 'd', b: 2 }

    const uniqueFooArr = [fooA, fooC]
    assert.equal(unique(fooSeoid)(uniqueFooArr), uniqueFooArr, 'preserve original array')
    assert.deepEqual(unique(fooSetoid)([fooA, fooB, fooC, fooD]), [fooA, fooC])
    assert.deepEqual(unique(fooSetoid)([fooB, fooA, fooC, fooD]), [fooB, fooC])
    assert.deepEqual(unique(fooSetoid)([fooA, fooA, fooC, fooD, fooA]), [fooA, fooC])
```

sledorze: Sorry @alex-ketch it may seems bully to push you such feedback like this..
I've been too tempted.

alex-ketch: All good @sledorze :) Thanks for pointing out the non-primitive use case.
I deviated a bit from your suggestion but let me know if you have any further feedback.

I know you mentioned that you didn't want a dependency on the `Set` module @gcanti, but now it's back with the `member` function.
I think it would be a bit too much if we duplicated all `member` dependencies. Is this ok?

gcanti: @alex-ketch we could add a `member` function for arrays

```ts
/**
 * Test if a value is a member of an array
 * @function
 * @since 1.3.0
 */
export const member = <A>(S: Setoid<A>) => (as: Array<A>, a: A): boolean => {
  return findIndex(as, e => S.equals(e, a)).isSome()
}

/**
 * Remove duplicates from an array, keeping the first occurance of an element.
 * @function
 * @since 1.3.0
 */
export const uniq = <A>(S: Setoid<A>): ((as: Array<A>) => Array<A>) => {
  const memberS = member(S)
  return as => {
    const r: Array<A> = []
    const len = as.length
      const a = as[i]
      if (!memberS(r, a)) {
        r.push(a)
      }
    }
    return r.length === len ? as : r
  }
}
```

test/Array.ts
```diff
@@ -339,6 +339,22 @@ describe('Array', () => {
...
+
+    const setoidA = contramap((f: A) => f.b, ordNumber)
+    const arrA: A = { a: 'a', b: 1 }
+    const arrB: A = { a: 'b', b: 1 }
+    const arrC: A = { a: 'c', b: 2 }
+    const arrD: A = { a: 'd', b: 2 }
+    const arrUniq = [arrA, arrC]
+
+    assert.equal(uniq(setoidA)(arrUniq), arrUniq, 'Preserve original array')
+    assert.deepEqual(uniq(setoidA)([arrA, arrB, arrC, arrD]), [arrA, arrC])
```
sledorze: @alex-ketch  Can you  please ad those for specifically checking relative order ?
```typescript
    assert.deepEqual(uniq(fooSeoid)([fooA, fooC]), [fooA, fooC])
    assert.deepEqual(uniq(fooSeoid)([fooC, fooA]), [fooC, fooA])
```

src/Array.ts
```diff
@@ -741,7 +742,21 @@ export const rotate = <A>(n: number, xs: Array<A>): Array<A> => {
...
+    const a = as[i]
+    if (!isMember(a)) {
+      set.add(a)
+    }
+    i += 1
+  }
+  if (l === set.size) {
+    return as
+  }
+  set.forEach(a => r.push(a))
```
sledorze: @alex-ketch you can presize the array, it will be more efficient (create less realloc and is generally faster).
Please take a look at:
https://github.com/alex-ketch/fp-ts/blob/d59e64d9a0510dbd4481c7f0460796c63de2fb91/src/Array.ts#L537-L544

src/Array.ts
```diff
@@ -741,7 +742,21 @@ export const rotate = <A>(n: number, xs: Array<A>): Array<A> => {
  */
 export const uniq = <A>(S: Setoid<A>) => (as: Array<A>): Array<A> => {
   const r: Array<A> = []
-  new Set(as).forEach(a => r.push(a))
+  const set: Set<A> = new Set()
+  const isMember = member(S)(set)
+  const l = as.length
+  let i = 0
+  while (i < l) {
```
sledorze: If I remember well, while is slower that for loops, hence the usage of for loop in the Array.ts file when we were optimising for speed. 
(not a proof: https://cdn-images-1.medium.com/max/2000/1*0b2Yf4R7eiETephfEjE2HA.png)

gcanti: @sledorze did you see my proposal (i.e. not using `Set`)? WDYT?

sledorze: @gcanti I'm in favor of not using Set in this use case, too, to keep things lean.

src/Array.ts
sledorze: Would be nice to not have to allocate an option for each element..

alex-ketch:
Took tests for Array.member from Set.member. Let me know if you’d like me to expand on them, or have more feedback.

src/Array.ts
```diff
@@ -735,29 +734,34 @@ export const rotate = <A>(n: number, xs: Array<A>): Array<A> => {
   }
 }
 
+/**
+ * Test if a value is a member of an array
+ * @function
+ * @since 1.3.0
+ */
+export const member = <A>(S: Setoid<A>) => (as: Array<A>, a: A): boolean => {
+  return findIndex(as, e => S.equals(e, a)).isSome()
```
sledorze: This operation is very useful and can be used a lot, can we get ride of the temporary Option to reduce the extra allocation / GC pressure?

```typescript
export const member = <A>(S: Setoid<A>) => (as: Array<A>, a: A): boolean => {
  const len = as.length
  const predicate = (e: A) => S.equals(e, a)
    if (predicate(as[i])) {
      return true
    }
  }
  return false
}
```

gcanti: @sledorze agreed

gcanti:
Thanks @alex-ketch

raveclassic:
Has this been released yet? Maybe the name could be changed to the full one `unique`?

gcanti:
@raveclassic not yet, FWIW both `lodash` and `rambda` use `uniq` though

sledorze:
Some people are used to 'distinct'

PR #398, opened 2018-04-18
Add shx and update npm scripts to allow building on Windows

wmaurer:
`rm -rf` doesn't work on windows, so this PR adds [shx](https://www.npmjs.com/package/shx) as a dev dependency and uses it in the `clean` and `docs` npm scripts.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -274,6 +274,12 @@
         "is-arrayish": "0.2.1"
       }
     },
+    "es6-object-assign": {
+      "version": "1.1.0",
+      "resolved": "https://registry.npmjs.org/es6-object-assign/-/es6-object-assign-1.1.0.tgz",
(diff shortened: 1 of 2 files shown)


gcanti:
Thanks @wmaurer, I didn't know of `shx`, what about the "standard" `rimraf` instead?

wmaurer:
@gcanti `rimraf` would also work for the `rm -rf` problem.
I personally use the more "modern" `shx` because it's more flexible - it can also do `cp` and `mv`, which might be useful in the future.
If you would prefer to use `rimraf` then I'll change it.

gcanti:
> If you would prefer to use rimraf then I'll change it

Yes please, looks like a lightweight solution. We might add shx later (if we'll ever need it)

PR #400, opened 2018-04-19
feature: add getDictionarySemigroup/getDictionaryMonoid getObjectSemigroup

raveclassic:


src/Monoid.ts
```diff
@@ -126,6 +129,17 @@ export const monoidVoid: Monoid<void> = {
   empty: undefined
 }
 
+const emptyObject = {}
+
+/**
+ * @instance
+ * @since 1.4.0
+ */
+export const monoidObject: Monoid<object> = {
```
gcanti: This looks more general

```ts
const getDictionaryMonoid = <A>(S: Semigroup<A>): Monoid<{ [key: string]: A }> => {
  ...
}
```

Then

```ts
const monoidObject: Monoid<object> = getDictionaryMonoid(getLastSemigroup<any>())

declare const x1: object
declare const x2: object
const x3 = monoidObject.concat(x1, x2)
```

WDYT?

raveclassic: Isn't `getDictionaryMonoid` supposed to build monoids for dictionary _values_, not dictionaries itselves?
Anyway I noticed some room for improvement - `semigroupObject`/`monoidObject` should better be replaced with factories accepting type argument with default value `never` like it's done for `getArraySemigroup`/`getArrayMonoid`

gcanti: > Isn't getDictionaryMonoid supposed to build monoids for dictionary *values*, not dictionaries itselves?

Not sure what you mean.

Here's an example

```ts
const d1: { [key: string]: number } = { foo: 1, bar: 2 }
const d2: { [key: string]: number } = { bar: 3, baz: 4 }

const M = getDictionaryMonoid(semigroupSum)

console.log(M.concat(d1, d2))
// { foo: 1, bar: 5, baz: 4 }
```

raveclassic: `getDictionaryMonoid(semigroupSum)` here you specify semigroup for dictionary values, but I don't need one. I need to concat `object`s using `Object.assign`, not values in of their keys.

```typescript
const foo = {
  foo: 123,
  bar: '000'
}
const bar = {
  bar: '123'
}
const monoidObject = getObjectMonoid<object>()
const result = fold(monoidObject)([foo, bar]) // { foo: 123, bar: '123' }
const expected = Object.assign({}, foo, bar)
assert.deepEqual(result, expected)
```

I've update the code replacing constants with functions.

gcanti: ok but what I mean is that `Object.assign` behaves like a `Last` semigroup on values (i.e. the second argument of `concat` wins)

```ts
const foo = {
  foo: 123,
  bar: '000'
}
const bar = {
  bar: '123'
}
const monoidObject: Monoid<object> = getDictionaryMonoid(getLastSemigroup<any>())
const result = fold(monoidObject)([foo, bar]) // { foo: 123, bar: '123' }
```

`getDictionaryMonoid` looks useful and more general (`monoidObject`, as you can see, is just a special case).
With the same abstraction I can handle other cases than `Object.assign`, like mine above

```ts
const d1: { [key: string]: number } = { foo: 1, bar: 2 }
const d2: { [key: string]: number } = { bar: 3, baz: 4 }

const M = getDictionaryMonoid(semigroupSum)

console.log(M.concat(d1, d2))
// { foo: 1, bar: 5, baz: 4 }
```

gcanti: Btw here's my quick and dirty implementation to test the snippets I wrote

```ts
const getDictionaryMonoid = <A>(S: Semigroup<A>): Monoid<{ [key: string]: A }> => {
  return {
    concat: (x, y) => {
      const r: { [key: string]: A } = { ...x }
      for (let k in y) {
        if (x.hasOwnProperty(k)) {
          r[k] = S.concat(x[k], y[k])
        } else {
          r[k] = y[k]
        }
      }
      return r
    },
    empty: {}
  }
}
```

raveclassic: But `getLastSemigroup` ignores the first value if last is undefined. So if second object does not have a key from the first one, it will be ignored. This does not align with `Object.assign` which collects everything available.
```typescript
export const getLastSemigroup = <A = never>(): Semigroup<A> => {
  return { concat: (_, y) => y }
}
```
I understand that I can build semigroup which picks either last or first (if last is undefined) but, anyway, key iteration does not look performant.

raveclassic: Looks good, although key iteration is less performant as straight assignments (I guess)

gcanti: This should be faster than my previous version

```ts
const getDictionaryMonoid = <A>(S: Semigroup<A>): Monoid<{ [key: string]: A }> => {
  return {
    concat: (x, y) => {
      const r: { [key: string]: A } = { ...x }
      const keys = Object.keys(y)
      const len = keys.length
        const k = keys[i]
        r[k] = x.hasOwnProperty(k) ? S.concat(x[k], y[k]) : y[k]
      }
      return r
    },
    empty: {}
  }
}
```

/cc @sledorze

gcanti: @raveclassic I wrote a benchmark (getDictionaryMonoid vs Object.assign)

**Code**

```js
const Benchmark = require('benchmark')

const { getLastSemigroup } = require('../lib/Semigroup')

const getDictionaryMonoid = S => {
  return {
    concat: (x, y) => {
      const r = { ...x }
      const keys = Object.keys(y)
      const len = keys.length
        const k = keys[i]
        r[k] = x.hasOwnProperty(k) ? S.concat(x[k], y[k]) : y[k]
      }
      return r
    },
    empty: {}
  }
}

const monoidObject1 = getDictionaryMonoid(getLastSemigroup())

const monoidObject2 = {
  concat: (x, y) => Object.assign({}, x, y),
  empty: {}
}

const foo = {
  foo: 123,
  bar: '000'
}
const bar = {
  bar: '123'
}

suite
  .add('monoidObject1 (getDictionaryMonoid)', function() {
    monoidObject1.concat(foo, bar)
  })
  .add('monoidObject2 (Object.assign)', function() {
    monoidObject2.concat(foo, bar)
  })
  .on('cycle', function(event) {
  })
  .on('complete', function() {
  })
  .run({ async: true })
```

**Results**

```
monoidObject1 (getDictionaryMonoid) x 2,288,891 ops/sec ±0.73% (88 runs sampled)
monoidObject2 (Object.assign) x 2,140,533 ops/sec ±0.46% (89 runs sampled)
Fastest is monoidObject1 (getDictionaryMonoid)
```

raveclassic: ok you got me :) 
V8 is such a weird thing...

raveclassic: What about `getObjectSemigroup`? Is it worth adding?

gcanti: Not a fan of duplicated APIs in `Semigroup` and `Monoid` sharing the same constraint (in this case `Semigroup<A>`)

raveclassic: What do you think about moving `concat` to `getDictionarySemigroup` accepting dictionary value semigroup? Then `getDictionaryMonoid` could just pass accepted semigroup to `getDictionarySemigroup`?

```typescript
type Dictinonary = { [key: string]: A } 
declare const getDictionarySemigroup = <A>(S: Semigroup<A>): Semigroup<Dictionary<A>>;
const getDictionaryMonoid = <A>(S: Semigroup<A>): Monoid<Dictionary<A>> {
  return {
    ..getDictionarySemigroup(S),
    empty: {}
  }
}
```

raveclassic: Just to better align with FL-spec, concat goes from semigroup, empty - from monoid

raveclassic: Some more thoughts here - `getDictionaryMonoid/getDictionarySemigroup` destroy the type of objects they concatenate:
```typescript
import { getLastSemigroup, Semigroup } from './Semigroup'
import { fold, Monoid } from './Monoid'

type Dictionary<A> = { [key: string]: A }

declare const getDictionarySemigroup: <A>(S: Semigroup<A>) => Semigroup<Dictionary<A>>

const getDictionaryMonoid = <A>(S: Semigroup<A>): Monoid<Dictionary<A>> => {
    return {
        ...getDictionarySemigroup(S),
        empty: {}
    }
}

type Config = {
  foo?: number
  bar: string
}

const configs: Config[] = [
  {
    foo: 123,
    bar: '123'
  },
  {
    bar: '231'
  }
]

const monoidConfig = getDictionaryMonoid(getLastSemigroup<number | string | undefined>())

const result = fold(monoidConfig)(configs)
const test = result.bar // any, the type is lost

```

raveclassic: @gcanti The problem is that `bar` is `string`, not `string | number | undefined` :)

gcanti: > bar is expected to be string

No, it isn't, since you are working with dictionaries (there's `getRecordSemigroup` / `getRecordMonoid` for that)

```ts
// monoidConfig: Monoid<Dictionary<string | number | undefined>>
const monoidConfig = getDictionaryMonoid(getLastSemigroup<number | string | undefined>())
```

That's how TypeScript works with index signatures

```ts
declare const x: Config
const y: Dictionary<string | number | undefined> = x
y.bar // string | number | undefined
```

Anyway this PR is about `Monoid<object>`

raveclassic: This PR is about folding lists of _objects_ of the same type (with explicitly defined interface) using `Object.assign` as binary operation. `Object.assign` preserves types of objects it operates on, and moreover it does not treat objects as key-value dictionaries. I'd like to stress it.
```typescript
type Config = {
    foo?: number
    bar: string
}

const configs: Config[] = [{ foo: 123, bar: '123' }, { bar: '456' }]
const res = Object.assign({}, configs[0], configs[1])
res.bar // string
```

On the other hand `getDictionaryMonoid` using any Semigroup instance for dictionary values _does_ destroy original type of objects, converting them to dictionaries where type of dictionary value is a union of all possible value types in original objects. This _is not_ the desired behavior, but it _does_ make sense for other usecases (concatenation of dictionary values).

gcanti: > This PR is about folding lists of objects of the same type (with explicitly defined interface)

I see, based on your original code and tests I thought we were talking about `Monoid<object>`. 

So `getDictionaryMonoid`, despite being a nice generalization of `Monoid<object>`, solves another user case (namely `Dictionary<A>`) than `getObjectMonoid` (as described above).

I guess we could add them both

```ts
export const getDictionaryMonoid = <A>(S: Semigroup<A>): Monoid<Dictionary<A>> => {
  return {
    concat: (x, y) => {
      ...
    },
    empty: {}
  }
}

const monoidObject = getDictionaryMonoid(getLastSemigroup<any>())

export const getObjectMonoid = <A extends object = never>(): Monoid<A> => monoidObject as any
```

raveclassic: Nice :) So what about splitting Monoid to Monoid and Semigroup parts? Setoid/Ord (and others) are splitted though you could get full ord each time :)

gcanti: In this particular case I don't see any value in splitting the `Semigroup` and `Monoid` factories, since the constraint is the same. 

However if you prefer to split them, feel free to do it so.

raveclassic:
@gcanti Please, review

src/Monoid.ts
```diff
@@ -88,6 +92,35 @@ export const getArrayMonoid = <A = never>(): Monoid<Array<A>> => {
...
+ * Gets {@link Monoid} instance for objects of given type preserving their type
+ * @function
+ * @since 1.4.0
+ * @example
+ * const M = getObjectMonoid<{ foo: number }>()
+ * const result = fold(M)([{ foo: 123 }, { foo: 456 }]) // { foo: 456 }
+ */
+export const getObjectMonoid = <A extends object = never>(): Monoid<A> => ({
+  ...getObjectSemigroup<A>(),
+  empty: emptyObject as A
```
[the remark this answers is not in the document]
raveclassic: Agree

src/Semigroup.ts
```diff
@@ -127,6 +127,41 @@ export const getArraySemigroup = <A = never>(): Semigroup<Array<A>> => {
...
+/**
+ * Gets {@link Semigroup} instance for objects of given type preserving their type
+ * @function
+ * @since 1.4.0
+ * @example
+ * const S = getObjectSemigroup<{ foo: number }>()
+ * const result = S.concat({ foo: 123 }, { foo: 456 }) // { foo: 456 }
+ */
+export const getObjectSemigroup = <A extends object = never>(): Semigroup<A> =>
+  getDictionarySemigroup(getLastSemigroup()) as any
```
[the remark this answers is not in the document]
raveclassic: yep, nice

PR #403, opened 2018-04-20
StrMap#filter function added

mlegenhausen:


src/StrMap.ts
```diff
@@ -52,6 +52,16 @@ export class StrMap<A> {
     }
     return out
   }
+  filter(p: Predicate<A>): StrMap<A> {
```
gcanti: tag `@since` is missing

src/StrMap.ts
```diff
@@ -52,6 +52,16 @@ export class StrMap<A> {
...
+    const o = this.value
+    const r: { [key: string]: A } = {}
+    for (const k in o) {
+      if (p(o[k])) {
```
gcanti: we can avoid duplication here with a `const value = o[k]`

src/StrMap.ts
```diff
@@ -272,6 +282,14 @@ export const pop = <A>(k: string, d: StrMap<A>): Option<[A, StrMap<A>]> => {
   return a.isNone() ? none : some(tuple(a.value, remove(k, d)))
 }
 
+/**
+ * @function
+ * @since 1.4.0
+ */
+export const filter = <A>(fa: StrMap<A>, p: Predicate<A>): StrMap<A> => {
```
gcanti: since there is a `filter` method, we can avoid exporting a static function here, WDYT?

mlegenhausen: I am fine with removing it, so do `Option` and `Either` also not reexporting it.

gcanti:
Thanks @mlegenhausen

PR #409, opened 2018-04-26
Internal: use getObjectSemigroup in StrMap.concat

gcanti:


src/StrMap.ts
```diff
@@ -70,8 +71,10 @@ export class StrMap<A> {
 
 const empty: StrMap<never> = new StrMap({})
 
+const semigroupAnyObject = getObjectSemigroup<any>()
+
 const concat = <A>(x: StrMap<A>, y: StrMap<A>): StrMap<A> => {
-  return new StrMap(Object.assign({}, x.value, y.value))
+  return new StrMap(semigroupAnyObject.concat(x.value, y.value))
```
raveclassic: Has come in handy :)

PR #410, opened 2018-04-26
Docs: add ReaderTaskEither instance to mtl example

gcanti:


examples/mtl.ts
```diff
@@ -35,12 +50,12 @@ function likePost<M>(M: Monad<M>, U: MonadUser<M>, F: MonadFB<M>): (token: strin
 
 import { URI as IOURI, io, IO } from 'fp-ts/lib/IO'
 
-const monadUserIO: MonadUser<IOURI> = {
+const monadUserIO: MonadUser1<IOURI> = {
```
RPallas92: Why IO and Task use Monad1 and ReaderTaskEither uses Monad instead of Monad3?

PR #413, opened 2018-05-01
New Feature: add Lattice typeclass hierarchy, closes #412

### LINKED ISSUE #412: Lattice typeclass hierarchy
It would be useful to have a lattice typeclass hierarchy similar to e.g. https://github.com/typelevel/algebra/tree/master/core/src/main/scala/algebra/lattice 

I am porting some data structures from scala to ts that need a lattice hierarchy.

Feature

Lattice typeclass hierarchy does not exist

**If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem via https://codesandbox.io/ or similar.**

Lattice typeclass hierarchy should exist in fp-ts

n/a

gcanti:
@rklaehn could you please review?

### DIFF
--- a/src/BooleanAlgebra.ts
+++ b/src/BooleanAlgebra.ts
@@ -1,0 +1,73 @@
+import { HeytingAlgebra } from './HeytingAlgebra'
+
+/**
+ * Boolean algebras are Heyting algebras with the additional constraint that the law of the excluded middle is true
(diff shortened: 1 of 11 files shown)


rklaehn:
Thanks, that was fast! Looks good as far as I can see. (I just compared to the typelevel algebra impl).

Issue #421, opened 2018-05-08
Question: Generalization of combine


raveclassic:
Currently we can use `sequence(someApplicativeInstance, array)` to turn a list of that applicatives inside out. But its signature requires inner values of that applicatives to be of the same type, so that you can't write:
```typescript
sequence(option, array)([some(1), some('2')]);
sequence(either, array)([right(1), right('2')]);
sequence(remoteData, array)([success('123'), success(123)]);
```
The only way is to write [a bunch of overloads](https://github.com/gcanti/fp-ts/issues/322#issuecomment-366228609).

So the question is if we can generalize it for a list of any applicatives without touching `sequence`? The closest HKT I've found is `Monoidal` but it's limited only to two arguments. (besides I think it may be  theoretically incorrect to use it, I'm not very strong with the theory).
Another way would be to add `combine` to `Array` module generalizing to `Applicative` and adding those overloads there. But it will require to add 1-10 overloads for each HKT, HKT2 etc.

UPD: Referencing Monoidal I mean I don't know if it is legal to change mult to accept a spread or a list instead of only two args.

raveclassic:
@gcanti Any ideas? I'd like to know opinion on this

gcanti:
I didn't comment because I have no smart ideas besides the bunch of overloadings thing.

Theoretically the abstraction which allows you to combine effects with multiple types is `Applicative`, via `liftAn` functions or equivalent.
`Monoidal` is an alternative and equivalent definition of `Applicative`

raveclassic:
So what do you think fits best? Adding separate `combine` to `Array` module with nesessary overloads or changing `Monoidal`'s `mult` to accept a spread or a list of values instead of only two arguments? Or is it out of scope of `fp-ts` at all?

gcanti:
I think that `combine` should be added to `Apply`. 

Its implementation could use `liftAn`

```ts
interface Combine<F> {
  <A, B>(fa: HKT<F, A>, fb: HKT<F, B>): HKT<F, [A, B]>
  <A, B, C>(fa: HKT<F, A>, fb: HKT<F, B>, fc: HKT<F, C>): HKT<F, [A, B, C]>
  <A, B, C, D>(fa: HKT<F, A>, fb: HKT<F, B>, fc: HKT<F, C>, fd: HKT<F, D>): HKT<F, [A, B, C, D]>
}

interface Combine1<F extends URIS> {
[...]

PR #425, opened 2018-05-10
Feature/apply sequenceT

raveclassic:
#421


src/Apply.ts
```diff
@@ -246,3 +246,107 @@ export function getSemigroup<F, A>(F: Apply<F>, S: Semigroup<A>): () => Semigrou
...
+ * @param F - {@link Apply} instance
+ */
+export function sequenceT<F extends URIS3, U, L>(F: Apply3<F>): SequenceT3<F>
+export function sequenceT<F extends URIS3, U, L>(F: Apply3C<F, U, L>): SequenceT3C<F, U, L>
+export function sequenceT<F extends URIS2>(F: Apply2<F>): SequenceT2<F>
+export function sequenceT<F extends URIS2, L>(F: Apply2C<F, L>): SequenceT2C<F, L>
+export function sequenceT<F extends URIS>(F: Apply1<F>): SequenceT1<F>
+export function sequenceT<F>(F: Apply<F>): SequenceT<F> {
+  return (...list: any[]) => {
+    const args = list.reverse()
```
gcanti: Ops sorry, I didn't notice that my results were reversed. We can avoid `list.reverse()` by changing the order in the `for` loop

```ts
export function sequenceT<F>(F: Apply<F>): SequenceT<F> {
  return (...args: any[]) => {
    const len = args.length
    let f = tupleConstructors[len]
    if (!Boolean(f)) {
      f = tupleConstructors[len] = curried((...args: any[]): any[] => args, len - 1, [])
    }
    let r = F.map(args[0], f)
      r = F.ap(r, args[i])
    }
    return r
  }
}
```

src/Apply.ts
```diff
@@ -246,3 +246,107 @@ export function getSemigroup<F, A>(F: Apply<F>, S: Semigroup<A>): () => Semigrou
...
+ * @function
+ * @since 1.5.0
+ * @param F - {@link Apply} instance
+ */
+export function sequenceT<F extends URIS3, U, L>(F: Apply3<F>): SequenceT3<F>
+export function sequenceT<F extends URIS3, U, L>(F: Apply3C<F, U, L>): SequenceT3C<F, U, L>
+export function sequenceT<F extends URIS2>(F: Apply2<F>): SequenceT2<F>
+export function sequenceT<F extends URIS2, L>(F: Apply2C<F, L>): SequenceT2C<F, L>
+export function sequenceT<F extends URIS>(F: Apply1<F>): SequenceT1<F>
+export function sequenceT<F>(F: Apply<F>): SequenceT<F> {
```
gcanti: We must also export the base signature here

```ts
export function sequenceT<F>(F: Apply<F>): SequenceT<F>
```

src/Apply.ts
raveclassic: nice, done

src/Apply.ts
raveclassic: You mean to duplicate it? ok, done

gcanti:
@raveclassic nice, thanks. Could you please squash the commits while merging?

PR #431, opened 2018-05-15
Use a Semigroup instead of a custom merge function for StrMap fromFoldable

mlegenhausen:
Breaking Change!


gcanti:
> Usecase?

Any binary closed operation.

I'm afraid I can't accept this PR: it contains a breaking change and makes `fromFoldable` less general.

Issue #433, opened 2018-05-16
IOEither


leemhenson:
So, given all the excitement around scalaz 8's bifunctor io i got to thinking "it would be nice to have an equivalent of TaskEither for IO, so you can do error-handling for impure non-async code". Would that be something that was allowed/lawful?

leemhenson:
...obviously what I mean is "do error-handling for impure non-async code in a more ergonomic way than `IO<Either<L, R>>`. 😅

leemhenson:
Mainly the reason is that using Task to wrap non-async code forces clients
to deal with promises where they didn’t before. It’s not a big deal, but
it’s irritating. Also, there are some admittedly minor perf considerations
about adding promises where there don’t need to be any.

sledorze:
Oh! that made me realise we have different definition of IO with funfix (in funfix Land, IO is Async):

https://funfix.org/api/effect/classes/io.html

I acknowledge its very late for both libraries to discuss it but I wonder who's on the wrong naming side (if any).

PR #437, opened 2018-05-17
New Feature: IOEither

leemhenson:
See https://github.com/gcanti/fp-ts/issues/433

I pretty much just `s/Task/IO/` across `TaskEither.ts` 😅


leemhenson:
Done - I went with the idea that everything is new in this file so no need to have version comments except on the `class`?

gcanti:
Thanks @leemhenson

PR #438, opened 2018-05-17
Feature/orElse

raveclassic:
Adds lazy alt (orElse) to Option, Either, Identity.
It's often useful to pass some heavy computation as alternative to missing value. `orElse` allows to pass a thunk which will be executed on failure

### DIFF
--- a/src/Either.ts
+++ b/src/Either.ts
@@ -85,13 +85,13 @@ export class Left<L, A> {
   /**
    * Lazy version of {@link alt}
    * @since 1.6.0
-   * @param {Lazy<Either<L, A>>} fy - thunk
+   * @param {(l: L) => Either<L, A>} fy - thunk
    * @example
-   * assert.deepEqual(right(1).altL(() => right(2)), right(1))
+   * assert.deepEqual(right(1).orElse(() => right(2)), right(1))
    * @returns {Either<L, A>}
    */
-  altL(fy: Lazy<Either<L, A>>): Either<L, A> {
-    return fy()
+  orElse(fy: (l: L) => Either<L, A>): Either<L, A> {
+    return fy(this.value)
   }
   extend<B>(f: (ea: Either<L, A>) => B): Either<L, B> {
     return this as any
@@ -184,12 +184,12 @@ export class Right<L, A> {
   /**
    * Lazy version of {@link alt}
    * @since 1.6.0
-   * @param {Lazy<Either<L, A>>} fy - thunk
+   * @param {(l: L) => Either<L, A>} fy - thunk
    * @example
-   * assert.deepEqual(right(1).altL(() => right(2)), right(1))
+   * assert.deepEqual(right(1).orElse(() => right(2)), right(1))
    * @returns {Either<L, A>}
    */
-  altL(fy: Lazy<Either<L, A>>): Either<L, A> {
+  orElse(fy: (l: L) => Either<L, A>): Either<L, A> {
     return this
   }
   extend<B>(f: (ea: Either<L, A>) => B): Either<L, B> {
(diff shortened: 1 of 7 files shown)


src/Identity.ts
```diff
@@ -47,6 +47,19 @@ export class Identity<A> {
...
+  /**
+   * Lazy version of {@link alt}
+   * @since 1.6.0
+   * @param {Lazy<Identity<A>>} fx - thunk
+   * @example
+   * const a = new Identity(1)
+   * assert.deepEqual(a.altL(() => new Identity(2)), a)
+   * @returns {Identity<A>}
+   */
+  altL(fx: Lazy<Identity<A>>): Identity<A> {
```
gcanti: in case of `Identity` is `altL` actually useful?

src/Either.ts
```diff
@@ -81,6 +81,18 @@ export class Left<L, A> {
...
+
+  /**
+   * Lazy version of {@link alt}
+   * @since 1.6.0
+   * @param {Lazy<Either<L, A>>} fy - thunk
+   * @example
+   * assert.deepEqual(right(1).altL(() => right(2)), right(1))
+   * @returns {Either<L, A>}
+   */
+  altL(fy: Lazy<Either<L, A>>): Either<L, A> {
```
gcanti: In case of `Either` I'd prefer a more general `orElse` method instead of `altL`

```ts
orElse<M>(f: (l: L) => Either<M, A>): Either<M, A> {
  ...
}
```

see `TaskEither.prototype.orElse`

src/Identity.ts
raveclassic: well, I just added it just because `alt` already exists

src/Either.ts
raveclassic: Should all `altL`s be renamed to `orElse`? Or only this because of `l: L` arg?

src/Identity.ts
gcanti: Never mind, please add a test though

src/Either.ts
gcanti: Good idea, let's rename all `altL` to `orElse`

src/Identity.ts
raveclassic: done

src/Either.ts
```diff
@@ -169,6 +181,17 @@ export class Right<L, A> {
...
   }
+  /**
+   * Lazy version of {@link alt}
+   * @since 1.6.0
+   * @param {(l: L) => Either<L, A>} fy - thunk
+   * @example
+   * assert.deepEqual(right(1).orElse(() => right(2)), right(1))
+   * @returns {Either<L, A>}
+   */
+  orElse(fy: (l: L) => Either<L, A>): Either<L, A> {
```
gcanti: As previously suggested, we could make this signature more general

```ts
orElse<M>(fy: (l: L) => Either<M, A>): Either<M, A>
```

raveclassic: Sorry, missed that. Will fix

raveclassic: @gcanti What about the `Right` side? It should return `this` where type of left side is `L`, not `M`.

gcanti: @raveclassic If the value is a `Right<L, A>` then is also a `Right<M, A>` for all `M`. You can return `this as any` (much like `map` in the `Left` case)

src/Either.ts
```diff
@@ -169,6 +181,17 @@ export class Right<L, A> {
   alt(fy: Either<L, A>): Either<L, A> {
     return this
   }
+  /**
+   * Lazy version of {@link alt}
```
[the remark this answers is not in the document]
gcanti: Ah! I use VS Code and just works. So you don't see any of the other comments? Say for `fold` or `getOrElse`, etc... ?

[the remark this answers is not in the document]
raveclassic: What the?... It's a kind of magic, I retried and it worked!

gcanti:
Thanks @raveclassic

PR #447, opened 2018-05-21
Docs: handle `example` tag

gcanti:


raveclassic:
Awesome change! 
However, note that links to methods are not resolved. https://github.com/gcanti/fp-ts/blob/master/docs/api/md/NonEmptyArray.md#ap_

gcanti:
@raveclassic https://github.com/gcanti/fp-ts/pull/449

PR #451, opened 2018-05-22
Add mapLeft to ReaderTaskEither

leemhenson:


src/ReaderTaskEither.ts
```diff
@@ -80,6 +86,14 @@ const map = <E, L, A, B>(fa: ReaderTaskEither<E, L, A>, f: (a: A) => B): ReaderT
   return fa.map(f)
 }
 
+/**
+ * @function
+ * @since 1.6.0
+ */
+export const mapLeft = <E, L, A, M>(fa: ReaderTaskEither<E, L, A>, f: (l: L) => M): ReaderTaskEither<E, M, A> => {
```
gcanti: There's no need for this static function since there's a corresponding method

PR #455, opened 2018-05-23
Handle undefined errors in callback of TaskEither taskify

dmechas:
When wrapping a function from elasticsearch client the callback is called with `undefined` as first argument preventing taskify to return a right and instead returns left

### DIFF
--- a/src/TaskEither.ts
+++ b/src/TaskEither.ts
@@ -219,7 +219,7 @@ export function taskify<L, R>(f: Function): () => TaskEither<L, R> {
       new Task(
         () =>
           new Promise(resolve => {
-            args.push((e: L, r: R) => (e !== null ? resolve(eitherLeft<L, R>(e)) : resolve(eitherRight<L, R>(r))))
+            args.push((e: L, r: R) => (e != null ? resolve(eitherLeft<L, R>(e)) : resolve(eitherRight<L, R>(r))))
             f.apply(null, args)
           })
       )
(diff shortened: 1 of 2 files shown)


gcanti:
Thanks @dmechas. There's another bug though

```ts
export declare function taskify<A, L, R>(f: (a: A, cb: (e: L, r: R) => void) => void): (a: A) => TaskEither<L, R>
```

should be

```ts
export declare function taskify<A, L, R>(
  f: (a: A, cb: (e: L | null | undefined, r?: R) => void) => void
): (a: A) => TaskEither<L, R>
```

otherwise you get

```ts
const api1 = (path: string, callback: (err: Error | null | undefined, result?: string) => void): void => {
  callback(undefined, 'ok')
}

// const f1: (a: string) => TaskEither<Error | null | undefined, string | undefined>
const f1 = taskify(api1)
```

instead of 

```ts
const f1: (a: string) => TaskEither<Error, string>
```

src/TaskEither.ts
```diff
@@ -174,17 +174,21 @@ export const tryCatch = <L, A>(f: Lazy<Promise<A>>, onrejected: (reason: {}) =>
 }
 
 /** Convert a node style callback function to one returning a `TaskEither` */
-export function taskify<L, R>(f: (cb: (e: L, r: R) => void) => void): () => TaskEither<L, R>
-export function taskify<A, L, R>(f: (a: A, cb: (e: L, r: R) => void) => void): (a: A) => TaskEither<L, R>
-export function taskify<A, B, L, R>(f: (a: A, b: B, cb: (e: L, r: R) => void) => void): (a: A, b: B) => TaskEither<L, R>
+export function taskify<L, R>(f: (cb: (e: L | null | undefined, r: R) => void) => void): () => TaskEither<L, R>
```
gcanti: `r` is still required, I think we should change it to `r?: R`

dmechas: Don't you think we should close this and open a new PR to fix that?

PR #462, opened 2018-05-24
Either: add functions based on refinements

gcanti:
@leemhenson @raveclassic do you have anything you want to add? Otherwise I'm going to release 1.6 after merging this


raveclassic:
Yeah, let's release! :)

PR #466, opened 2018-05-28
Polish: Reader.local, ReaderTaskEither.local should be able to change…

gcanti:
… the environment type

/cc @tbrisbane @raveclassic

I would release this as 1.6.1


raveclassic:
What do you think about adding `local` to `Reader.prototype`? This would be super useful

PR #488, opened 2018-06-20
Feature/compactable instances

raveclassic:
Adds instances of `Compactable` for `Array`, `Set`, `Option`, `StrMap` and `getCompactable` helper to `Either` and `Validation`

### DIFF
--- a/src/Option.ts
+++ b/src/Option.ts
@@ -1,18 +1,28 @@
 import { Alternative1 } from './Alternative'
 import { Applicative } from './Applicative'
-import { Either } from './Either'
+import { Either, fromOption } from './Either'
 import { Extend1 } from './Extend'
 import { Foldable1 } from './Foldable'
-import { HKT } from './HKT'
+import { HKT, URIS, URIS2, URIS3 } from './HKT'
 import { Monad1 } from './Monad'
-import { Monoid, getDualMonoid } from './Monoid'
+import { getDualMonoid, Monoid } from './Monoid'
 import { Ord } from './Ord'
 import { Plus1 } from './Plus'
 import { Semigroup } from './Semigroup'
 import { Setoid } from './Setoid'
 import { Traversable1 } from './Traversable'
 import { identity, Lazy, Predicate, Refinement, toString } from './function'
-import { Compactable1, separated, Separated } from './Compactable'
+import {
+  Compactable,
+  Compactable1,
+  Compactable2,
+  Compactable2C,
+  Compactable3,
+  Compactable3C,
+  separated,
+  Separated
+} from './Compactable'
+import { Functor, Functor1, Functor2, Functor2C, Functor3, Functor3C } from './Functor'
 
 declare module './HKT' {
   interface URI2HKT<A> {
@@ -554,6 +564,31 @@ export const isNone = <A>(fa: Option<A>): fa is None<A> => {
  */
 export const fromRefinement = <A, B extends A>(refinement: Refinement<A, B>) => (a: A): Option<B> => {
   return refinement(a) ? some(a) : none
+}
+
+/**
+ * Gets default implementation of {@link Compactable.compact} using {@link Compactable.separate}
+ * @function
+ * @since 1.7.0
+ */
+export function compactDefault<F extends URIS3, U, L>(
+  F: Functor3C<F, U, L> & Pick<Compactable3C<F, U, L>, 'separate'>
+): Compactable3C<F, U, L>['compact']
+export function compactDefault<F extends URIS3>(
+  F: Functor3<F> & Pick<Compactable3<F>, 'separate'>
+): Compactable3<F>['compact']
+export function compactDefault<F extends URIS2, L>(
+  F: Functor2C<F, L> & Pick<Compactable2C<F, L>, 'separate'>
+): Compactable2C<F, L>['compact']
+export function compactDefault<F extends URIS2>(
+  F: Functor2<F> & Pick<Compactable2<F>, 'separate'>
+): Compactable2<F>['compact']
+export function compactDefault<F extends URIS>(
+  F: Functor1<F> & Pick<Compactable1<F>, 'separate'>
+): Compactable1<F>['compact']
+export function compactDefault<F>(F: Functor<F> & Pick<Compactable<F>, 'separate'>): Compactable<F>['compact']
+export function compactDefault<F>(F: Functor<F> & Pick<Compactable<F>, 'separate'>): Compactable<F>['compact'] {
+  return foa => F.separate(F.map(foa, fromOption(null))).right
 }
 
 const compact = <A>(fa: Option<Option<A>>): Option<A> => fa.chain(identity)
--- a/src/Option.ts
+++ b/src/Option.ts
@@ -1,9 +1,9 @@
 import { Alternative1 } from './Alternative'
 import { Applicative } from './Applicative'
-import { Either, fromOption } from './Either'
+import { Either } from './Either'
 import { Extend1 } from './Extend'
 import { Foldable1 } from './Foldable'
-import { HKT, URIS, URIS2, URIS3 } from './HKT'
+import { HKT } from './HKT'
 import { Monad1 } from './Monad'
 import { getDualMonoid, Monoid } from './Monoid'
 import { Ord } from './Ord'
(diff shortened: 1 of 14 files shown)


gcanti:
@raveclassic could you please rebase on master rather than merging

raveclassic:
@gcanti Fixed circular reference by moving `compactDefault` to `Option.ts` and `separateDefault` to `Either.ts`. That's the only way I think because these helpers require `Option` and `Either` classes. Since these helpers are likely to be autoimported most of the time by IDE/Editor I think it's ok to store them in `Option.ts` and `Either.ts` modules, what do you think?

gcanti:
@raveclassic looks like the build was failing due to bad formatting of `test/Validation.ts`

```
...
> fp-ts@1.6.2 prettier /home/travis/build/gcanti/fp-ts
> prettier --no-semi --single-quote --print-width 120 --parser typescript --list-different "{src,test,examples,exercises}/**/*.ts"
test/Validation.ts
npm ERR! code ELIFECYCLE
npm ERR! errno 1
npm ERR! fp-ts@1.6.2 prettier: `prettier --no-semi --single-quote --print-width 120 --parser typescript --list-different "
...
```

(from https://travis-ci.org/gcanti/fp-ts/builds/394664387?utm_source=github_status&utm_medium=notification)

Anyway, isn't the circular dependencies just caused by importing the helper `separated` in `src/Either.ts` and `src/Option.ts`? If that's the case
you could just build a `Separated` by hand wihtout using the helper so that all the importa are just types

raveclassic:
@gcanti You're right! 
I moved defaults back to Compactable and all tests even with circular reference. I think it's working because `separated` isn't accessed on module-level but in runtime so it's able to be resolved correctly.

gcanti:
@raveclassic ok but there's still an issue with `separated`: every module that imports that small helper is now also importing the whole module `Option` and the whole module `Either` (via `Compactable`)

So far I carefully tried to import only types unless strictly necessary. Also I don't like having runtime circular dependencies, even if it seems to not cause issues (for now..)

Do we really need that helper? I don't see inference problems for the type checker (as opposed to tuple types, hence the `tuple` helper).
Can we at least move it into the `function` module?

raveclassic:
@gcanti Fair enough. 

>Do we really need that helper? I don't see inference problems for the type checker (as opposed to tuple types, hence the tuple helper).

Yeah I think it can be removed because the only reason to add it was brevity.

src/Set.ts
```diff
@@ -182,21 +204,24 @@ export const intersection = <A>(S: Setoid<A>): ((x: Set<A>, y: Set<A>) => Set<A>
  * @function
  * @since 1.2.0
  */
-export const partitionMap = <A, L, R>(x: Set<A>, f: (a: A) => Either<L, R>): { left: Set<L>; right: Set<R> } => {
+export const partitionMap = <A, L, R>(x: Set<A>, f: (a: A) => Either<L, R>): Separated<Set<L>, Set<R>> => {
```
gcanti: this is not related to this PR but IMO this signature is wrong, should be 

```ts
export const partitionMap = <L, R>(SL: Setoid<L>, SR: Setoid<R>) => <A>(x: Set<A>, f: (a: A) => Either<L, R>): Separated<Set<L>, Set<R>>
```

src/Set.ts
```diff
@@ -292,3 +317,46 @@ export const fromArray = <A>(S: Setoid<A>) => (as: A[]): Set<A> => {
   }
   return r
 }
+
+const compact = <A>(fa: Set<Option<A>>): Set<A> => {
```
gcanti: Should be

```ts
const compact = <A>(S: Setoid<A>) => (fa: Set<Option<A>>): Set<A> =>
```

src/Set.ts
```diff
@@ -292,3 +317,46 @@ export const fromArray = <A>(S: Setoid<A>) => (as: A[]): Set<A> => {
...
+  // tslint:disable:no-conditional-assignment
+  while (!(e = values.next()).done) {
+    const optionA = e.value
+    if (optionA.isSome()) {
+      result.add(optionA.value)
+    }
+  }
+  return result
+}
+const separate = <RL, RR>(fa: Set<Either<RL, RR>>): Separated<Set<RL>, Set<RR>> => {
```
gcanti: Should be

```ts
const separate = <A, B>(SA: Setoid<A>, SB: Setoid<B>) => (fa: Set<Either<A, B>>): Separated<Set<A>, Set<B>> =>
```

src/Set.ts
```diff
@@ -292,3 +317,46 @@ export const fromArray = <A>(S: Setoid<A>) => (as: A[]): Set<A> => {
...
+    left,
+    right
+  }
+}
+
+/**
+ * @instance
+ * @since 1.7.0
+ */
+export const set: Compactable1<URI> = {
```
gcanti: this is not possible due to previous observations

src/Set.ts
raveclassic: But why? I was also wondering why Set needs a Setoid to check the existence in current implementation. JavaScript Set implementation already filters out duplicates so you can't create `new Set([1, 1])`, result will always be `new Set([1])`. The same applies for `Set.prototype.add`

src/Option.ts
```diff
@@ -555,6 +556,26 @@ export const fromRefinement = <A, B extends A>(refinement: Refinement<A, B>) =>
   return refinement(a) ? some(a) : none
 }
 
+const compact = <A>(fa: Option<Option<A>>): Option<A> => fa.chain(identity)
+const separate = <RL, RR, A>(fa: Option<Either<RL, RR>>): Separated<Option<RL>, Option<RR>> =>
```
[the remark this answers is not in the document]
raveclassic: fixed

src/Set.ts
raveclassic: Furthermore `Filterable` instance requires `Set` to be a valid `Functor` without additional `Setoid`'s. And all tests pass for `map` without `Setoid`

gcanti: Because equality is a serious matter, actually as a mathematician I think that equality is the only thing that really matters :)

`Set` uses implicitly the following setoid

```ts
const setoidEvil: Setoid<any> = {
  equals: (x, y) => x === y
}
```

which is useless, try this `new Set([{ a: 1 }, { a: 1 }])`. 

Actually is worse than useless, is evil, because if you use that setoid, you are pretty much out of funcitonal programming (unless you add some machinery everytime you define a struct/function).

Let's define a `Person` struct

```ts
interface Person {
  name: string
}

const p1: Person = { name: 'foo' }
const p2: Person = { name: 'foo' }
```

"obviously" `p1` and `p2` are equal, however

```ts
console.log(setoidEvil.equals(p1, p2)) // => false
```

Even worse, consider the following function `person`

```ts
// constructor for Person
const person = (name: string): Person => ({ name })

const x1 = 'foo'
const x2 = 'foo'
const y1 = person(x1)
const y2 = person(x2)

console.log(setoidEvil.equals(x1, x2)) // => true
// but...
console.log(setoidEvil.equals(y1, y2)) // => false
```

as you can see `person` is not a pure function, which is (in practice) absurd.

So basically everytime you define an interface or a function or whatever, you are also defining an implicit setoid (or more setoids) instance.
Normally keeping those setoids implicit isn't much of a problem, but in the case of `Set` they are important and we should make them explicit, otherwise you can get duplicates (or even worse results).

raveclassic: Totally makes sense!
What about implementing it similarly to `Either`/`Validation` and accept a setoid in `getCompactable`?

raveclassic: @gcanti I've updated `Set`, take a look

raveclassic: @gcanti What do you think about such approach in general? When you can define all type arguments for a typeclass? I mean that `CompactableA`/`Compactable1A` which is similar to `Compactable2C`. What do you think about it? Is it ok? Is naming appropriate?

gcanti: I'm not sure, it doesn't feel right (and there's no equivalent in purescript-filterable). 
Likely there's no `Compactable` instance for `Set` just as there's no `Functor` instance.
For now I propose to either 

- remove `Set` from this PR and rediscuss later
- or remove `getCompactable` and define `compact` and `separate` as

```ts
export const compact = <A>(S: Setoid<A>) => (fa: Set<Option<A>>): Set<A> => ...
export const separate = <A, B>(SA: Setoid<A>, SB: Setoid<B>) => (fa: Set<Either<A, B>>): Separated<Set<A>, Set<B>> => ...
```

raveclassic: Ok, let's revert Set now and come back to it later.

raveclassic: @gcanti Ready

gcanti:
@raveclassic I missed `CompactableA` and `Compactable1A` in my review (they are still there, I think we should remove them)

raveclassic:
@gcanti Oops, are you ok to see them removed in the next PR with `Filterable`?

gcanti:
np

PR #492, opened 2018-06-22
Feature/filterable

raveclassic:

Adds:
  -  `Filterable` typeclass
  - default implementations for `Filterable` methods
  - default implementations for `Compactable` methods via `Filterable`
  - `eitherBool`, `optionBool` helpers
  - tests
  - `.prettierrc` and `.prettierignore` to support WebStorm watchers

Changes:
  - `Compactable` helper names
  - removes `CompactableA` and `Compactable1A` which were accidentaly added in https://github.com/gcanti/fp-ts/pull/488


src/Compactable.ts
```diff
@@ -119,51 +98,105 @@ export interface Compactable3C<F extends URIS3, U, L> {
  * @function
  * @since 1.7.0
  */
-export function compactDefault<F extends URIS3, U, L>(
```
gcanti: Why the renaming? Also I don't understand `compactDefaultFilterMap` and `separateDefaultPartitionMap`, they seem trivial. Maybe we can just add some documentation explaining how to derive compact/separate from filterMap/partitionMap and avoid some bloat

src/Filterable.ts
```diff
@@ -0,0 +1,357 @@
...
+  readonly filterMap: <A, B>(fa: Type3<F, U, L, A>, f: (a: A) => Option<B>) => Type3<F, U, L, B>
+  readonly filter: <A>(fa: Type3<F, U, L, A>, p: Predicate<A>) => Type3<F, U, L, A>
+}
+
+/**
+ * Upgrade a boolean-style predicate to an either-style predicate mapping.
+ * @function
+ * @since 1.7.0
+ */
+export const eitherBool = <A>(p: Predicate<A>) => (a: A): Either<A, A> => (p(a) ? right(a) : left(a))
```
gcanti: we can remove `eitherBool` since `eitherBool(p) = fromPredicate(p, identity)`

src/Filterable.ts
```diff
@@ -0,0 +1,357 @@
...
+
+/**
+ * Upgrade a boolean-style predicate to a maybe-style predicate mapping.
+ * @function
+ * @since 1.7.0
+ */
+export const optionBool = <A>(p: Predicate<A>) => (a: A): Option<A> => (p(a) ? some(a) : none)
```
gcanti: we can remove `optionBool` since is `fromPredicate` (<= `Option` module)

src/Filterable.ts
```diff
@@ -0,0 +1,357 @@
...
+  return (fa, f) => F.separate(F.map(fa, f))
+}
+
+/**
+ * Gets default implementation of {@link Filterable.partition} using {@link Filterable.partitionMap}
+ * @function
+ * @since 1.7.0
+ */
+export function partitionDefaultPartitionMap<F extends URIS3, U, L>(
+  F: Pick<Filterable3C<F, U, L>, 'URI' | 'partitionMap'>
```
gcanti: why an object here (as argument) instead of

```ts
partitionMap: Pick<Filterable3C<F, U, L>, 'partitionMap'>
```

(idem for the other default-functions)

src/Compactable.ts
raveclassic: Renaming is to be explicit in what is needed for default implementation since there's more than one way to get it. The name reflects purescript implementation, firstly I picked this convention: `get<Name>From<OtherName>` which is more readabe.

Although they are trivial you still need to write them each time, I think it's just handy to have them as a part of the library as well as they are a good way of reflecting relation between these typeclasses, don't you think?

src/Compactable.ts
```diff
@@ -36,19 +38,7 @@ export interface Compactable<F> {
...
-  readonly URI: F
-  readonly _A: A
-  readonly _RL: RL
-  readonly _RR: RR
-  readonly compact: (fa: HKT<F, Option<A>>) => HKT<F, A>
-  readonly separate: (fa: HKT<F, Either<RL, RR>>) => Separated<HKT<F, RL>, HKT<F, RR>>
-}
-
-/**
+ * @typeclass
```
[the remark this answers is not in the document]
raveclassic: ok

src/Filterable.ts
raveclassic: ok

src/Filterable.ts
raveclassic: ok

src/Filterable.ts
raveclassic: Object is more handy when default implementation requires more than one function (like `wither`). Also I have some problems with inference without passing `URI` in `getFilterable` of `Either` (next PR).

gcanti: > I picked this convention: get<Name>From<OtherName> which is more readabe

ok

> as well as they are a good way of reflecting relation between these typeclasses

ok

> Object is more handy when default implementation requires more than one function (like wither)

ok

> Also I have some problems with inference without passing URI in getFilterable of Either (next PR)

Ah I see, ok

PR #493, opened 2018-06-25
Feature/filterable instances

raveclassic:

Adds:
  - `Filterable` instance for `Array`
  - `Filterable` instance for `Either` (`getFilterable` using `Monoid`)
  - `Filterable` instance for `Option`
  - `Filterable` instance for `Validation` (`getFilterable` using `Monoid`)
  - `Filterable` instance for `StrMap`
  -  tests

Changes:
  - ~`Compactable`/`Filterable` default methods signature for `-2C` and `-3C` to require `_L` and `_U`~

Removes:
  - default implementations for `Compactable` methods
  - default implementations for `Filterable` methods


src/Array.ts
```diff
@@ -14,8 +14,9 @@ import { Plus1 } from './Plus'
 import { Setoid, getArraySetoid } from './Setoid'
 import { Traversable1 } from './Traversable'
 import { Unfoldable1 } from './Unfoldable'
-import { Endomorphism, Predicate, Refinement, concat, identity, tuple } from './function'
-import { Compactable1, Separated } from './Compactable'
+import { Endomorphism, Predicate, Refinement, concat, tuple } from './function'
+import { Compactable1, getCompactFromFilterMap, getSeparateFromPartitionMap } from './Compactable'
```
gcanti: I wouldn't use default implementations for the core data structures

src/Filterable.ts
```diff
@@ -123,25 +123,25 @@ export interface Filterable3C<F extends URIS3, U, L> extends Functor3C<F, U, L>,
  * @since 1.7.0
  */
 export function getPartitionMapFromSeparate<F extends URIS3, U, L>(
-  F: Functor3C<F, U, L> & Pick<Compactable3C<F, U, L>, 'separate'>
+  F: Pick<Functor3C<F, U, L> & Compactable3C<F, U, L>, 'URI' | 'map' | '_U' | '_L' | 'separate'>
```
gcanti: mmh this is becoming a bit too messy don't you think?

src/Array.ts
raveclassic: Ok, I'll update all ADTs with efficient versions

src/Filterable.ts
raveclassic: Well, default implementations are likely to be used only by developers building custom ADTs who are already confident with a tricky HKTs in fp-ts. So I don't think there will be any problems to understand this signature because they should already know 'URI', '_U', '_L' and others :) In ther worst case we can add `@example` tag to each default function.

src/Array.ts
gcanti: > Ok, I'll update all ADTs with efficient versions

May I help you? I'm playing with some implementations. Looks like `compact` and `separate` can be derived efficiently from (efficient) `filterMap` and `partitionMap`, no matter what data structure

```
compact = fa => filterMap(fa, identity)
separate = fa => partitionMap(fa, identity)
```

src/StrMap.ts
```diff
@@ -295,32 +288,94 @@ export const pop = <A>(k: string, d: StrMap<A>): Option<[A, StrMap<A>]> => {
   return a.isNone() ? none : some(tuple(a.value, remove(k, d)))
 }
 
-const compact = <A>(fa: StrMap<Option<A>>): StrMap<A> => {
+const filterMap = <A, B>(fa: StrMap<A>, f: (a: A) => Option<B>): StrMap<B> => {
+  const value = fa.value
+  const result: { [key: string]: B } = {}
+  const keys = Object.keys(value)
+  for (const key of keys) {
```
gcanti: Isn't scanning twice here?

```ts
const keys = Object.keys(value)
for (const key of keys) {
}
```

what about?

```ts
for (const key in values) {
}
```

raveclassic: The problem with `for ... in` is that it iterates over all keys even from prototype so either `Object.hasOwnProperty` check should be included or `Object.keys` should be used instead.

raveclassic: @gcanti Search in fp-ts code shows 3 incorrect usages of `for ... in` in `Monoid`, `Semigroup` and `Setoid`. Should we fix them too?

gcanti: I always assumed `value` as a "simple" literal object (as it should be). In fact `for ... in` is used in many other functions in this module (`StrMap`)

raveclassic: But typesystem can't restrict such value to always be a "simple" literal object so you can't be sure that you'll never get some additional keys from the prototype. 
```typescript
const numbers: {[key: string]: number} = {}
Object.setPrototypeOf(numbers, { evil: 'evil' })
for (const key in numbers) {
  const n = numbers[key]
  n.toFixed(2) // OOUCH!
}
```

I think the most safe and efficient way is to add `hasOwnProperty` everywhere, what do you think?

raveclassic: There's even a tslint rule for such cases: https://palantir.github.io/tslint/rules/forin/

raveclassic: On the other hand if you have an object with 10k keys set as prototype you end up with iterating over all that keys even if you object contains only one key. I think we should check what is faster - iterating over all keys with `hasOwnProperty` check or calling `Object.keys` and iterating over collection of keys. I think `V8` should perform some tricks to optimize calls to `Object.keys`, `Object.values` etc.

raveclassic: Aaaaand the winner is: https://jsperf.com/object-keys-vs-hasownproperty/45

gcanti: LOL, ok let's add `"forin": true` to tslint.json

PR #508, opened 2018-06-28
Add Group Function for Array

MaximeRDY:
Hi,

I've just added a `group` function for Array.

### DIFF
--- a/src/NonEmptyArray.ts
+++ b/src/NonEmptyArray.ts
@@ -1,5 +1,5 @@
 import { Applicative } from './Applicative'
-import { array, last, sort } from './Array'
+import { array, last, sort, span, head, snoc } from './Array'
 import { Comonad1 } from './Comonad'
 import { Foldable1 } from './Foldable'
 import { HKT } from './HKT'
@@ -8,7 +8,8 @@ import { Option, none, some } from './Option'
 import { Ord } from './Ord'
 import { Semigroup, fold, getJoinSemigroup, getMeetSemigroup } from './Semigroup'
 import { Traversable1 } from './Traversable'
-import { concat as uncurriedConcat, toString } from './function'
+import { concat as uncurriedConcat, toString, curry, compose } from './function'
+import { Setoid } from './Setoid'
 
 declare module './HKT' {
   interface URI2HKT<A> {
@@ -288,6 +289,35 @@ const concat = <A>(fx: NonEmptyArray<A>, fy: NonEmptyArray<A>): NonEmptyArray<A>
 export const getSemigroup = <A = never>(): Semigroup<NonEmptyArray<A>> => {
   return { concat }
 }
+
+/**
+ *  Group equal, consecutive elements of an array into arrays.
+ */
+export const group = <A>(S: Setoid<A>) => (as: Array<A>): Array<NonEmptyArray<A>> =>
+  head(as).fold([], a => recursiveSpan<A>([])(S)(a)(as))
+
+const recursiveSpan = <A>(mem: Array<NonEmptyArray<A>>) => (S: Setoid<A>) => (a: A) => (
+  as: Array<A>
+): Array<NonEmptyArray<A>> => {
+  const spanned = span(as, curry(S.equals)(a))
+  const first = head(spanned.rest)
+  const newMem = snoc(mem, unsafeFromArray(spanned.init))
+
+  if (first.isSome()) {
+    return recursiveSpan(newMem)(S)(first.value)(spanned.rest)
+  } else {
+    return newMem
+  }
+}
+
+/**
+ *  Sort and then group the elements of an array into arrays.
+ */
+export const groupSort = <A>(O: Ord<A>) => (as: Array<A>): Array<NonEmptyArray<A>> =>
+  compose(
+    group(O),
+    sort(O)
+  )(as)
 
 const reduce = <A, B>(fa: NonEmptyArray<A>, b: B, f: (b: B, a: A) => B): B => {
   return fa.reduce(b, f)
--- a/src/NonEmptyArray.ts
+++ b/src/NonEmptyArray.ts
@@ -299,19 +299,19 @@ export const group = <A>(S: Setoid<A>) => (as: Array<A>): Array<NonEmptyArray<A>
   if (len === 0) {
     return r
   }
-  let a: A = as[0]
-  let nonEmpty: Array<A> = [a]
+  let head: A = as[0]
+  let tail: Array<A> = []
   for (let i = 1; i < len; i++) {
     const x = as[i]
-    if (S.equals(x, a)) {
-      nonEmpty.push(x)
+    if (S.equals(x, head)) {
+      tail.push(x)
     } else {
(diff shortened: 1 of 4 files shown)


gcanti:
@MaximeRDY is this a porting of https://github.com/purescript/purescript-arrays/blob/7bdd04cef18f63afe8fe0bb59b4180e030e1e0f5/src/Data/Array.purs#L843 ?

MaximeRDY:
@gcanti Yes a kind of

gcanti:
Ok, then

1) should return an `Array<NonEmptyArray<A>>` so I'd move this function to the `NonEmptyArray` module

2) looks like there's something wrong in the implementation: the following tests (from `purescript-arrays`) fail

```ts
const isOdd = (n: number): boolean => n % 2 === 1

const setoidOdd: Setoid<number> = {
  equals: (x, y) => isOdd(x) && isOdd(y)
}

assert.deepEqual(group(setoidNumber)([1, 1, 2, 2, 1]), [[1, 1], [2, 2], [1]])
assert.deepEqual(group(setoidOdd)([1, 3, 2, 4, 3, 3]), [[1, 3], [2], [4], [3, 3]])
```

MaximeRDY:
Ok maybe I've figured why the tests fail. 

The group method work like this:
```
assert.deepEqual(group(setoidNumber)([1, 2, 1, 1, 2]), [[1, 1, 1], [2, 2]])
```

I'm wondering why it should respond like this: 
```
assert.deepEqual(group(setoidNumber)([1, 1, 2, 2, 1]), [[1, 1], [2, 2], [1]])
```

sledorze:
Maybe because it groups successive values? (You may look at purescript test/impl)

gcanti:
I think that you implemented something more like `group'`

https://github.com/purescript/purescript-arrays/blob/7bdd04cef18f63afe8fe0bb59b4180e030e1e0f5/src/Data/Array.purs#L851

however `group` is more general (`group'` can be derived)

MaximeRDY:
@gcanti That make sense :-) 
I'll try to do it today or this week-end. 
Do you have any advice about the naming for `group'` ?

raveclassic:
Such `group` can generalized for any `Filterable`. What about moving to `Filterable.ts`? (similar to `unique` in https://github.com/gcanti/fp-ts/pull/510)

gcanti:
> Do you have any advice about the naming for group' ?

mhh actually no.. @sledorze any suggestion?

gcanti:
> Such `group` can be generalized for any `Filterable`

Ah, I missed this. @raveclassic how? Could you please elaborate?

raveclassic:
Well I missed that there should be a notion of order of Filterable structures for such functions to work generally

sledorze:
@gcanti I think the difference between group and group' in purescript is not very explicit.
May we be more explicit with a two word name like: `globalGroup` or `groupSorted` or `groupDistinct`?

(I'm bad at naming)

gcanti:
@sledorze maybe `groupSort`, following this standard pattern:

- `flatMap` = `flatten` after `map`
- `filterMap` = `filter` after `map`
- `groupSort` = `group` after `sort`

sledorze:
@gcanti like it, coherent and in line with point free composition order.

MaximeRDY:
@gcanti Hi, 

I've moved the function in `NonEmptyArray` and I've added the `groupSort` function. 

Tell me if it's ok  :-)

src/NonEmptyArray.ts
```diff
@@ -289,6 +290,35 @@ export const getSemigroup = <A = never>(): Semigroup<NonEmptyArray<A>> => {
   return { concat }
 }
 
+/**
+ *  Group equal, consecutive elements of an array into arrays.
+ */
+export const group = <A>(S: Setoid<A>) => (as: Array<A>): Array<NonEmptyArray<A>> =>
```
gcanti: Nice. However for a core API like this I'd go for a more efficient mutable solution, something along the lines of

```ts
export const group2 = <A>(S: Setoid<A>) => (as: Array<A>): Array<NonEmptyArray<A>> => {
  const r: Array<NonEmptyArray<A>> = []
  const len = as.length
  if (len === 0) {
    return r
  }
  let a: A = as[0]
  let nonEmpty: Array<A> = [a]
  for (let i = 1; i < as.length; i++) {
    const x = as[i]
    if (S.equals(x, a)) {
      nonEmpty.push(x)
    } else {
      r.push(unsafeFromArray(nonEmpty))
      nonEmpty = [x]
    }
    a = x
  }
  r.push(unsafeFromArray(nonEmpty))
  return r
}
```

Quick benchmark (node v10.1.0)

```
group x 629,882 ops/sec ±0.82% (73 runs sampled)
group2 x 4,459,547 ops/sec ±0.77% (87 runs sampled)
```

src/NonEmptyArray.ts
```diff
@@ -289,6 +290,35 @@ export const getSemigroup = <A = never>(): Semigroup<NonEmptyArray<A>> => {
...
+    return recursiveSpan(newMem)(S)(first.value)(spanned.rest)
+  } else {
+    return newMem
+  }
+}
+
+/**
+ *  Sort and then group the elements of an array into arrays.
+ */
+export const groupSort = <A>(O: Ord<A>) => (as: Array<A>): Array<NonEmptyArray<A>> =>
```
gcanti: You can apply an eta-reduction here

```ts
export const groupSort = <A>(O: Ord<A>): ((as: Array<A>) => Array<NonEmptyArray<A>>) => {
  return compose(
    group(O),
    sort(O)
  )
}
```

MaximeRDY: Oh thanks. I didn't know this trick :-)

MaximeRDY:
@gcanti 
Ok I've pushed what you've suggested. 

The only change I've make it's that: 
`for (let i = 1; i < len; i++) {`

gcanti:
@MaximeRDY this is even better

```ts
export const group3 = <A>(S: Setoid<A>) => (as: Array<A>): Array<NonEmptyArray<A>> => {
  const r: Array<NonEmptyArray<A>> = []
  const len = as.length
  if (len === 0) {
    return r
  }
  let head: A = as[0]
  let tail: Array<A> = []
    const x = as[i]
    if (S.equals(x, head)) {
      tail.push(x)
    } else {
      r.push(new NonEmptyArray(head, tail))
      head = x
      tail = []
    }
  }
  r.push(new NonEmptyArray(head, tail))
  return r
}
```

```
group x 627,242 ops/sec ±0.56% (78 runs sampled)
group2 x 4,381,062 ops/sec ±0.79% (86 runs sampled)
group3 x 10,139,893 ops/sec ±0.43% (89 runs sampled)
```

gcanti:
@raveclassic your version gives

```
group x 429,640 ops/sec ±0.84% (88 runs sampled)
```

raveclassic:
The point is in generalization, not in perf :)

gcanti:
I see but that means we can't use your generic function to implement `group` for arrays, which this PR is all about

gcanti:
Thanks @MaximeRDY

PR #529, opened 2018-07-17
Internal: fix upstream weird behaviour when type checking Type with w…

gcanti:
…rong URI

Without this fix the following type-checks even if the provided URI is invalid

```ts
import { Type } from 'fp-ts/lib/HKT'

type T1 = Type<'a', string>
```


sledorze:
Typescript typing is sometimes so weird... something not easy to reason about.

gcanti:
@sledorze that's a really polite way to say WTF :)

p.s.
mes félicitations pour l'équipe

sledorze:
@gcanti Yes we really had a great team play ! :)

PR #530, opened 2018-07-19
Add forEach to Option

kylecorbelli:
Scala's `Option` defines a `foreach` method:
https://www.scala-lang.org/api/current/scala/Option.html#foreach[U](f:A=%3EU):Unit

This PR adds and tests an analogous `forEach` method to both `Some` and `None` as well as provides a corresponding standalone `forEach` utility function.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "fp-ts",
-  "version": "1.7.1",
+  "version": "1.8.0",
   "description": "Functional programming in TypeScript",
   "files": [
     "lib"
--- a/src/Option.ts
+++ b/src/Option.ts
@@ -230,6 +230,17 @@ export class None<A> {
   refine<B extends A>(refinement: Refinement<A, B>): Option<B> {
     return none
   }
+
+  /**
+   * Apply the given procedure f to the option's value, if it is nonempty.
+   * @example
+   * some('foo').forEach(console.log) // logs 'foo'
+   * none.forEach(console.log) // does nothing
+   * @since 1.8.0
+   */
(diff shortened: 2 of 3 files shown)


gcanti:
@kylecorbelli Thanks for this PR, I'm afraid I can't accept it though, this signature is inherently impure (or does nothing)

```ts
forEach(f: (a: A) => void): void
```

Related https://github.com/gcanti/fp-ts/pull/184#issuecomment-319211460

kylecorbelli:
@gcanti I totally understand. The Haskeller in me cringed a little when writing it but this repo in general seemed more inspired by Scala than Haskell so I figured it might be worth a shot.

Now, assuming this repo _is_ inspired by Scala but aims to improve Scala's shortcomings (i.e. an untyped and unaccounted for side-effect in `foreach`), do you think there's room for this method if we can improve the type? Perhaps some riff on `IO`?

dashuser33:
Since this was not merged, what is the recommended way to run a side effect only if it is Some? Do a map and ignore the result?

mlegenhausen:
To keep you code functional you need to go from `Option<A> -> Task<void>` this can be done with folding

```ts
const task = some(...).fold(
  () => task.of(undefined),
  (v) => new Task(() => /* do you side effects here */)
);

// Your side effects gets executed here
task.run();
```

or cleaner with `foldMap`

```ts
const foldOption = foldMap(option, getMonoid(monoidVoid))

const task = foldOption(some(...), (v) => new Task(() => /* do you side effects here */));

task.run();
```

PR #533, opened 2018-07-21
New Feature: add LocalStorage module

gcanti:


mlegenhausen:
Does this not break with the current purity of `fp-ts` to be platform independent? Currently all features can be used either in the browser or node. IMHO this seems to be an extension like `browser-ts` for me?

gcanti:
Yes, I wasn't sure about this. All modules are optional though, you can import just what you need.

test/index.ts
```diff
@@ -17,12 +17,16 @@ const getExportName = (name: string): string => {
   return name.substring(0, 1).toLowerCase() + name.substring(1)
 }
 
+const shouldExport = (name: string): boolean => {
+  return !(name in { index: null, LocalStorage: null })
```
raveclassic: isn't it too specific?

gcanti: This test is just a dirty way to remember the export when adding a new module

mlegenhausen:
> All modules are optional though, you can import just what you need.

Where would you draw the line? This can be interpreted as to collapse the whole `fp-ts` ecosystem into one package? I am just curious as a (possible) contributor where to contribute which feature.

gcanti:
> Where would you draw the line?

Not sure, let's decide together. I'd say

Criteria

- no dependencies (for example fp-ts-rxjs is an external package because of the rxjs dependency)
- non opinionated (for example monocle-ts is an external package because is somehow opinionated)
- the new module is just a (type-safe) wrapper of native APIs (for example LocalStorage)
- there's prior art (a Haskell / PureScript / Scala established library)
- other?

mlegenhausen:
It is really hard to argue. It is just what I see how most of the libraries are organized. You have some core (what I would see is `fp-ts`) and an ecosystem around it. Some examples would be `react` and `react-dom`, `reflex` and `reflex-dom` and so on. It feels like `lodash` or `rambda` would add `localStorage` support. Which seems to make the library as I said "impure" and destroys this "core" aspect.

So an argument I would like to add is "runs in every actual js engine (v8)".

raveclassic:
@mlegenhausen 

> "runs in every actual js engine (v8)"

I'd say  "every part of the library runs in every js engine"

gcanti:
Ok, I'll put up a `fp-ts-local-storage` package, thanks for the discussion.

mlegenhausen:
@gcanti I like your critera list should we add a `CONTRIBUTOR.md` file?

gcanti:
@mlegenhausen sure, would you like to send a PR?

mlegenhausen:
@gcanti what would be the prefix/suffix we should use for new packages? Some developers already used the `-ts` suffix for custom packages like you do for `io-ts`, `monocle-ts` and `elm-ts`. Which seems sometimes misleading when we think about `node-ts` which has nothing todo with the project.

It would be nice if could define one, so packages can be found easier even when they are not referenced in the readme.

gcanti:
@mlegenhausen Personally I use the suffix `-ts` to mean an existing library ported to TypeScript and the prefix `fp-ts-` if is a kind of add-on of `fp-ts` (for example some bindings)

PR #555, opened 2018-09-05
Update HKT.ts

babakness:
Hello, I'm using HKTs with my library and have a need for four parameters. In my case I need to propagate a before and after type transformation for two related objects. This is a pull request to create URI2HKT4. I chose the variable `V` simply because it is the successor to `U`.

### DIFF
--- a/src/HKT.ts
+++ b/src/HKT.ts
@@ -44,5 +44,5 @@ export type Type4<URI extends URIS4, V, U, L, A> = {} & URI2HKT4<V, U, L, A>[URI
 (null! as URI2HKT<any>) as { [k in keyof URI2HKT<any>]: HKT<k, any> }
 (null! as URI2HKT2<any, any>) as { [k in keyof URI2HKT2<any, any>]: HKT2<k, any, any> }
 (null! as URI2HKT3<any, any, any>) as { [k in keyof URI2HKT3<any, any, any>]: HKT3<k, any, any, any> }
-(null! as URI2HKT4<any, any, any>) as { [k in keyof URI2HKT4<any, any, any, any>]: HKT4<k, any, any, any, any> }
+(null! as URI2HKT4<any, any, any, any>) as { [k in keyof URI2HKT4<any, any, any, any>]: HKT4<k, any, any, any, any> }
 /* tslint:enable */
--- a/src/HKT.ts
+++ b/src/HKT.ts
@@ -11,6 +11,10 @@ export interface HKT3<URI, U, L, A> extends HKT2<URI, L, A> {
   readonly _U: U
 }
 
+export interface HKT3<URI, V, U, L, A> extends HKT2<URI, L, A> {
+  readonly _U: U
+}
+
 // type-level dictionaries for HKTs
 
 export interface URI2HKT<A> {}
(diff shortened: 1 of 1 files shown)


src/HKT.ts
```diff
@@ -24,15 +25,21 @@ export type URIS2 = (URI2HKT2<any, any> & { never: HKT<never, never> })[keyof UR
...
   | 'never']['_URI']
+export type URIS4 = (URI2HKT3<any, any, any, any> & { never: HKT<never, never> })[
+  | keyof URI2HKT3<any, any, any, any>
+  | 'never']['_URI']
+
 
 export type Type<URI extends URIS, A> = {} & URI2HKT<A>[URI]
 export type Type2<URI extends URIS2, L, A> = {} & URI2HKT2<L, A>[URI]
 export type Type3<URI extends URIS3, U, L, A> = {} & URI2HKT3<U, L, A>[URI]
+export type Type4<URI extends URIS4, V, U, L, A> = {} & URI2HKT3<V, U, L, A>[URI]
```
sledorze: URI2HKT3 => URI2HKT4

src/HKT.ts
```diff
@@ -24,15 +25,21 @@ export type URIS2 = (URI2HKT2<any, any> & { never: HKT<never, never> })[keyof UR
 export type URIS3 = (URI2HKT3<any, any, any> & { never: HKT<never, never> })[
   | keyof URI2HKT3<any, any, any>
   | 'never']['_URI']
+export type URIS4 = (URI2HKT3<any, any, any, any> & { never: HKT<never, never> })[
```
sledorze: URI2HKT3 => URI2HKT4

src/HKT.ts
```diff
@@ -24,15 +25,21 @@ export type URIS2 = (URI2HKT2<any, any> & { never: HKT<never, never> })[keyof UR
...
+  | keyof URI2HKT3<any, any, any, any>
```
sledorze: URI2HKT3 => URI2HKT4

raveclassic:
This is insteresting, could you please describe your usecase for 4 type args? It's quite rare

babakness:
Apologies for all the commits, was working using GitHub's online editor and no color coding makes it harder to read. 

@raveclassic Sure, I working on a typed Lens ADT, it takes an object and a path--the type for the object before/after the transform is needed, that's two. The type at the path to be transform is the input for the value for the next morphism, that's three. The default value should the path be null / undefined, four.

src/HKT.ts
```diff
@@ -11,6 +11,10 @@ export interface HKT3<URI, U, L, A> extends HKT2<URI, L, A> {
   readonly _U: U
 }
 
+export interface HKT3<URI, V, U, L, A> extends HKT2<URI, L, A> {
```
sledorze: HKT3 => HKT4 :)

babakness: I believe in this version, this is correct

https://github.com/gcanti/fp-ts/blob/2fa6998fa2e9b1b2e9bb823afc8888ac5fa5fe39/src/HKT.ts#L14

gcanti:
Thanks @babakness

PR #558, opened 2018-09-06
New Feature: add Zipper module

gcanti:


src/Zipper.ts
```diff
@@ -0,0 +1,279 @@
...
+   * @since 1.9.0
+   */
+  modify(f: (a: A) => A): Zipper<A> {
+    return this.update(f(this.focus))
+  }
+  /**
+   * @since 1.9.0
+   */
+  toArray(): Array<A> {
+    return snoc(this.lefts, this.focus).concat(this.rights)
```
gabro: Just wondering if this.lefts.concat(cons(this.focus, this.rights)) would be more efficient.

This assumes cons to be constant time, and snoc being linear.

gabro: ah no, I see cons is linear too, so no difference there

PR #562, opened 2018-09-09
[RFC] New Feature: add Foldable2v module

gcanti:
This is a backward compatible "replacement" of the `Foldable` module that adds the following operations to the existing type class

- `foldMap`
- `foldr`

This addition makes possible to implement more efficient versions of these functions based on the specific data types.

This PR also implements instances of `Foldable2v` for the following data types (more to come)

- `Array`
- `Option`

The `Foldable` module should be deprecated, and replaced with the `Foldable2v`'s implementation in fp-ts@2.0 .

The goal is to move towards better type class definitions in a soft way.

/cc @raveclassic @sledorze

### DIFF
--- a/src/Array.ts
+++ b/src/Array.ts
@@ -3,7 +3,7 @@ import { Applicative, Applicative1, Applicative2, Applicative2C, Applicative3, A
 import { liftA2 } from './Apply'
 import { Either } from './Either'
 import { Extend1 } from './Extend'
-import { Foldable1 } from './Foldable'
+import { Foldable2v1 } from './Foldable2v'
 import { HKT, Type, Type2, Type3, URIS, URIS2, URIS3 } from './HKT'
 import { Monad1 } from './Monad'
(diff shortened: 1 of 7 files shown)


sledorze:
@gcanti what's to discuss here specifically? (sorry to ask without diving in it - very, very few extra time those days)

gcanti:
> what's to discuss here specifically?

@sledorze I'd love to hear your opinion on this course of action

- add `Foldable2v` instead of a breaking change
- deprecate `Foldable`
- replace `Foldable` with `Foldable2v` in `fp-ts@2.0.0` (in a possibly distant future...)

Do you think it's worth it, or should we just wait?

sledorze:
@gcanti to be totally honest, I've not checked it (struggling with limited time and focus).
I think that's the right course of action, adding the new implementation in a non breaking way before 2.0.0 is totally fine.

PR #563, opened 2018-09-13
Make TaskEither a Monoid

mlegenhausen:
The added functions are a combination of the Monoid definitions from `Task` and `Either`.

### DIFF
--- a/src/TaskEither.ts
+++ b/src/TaskEither.ts
@@ -1,12 +1,21 @@
 import { Alt2 } from './Alt'
 import { Bifunctor2 } from './Bifunctor'
-import { Either, fromPredicate as eitherFromPredicate, left as eitherLeft, right as eitherRight } from './Either'
+import {
+  Either,
+  fromPredicate as eitherFromPredicate,
+  left as eitherLeft,
+  right as eitherRight,
+  getSemigroup as eitherGetSemigroup,
+  getApplySemigroup as eitherGetApplySemigroup
+} from './Either'
 import * as eitherT from './EitherT'
 import { IO } from './IO'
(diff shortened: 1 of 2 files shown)


src/TaskEither.ts
```diff
@@ -174,6 +183,50 @@ export const fromPredicate = <L, A>(predicate: Predicate<A>, whenFalse: (a: A) =
...
+  return {
+    concat: (x, y) => new TaskEither<L, A>(S2.concat(x.value, y.value))
+  }
+}
+
+/**
+ * @function
+ * @since 1.9.0
+ */
+export const getMonoid = <L, A>(M: Monoid<A>): Monoid<TaskEither<L, A>> => {
```
gcanti: This is unlawful, note that there's no `getMonoid` in the `Either` module.

Counterexample

```ts
it('right unit and a left', () => {
  return M.concat(left(delay(10, 'a')), M.empty)
    .run()
    .then(x => assert.deepEqual(x, eitherLeft('a'))) // fails
})
```

mlegenhausen: Was already wondering. I will remove the `getMonoid` function.

test/TaskEither.ts
```diff
@@ -215,31 +214,6 @@ describe('TaskEither', () => {
     })
   })
 
-  describe('getMonoid', () => {
```
gcanti: we can keep the tests related to Semigroup though

mlegenhausen: I added a separate set of `getSemigroup` tests which are equal to the tests from `Either`.

gcanti:
Thanks @mlegenhausen

PR #578, opened 2018-10-07
New Feature: add Date module

gcanti:


vegansk:
@gcanti , should this really be in the core library? Maybe it's like the `date` in `io-ts`, should be in some specific primitives library?

mlegenhausen:
We already added `ordDate` so adding `Date` to `fp-ts` seems reasonable.

sledorze:
Environement could include running some typescript code in a jvm context (nashorn, etc..) which may not provide the js but java api.
Adding that to the discussion.. :)

mlegenhausen:
Maybe if we think bigger we can move it completely to someting like `fp-ts-date` and add more date manipulation functions to it so it can grow to something like `momentjs` for FP.

@sledorze seems still possible to use `Date` even when combined with nashorn. Of cause if a lot more code but nothing a Java Developer really fears :wink: (https://stackoverflow.com/questions/25385911/nashorn-nativedate-conversion-to-java-util-date)

PR #579, opened 2018-10-07
New Feature: add TaskEither.attempt

gcanti:


gcanti:
> where does it come from?

@giogonzo From `MonadError`

- [purescript-aff](https://github.com/slamdata/purescript-aff/blob/f6d95e934b662f8832cc3d68bbac59526b6ab6c1/src/Effect/Aff.purs#L259)
- which is a specialization of [purescript-transformers](https://github.com/purescript/purescript-transformers/blob/87bd95f8ca68b7710f3e45c16fb9bb3e74c7f723/src/Control/Monad/Error/Class.purs#L58)
- [in Scala's cats](https://github.com/typelevel/cats/blob/9c56c4bac936ed825cf2cf6ee69d5b3fefa411bb/core/src/main/scala/cats/ApplicativeError.scala#L65)

> what's the use case for this?

useful for some combinator like `braket`...

```ts
import { fromLeft, TaskEither, taskEither } from 'fp-ts/lib/TaskEither'

/**
 * Make sure that a resource is cleaned up in the event of an exception. The
 * release action is called regardless of whether the body action throws or
 * returns.
 */
export const bracket = <L, A, B>(
  acquire: TaskEither<L, A>,
  use: (a: A) => TaskEither<L, B>,
  release: (a: A) => TaskEither<L, void>
): TaskEither<L, B> => {
  return acquire.chain(a =>
    attempt(use(a)).chain(e => release(a).chain(() => e.fold<TaskEither<L, B>>(fromLeft, taskEither.of)))
  )
}
```

...or `handle` (a temporary name for a simplified version of https://github.com/gcanti/fp-ts/issues/583)

```ts
const handle = <L, A, B>(
  fa: TaskEither<L, A>,
  left: (l: L) => TaskEither<L, B>,
  right: (a: A) => TaskEither<L, B>
): TaskEither<L, B> => {
  return attempt(fa).chain(e => e.fold(left, right))
}
```

> Also, why any as left type?

Ideally should be `never`, however it would make using `attempt` awkward to use. `any` was my (clumsy) way to make `attempt` more versatile

```ts
//                    v--- additional type parameter
const handle = <L, A, M, B>(
  fa: TaskEither<L, A>,
  left: (l: L) => TaskEither<M, B>,
  right: (a: A) => TaskEither<M, B>
): TaskEither<M, B> => {
  return attempt(fa).chain(e => e.fold(left, right)) // it doesn't compile without any
}
```

however it's not safe so I'm bending over

```ts
export const attempt = <L, A, M = L>(fa: TaskEither<L, A>): TaskEither<M, Either<L, A>> => {
  return new TaskEither(fa.value.map(e => right(e)))
}
```

mlegenhausen:
@gcanti thats great! I need them all! :wink:

The latter defintion of `attempt` looks great so far. In combination with `handle` and `bracket` everything looks good (everything compiles) in my current project.

Would be nice if you add `handle` and `bracket` too.

mlegenhausen:
Just an example for transaction management with `handle`

```ts
handle(
  client.query('begin').chain(() => /* your query here */),
  (l) => client.query('rollback').chain(() => fromLeft(l)),
  (r) => client.query('commit').map(() => r)
);
```

if we would extend the `release` to `release: (a: A, e: Either<L, B>) => TaskEither<L, void>` we could decide what todo depending on the `use` result. My example could be reduced to

```ts
bracket(
  client.query('begin'),
  () => /* your query here */
  (_, e) => client.query(e.fold(() => 'rollback', () => 'commit'))
```

gcanti:
Looks like `attempt` is even more handy if implemented as a method (same for `handle`).

> Would be nice if you add handle and bracket too

@mlegenhausen ok I'll put up specific PRs. Btw is there a better name for `handle`?

> if we would extend the release...

I'm fine with that, in cats is named [bracketCase](https://github.com/typelevel/cats-effect/blob/c0b71639fcfe2daf04989621b3be67f1ad90a4a6/core/shared/src/main/scala/cats/effect/Bracket.scala#L45)

mlegenhausen:
Maybe `attemptFold`?

Issue #265, opened 2017-11-04
Documentation examples


brucou:
Hi there.

First of all, congrats on this extensive library. I followed for some time your development effort and I must say, and saw the adjunction of documentation (mostly based on type signature), and I think that is great. 

However I think that users like me would benefit enormously from a documentation which also portrays examples. A good example in that aspect is Ramda's, which I guess you all know (http://ramdajs.com/docs). Leaving aside the bells and whistles of ramda's doc (playground, etc.), what is great is the examples and the categorization (Function, Map, List, Object). Current documentation for fp-ts consist mainly of the name of the function and the corresponding type, which makes the whole thing not very friendly to absorb for beginners.

A part of the docs already do some of this :  

- https://github.com/gcanti/io-ts
- https://github.com/gcanti/newtype-ts

Something I was looking at recently with a view to using it and I could not figure out exactly how to do so : 

- https://github.com/gcanti/fp-ts/blob/master/docs/api/md/Store.md
- https://github.com/gcanti/fp-ts/blob/master/docs/api/md/Traversable.md
  - That http://ramdajs.com/docs/#traverse reads better than `traverse: <F>(F: Applicative<F>) => <A, B>(f: (a: A) => HKT<F, B>, ta: HKT<T, A>) => HKT<F, HKT<T, B>>`

I can probably figure out all of what I need to figure out but my fear is that I would spend too much time doing so. I can see in my code that I can benefits from functional abstraction (for instance I have a contract library which check assertions and aggregate errors, all of which could benefit from the applicative programming tools you have in fp-ts). But with the perceived syntax, abstraction, and documentation barriers, I feel like it is safer to keep things in plain and simple js/ts. Even supposing I figure everything out in a relatively short time, my colleagues or whoever comes after me, will have to do the exact same job, which compounds the perceived productivity loss.

Now, I am aware that there are a HUGE number of functions, and functionalities addressed in this library, so it looks like a large quantity of work to add examples everywhere. I still wanted however to give you guys my feedback, as I think it is also interesting to hear about the people who are not using the library, and why.

When I am between projects and with more time, I will be happy to properly dedicate resources to see how to integrate this library in my work.

raveclassic:
@brucou 

> Something I was looking at recently with a view to using it and I could not figure out exactly how to do so

The reason is that ramda and other fp libs use Hindley–Milner notation (which is common for Haskell) when fp-ts being designed for TypeScript uses TypeScript signatures to express the type.
[...]

PR #594, opened 2018-10-16
Add example to docs for Array.member function

fozcodes:
Hey there!

My team just started using this lib and I saw in #265 that you're taking PRs to help
improve the docs. I personally would love some examples for functions like `member` so
I thought I'd add one.

I'll be sure to throw up more PRs of examples as I go.

Great library!

### DIFF
--- a/src/Array.ts
+++ b/src/Array.ts
@@ -705,7 +705,13 @@ export const rotate = <A>(n: number, xs: Array<A>): Array<A> => {
 }
 
 /**
- * Test if a value is a member of an array
+ * Test if a value is a member of an array. Takes a `Stetoid<A>` as a single
+ * argument which returns the function to use to search for a value of type `A` in
+ * an array of type `Array<A>`.
+ *
+ * @example
+ * member(stetoid.stetoidString)(['thing one', 'thing two', 'cat in the hat'], 'thing two') // true
+ *
  * @function
  * @since 1.3.0
  */
--- a/src/Array.ts
+++ b/src/Array.ts
@@ -705,12 +705,14 @@ export const rotate = <A>(n: number, xs: Array<A>): Array<A> => {
 }
 
 /**
- * Test if a value is a member of an array. Takes a `Stetoid<A>` as a single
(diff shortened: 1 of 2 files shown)


gcanti:
Thank you @fozcodes. The md files are generated though, so you should modify the [source file](https://github.com/gcanti/fp-ts/blob/f6580b7975fd7ab2822e6f148173b3c4958a792d/src/Array.ts#L708).

Something like

```ts
/**
 * Test if a value is a member of an array. Takes a `Stetoid<A>` as a single
 * argument which returns the function to use to search for a value of type `A` in
 * an array of type `Array<A>`.
 *
 * @example
 * member(stetoid.stetoidString)(['thing one', 'thing two', 'cat in the hat'], 'thing two') // true
 *
 * @function
 * @since 1.3.0
 */
```

fozcodes:
Ah! Of course. I knew there was some generation happening somewhere. Just picked the wrong place 🤦‍♂️ .

Updated.

src/Array.ts
```diff
@@ -705,7 +705,13 @@ export const rotate = <A>(n: number, xs: Array<A>): Array<A> => {
 }
 
 /**
- * Test if a value is a member of an array
+ * Test if a value is a member of an array. Takes a `Stetoid<A>` as a single
```
raveclassic: there's a small type here: `Stetoid<A>`

gcanti: Thanks @raveclassic. Looks like there's another typo in the example. I think we can drop the prefix `setoid.` and go for

```
@example
import { setoidString } from 'fp-ts/lib/Setoid'

member(setoidString)(['thing one', 'thing two', 'cat in the hat'], 'thing two') // true
```

which also shows the required import

fozcodes: Great catch @raveclassic and good call @gcanti - that's much clearer to show the dependency.

Updated. 👍

gcanti:
@raveclassic if it's ok for you I'm gonna merge this

raveclassic:
@gcanti Yeah, ok

PR #595, opened 2018-10-17
Docs: comparison with ramda

gcanti:


raveclassic:
Wow, I bet this immediately ends in filling the gaps in the list and replacing `ramda` with `fp-ts`! 🤣

PR #597, opened 2018-10-19
Docs: add examples to Array module

gcanti:


src/Array.ts
```diff
@@ -832,6 +1017,29 @@ export const uniq = <A>(S: Setoid<A>): ((as: Array<A>) => Array<A>) => {
...
+ *
+ * @example
+ * import { contramap, ordString, ordNumber } from 'fp-ts/lib/Ord'
+ *
+ * interface Person {
+ *   name: string
+ *   age: number
+ * }
+ * const byName = contramap((p: Person) => p.name, ordString)
+ * const byAge = contramapOrd((p: Person) => p.age, ordNumber)
```
giogonzo: should be just `contramap`
```suggestion
 * const byAge = contramap((p: Person) => p.age, ordNumber)
```

PR #610, opened 2018-10-28
getObjectSemigroup signature fix in docs

andys8:
I assume the trailing semicolon is a typo. This will remove it from the signature of `getObjectSemigroup`.


gcanti:
@andys8 thanks for reporting :+1:, docs are generated so this signature must be fixed [in the source](https://github.com/gcanti/fp-ts/pull/611)

PR #611, opened 2018-10-29
Docs: fix getObjectSemigroup signature

gcanti:


andys8:
@gcanti So you had to fix it by adding a (unnecessary) `return` statement to the implementation? Why that? 😁

By the way, here is another semicolon: https://github.com/gcanti/fp-ts/blob/master/docs/function.md#unsafecoerce

gcanti:
because the (custom) doc generator is terrible

PR #613, opened 2018-10-29
TS node dependency

andys8:
Adding `ts-node` as dependency. It's missing, but used here: https://github.com/gcanti/fp-ts/blob/6d48327c5e9e12d1e68d9856a9a7fb72a3f3aee2/package.json#L21


gcanti:
Thanks @andys8

PR #614, opened 2018-10-29
Update generated docs, diff in update

andys8:
Checking out `master` and executing the `docs` script, I get a diff in the output. `string` is used over `String`, which looks like it would make sense.

Creating this commit mostly to raise awareness, as the diff shouldn't be the case in an ideal world ;)


andys8:
Okay, it was this recent commit, that didn't update the docs. https://github.com/gcanti/fp-ts/commit/6d48327c5e9e12d1e68d9856a9a7fb72a3f3aee2

No big deal, but perhaps travis could generate the docs or that there is no diff (enforce them to be updated manually).

PR #621, opened 2018-10-31
Add this argument to taskify.

DenisFrezzato:
I add the capability to pass a `this` argument to `taskify`, in order to be able to *taskify* a class method.


raveclassic:
Please don't add `thisArgs` and similar constructs anywhere. It's a terrible attempt to "fix" javascript. As @gcanti noted any function can be bound to needed context _or_ you can manually bind using class-properties: 
```typescript
class Api {
  method2: () => {}
}
```

DenisFrezzato:
@gcanti Me idiot. I tried to bind, but I binded the returning function from `taskify` and not the input.

PR #623, opened 2018-11-02
WIP: Feature/coyoneda

josete89:
Hi @gcanti 

This a pr request still in progress about coyoneda, I resolved the coyoneda problem using regular javascript. But I need some guidance from you in order to do it properly on your lib. The main problem is how to create the Partial type in order to keep the composed functions to next step. Anyway I will keep trying to do it.

Regards


josete89:
> @josete89 ok but first of all, is there a practical use case for Coyoneda?

Thanks @gcanti , Basically the main purpose of Coyoneda is convert any type into a functor. Also the key thing in coyoneda is the maps are a composition of the one before this means you apply several maps in one iteration, which is more performant. ( You can look into the test ) -> https://typelevel.org/cats/api/cats/free/Coyoneda.html

josete89:
> Yes that's the theoretical purpose, but what would be a practical application in fp-ts?
> Keep in mind that fp-ts/lib/Free doesn't require a Functor instance (because AFAIK it already comprises Coyoneda).

That's true, but the thing is Coyoneda is more simple than create free monads if you only want to have the map method, which is quite handy.

>Right, but who would even use Coyoneda for that since you can just manually compose the mappings:

Also true the benefit that I see is you are deferring the execution.

raveclassic:
@josete89 I'd say such things make quite an impression on me because I'm no mathematician :) However I'd love to know how to use such theory on practice, could you show some useful examples?

josete89:
@raveclassic Imagine the following program:

````
// Business logic

const bigArray = Array(100000)
const coyo = Coyoneda.lift(bigArray).map(expensiveOperation)

// View (Side effects)

if(something){
 show(coyo.lower())
} else {
 showError()
}
```
Delays the execution until the functor is needed.

gcanti:
Closing due to inactivity

PR #637, opened 2018-11-20
Fix taskify with immutable arguments.

DenisFrezzato:
`taskify` mutates the `args` adding a callback every time the task runs, causing the wrapping promise not being resolved. This PR fixes this behaviour.


gcanti:
@DenisFrezzato good catch, thanks

Issue #616, opened 2018-10-30
add FunctorWithIndex


gcanti:
Reference implementation:  https://pursuit.purescript.org/packages/purescript-foldable-traversable/3.4.0/docs/Data.FunctorWithIndex

PR #642, opened 2018-11-22
Add FunctorWithIndex

MaximeRVY:
Hi, 

I've added a function `mapWithIndex` in Array (not sure if the naming is good) 

Here the test to understand what the function do:

```typescript
it('mapWithIndex', () => {
   assert.deepEqual(mapWithIndex([1, 2, 3], (n, i) => n + i), [1, 3, 5])
})
```

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "fp-ts",
-  "version": "1.11.0",
+  "version": "1.10.0",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {


raveclassic:
I guess it should be implemented with https://github.com/gcanti/fp-ts/issues/616

MaximeRDY:
@raveclassic Ok I'm on it

MaximeRDY:
@raveclassic @gcanti 
I've read the definition in purescript (https://pursuit.purescript.org/packages/purescript-foldable-traversable/3.4.0/docs/Data.FunctorWithIndex) and I saw FunctorWithIndex works with Array and Option
But I saw there is type parameter for the index :
```purescript
FunctorWithIndex Int Array
FunctorWithIndex Unit Maybe
```

So should I put something like that : 
```typescript
export interface FunctorWithIndex<F> {
  readonly URI: F
  readonly mapWithIndex: <A, B>(fa: HKT<F, A>, f: (i: number, a: A) => B) => HKT<F, B>
}
```
or 
```typescript
export interface FunctorWithIndex<F, I> {
  readonly URI: F
  readonly mapWithIndex: <A, B, I>(fa: HKT<F, A>, f: (i: I, a: A) => B) => HKT<F, B>
}
```

And the implementation for `Some<A>` should be something like that:

```typescript
mapWithIndex<B>(f: (_i: void ,a: A) => B): Option<B> {
  return new Some(f(this.value))
}
```

thanks

gcanti:
@MaximeRDY the type class definition is

```ts
import { Functor, Functor1 } from './Functor'
import { HKT, Type, URIS } from './HKT'

export interface FunctorWithIndex<F, I> extends Functor<F> {
  readonly mapWithIndex: <A, B>(fa: HKT<F, A>, f: (i: I, a: A) => B) => HKT<F, B>
}

export interface FunctorWithIndex1<F extends URIS, I> extends Functor1<F> {
  readonly mapWithIndex: <A, B>(fa: Type<F, A>, f: (i: I, a: A) => B) => Type<F, B>
}

// etc...
```

The instances with `I = Unit` are not interesting IMO

MaximeRDY:
@gcanti 
I've added `FunctorWithIndex`  but I'm not sure about the implementation in `Option` and the method 
`getFunctorWithIndexComposition`

gcanti:
@MaximeRDY thanks

- the `Option` instance doesn't look interesting, let's drop it
- the composition doesn't seem well defined, I would look for prior art first

src/FunctorWithIndex.ts
```diff
@@ -0,0 +1,146 @@
...
+ * `mapWithIndex` can be used to turn functions `a -> b` into functions `f a -> f b` whose argument and return types use the type
+ * constructor `f` to represent some computational context.
+ *
+ * Instances must satisfy the following laws:
+ *
+ * 1. Identity: `F.mapWithIndex(fa, (_i, a) => a) = fa`
+ * 2. Composition: `F.mapWithIndex(fa, (_i, a) => bc(ab(a))) = F.mapWithIndex(F.mapWithIndex(fa, ab), bc)`
+ *
+ * @typeclass
+ * @since 1.0.0
```
gcanti: should be `1.12.0`

src/FunctorWithIndex.ts
```diff
@@ -0,0 +1,146 @@
...
+
+export interface FunctorWithIndex2C<F extends URIS2, L, I> extends Functor2C<F, L> {
+  readonly mapWithIndex: <A, B>(fa: Type2<F, L, A>, f: (i: I, a: A) => B) => Type2<F, L, B>
+}
+
+export interface FunctorWithIndex3C<F extends URIS3, U, L, I> extends Functor3C<F, U, L> {
+  readonly URI: F
+  readonly _L: L
+  readonly _U: U
+  readonly mapWithIndex: <A, B>(fa: Type3<F, U, L, A>, f: (i: I, a: A) => B) => Type3<F, U, L, B>
```
gcanti: why

```ts
readonly URI: F
readonly _L: L
readonly _U: U
```

here?

MaximeRDY:
@gcanti 
So should I drop composition too ?

gcanti:
@MaximeRDY yes, or find some prior art / reference implementation on that, currently composition doesn't look well defined

MaximeRDY:
@gcanti 
Ok I've dropped the Composition and FunctorWithIndex in Option. 

I've found the reference of composition here :
https://github.com/purescript/purescript-functors/blob/master/src/Data/Functor/Compose.purs#L51

But my level in Purescript is beginner so it's hard for me to well understand the implementation.

gcanti:
> I've found the reference of composition here

@MaximeRDY nice, so the key is `Tuple a b`

```purescript
//                                                                              the important thing --v
instance functorWithIndexCompose :: (FunctorWithIndex a f, FunctorWithIndex b g) => FunctorWithIndex (Tuple a b) (Compose f g) where
  mapWithIndex f (Compose fga) = Compose $ mapWithIndex (mapWithIndex <<< curry f) fga
```

The porting is something like

```ts
export interface FunctorWithIndexComposition<F, FI, G, GI> extends FunctorComposition<F, G> {
//                                                     tuple --v
  readonly mapWithIndex: <A, B>(fga: HKT<F, HKT<G, A>>, f: (i: [FI, GI], a: A) => B) => HKT<F, HKT<G, B>>
}
```

and then

```ts
export function getFunctorWithIndexComposition<F, FI, G, GI>(
  F: FunctorWithIndex<F, FI>,
  G: FunctorWithIndex<G, GI>
): FunctorWithIndexComposition<F, FI, G, GI> {
  return {
    ...getFunctorComposition(F, G),
    mapWithIndex: (fga, f) => F.mapWithIndex(fga, (fi, ga) => G.mapWithIndex(ga, (gi, a) => f([fi, gi], a)))
  }
}
```

MaximeRDY:
@gcanti 
I've added the composition. 
I've added `FunctorWithIndex` to `NonEmptyArray` too.

gcanti:
Great, thanks a lot @MaximeRDY

PR #643, opened 2018-11-22
Pass read value to 'onrejected' handler for ReaderTaskEither's tryCatch

ascariandrea:
Now `tryCatch` from `ReaderTaskEither` pass the reader's value as the second argument to `onrejected` function.


gcanti:
Thanks @ascariandrea

PR #655, opened 2018-11-28
add support for refinements, closes #652

### LINKED ISSUE #652: Add type guard support to the Filterable data structures
feature

Example for `Option`

```ts
some<string | number>(42)
  .filter(t.string.is)
  .map(s => s.startsWith('i am a string'));
```

This does not work cause the the `filter` method does not support type guards.

As for th native `Array.filter` function it would be nice to have the same behaviour for all `Filterable` types.

1.11.2

When accepted I would provide a pull request.

gcanti:


src/Record.ts
```diff
@@ -514,8 +510,8 @@ export function wilt<F>(
  * @function
  * @since 1.10.0
  */
-export const filterMap = <A, B>(fa: { [key: string]: A }, f: (a: A) => Option<B>): { [key: string]: B } => {
-  const r: { [key: string]: B } = {}
+export const filterMap = <A, B>(fa: Record<string, A>, f: (a: A) => Option<B>): Record<string, B> => {
+  const r: Record<string, B> = {}
   const keys = Object.keys(fa)
   for (const key of keys) {
     const optionB = f(fa[key])
```
mlegenhausen: Should these changes be included in this pull request?

mlegenhausen:
@gcanti thanks for adding this feature

PR #678, opened 2018-12-22
FunctorWithIndex comment typo fix: `map` -> `mapWithIndex`

joshburgess:
I think this was a mistake. Just correcting what appeared to be a typo.


gcanti:
Thanks @joshburgess

Issue #408, opened 2018-04-25
Node 10 DeprecationWarning


gcanti:
Reported by @NN--- here https://github.com/gcanti/io-ts/issues/163

```ts
import * as t from 'io-ts'
console.log( t.type({}).decode(""));
```

```
[DEP0079] DeprecationWarning: Custom inspection function on Objects via .inspect() is deprecated
```

babakness:
Maybe related / useful https://nodejs.org/api/util.html#util_custom_inspection_functions_on_objects

PR #693, opened 2019-01-25
Deprecated inspect() replaced with symbol keyed function.

kolektiv:
One possible solution to #408 is to replace the inspect member functions with functions keyed on the appropriate symbol (as this pull request does). It also uses the `Symbol.for` approach, as this shouldn't error if the symbol doesn't exist, simply create a new symbol in the global symbol registry (in a browser, this would probably be the case).

I'm not personally sure of the status of support for custom inspection across browsers (or the mechanism). I can't see any sign of support, in which case this might well be effectively "no change" for browsers, but a fix for Node as the runtime.

### DIFF
--- a/src/Const.ts
+++ b/src/Const.ts
@@ -36,7 +36,7 @@ export class Const<L, A> {
   fold<B>(f: (l: L) => B): B {
     return f(this.value)
   }
-  inspect(): string {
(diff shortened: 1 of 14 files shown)


gcanti:
Superseded by https://github.com/gcanti/fp-ts/pull/892

PR #695, opened 2019-01-28
add a performance check to pipe, and switch to the 'often' most perfo…

massimosiani:
Hi @gcanti, thanks for this library.
I was wondering whether the for of syntax would bring some improvement in terms of performance, so I added a benchmark with three options based on your work.
The results are not definitive, sometimes the c-style for loop and the for of one are comparable. In any case, sometimes the for of loop is more performant, with no overlap of the error bars, up to 4%. It's not that much, but I think that having a benchmark in the codebase at least justifies the implementation used.

Let me know if you want to add it.
Thanks!
Massimo

### DIFF
--- a/perf/pipe.js
+++ b/perf/pipe.js
@@ -1,0 +1,68 @@
+var Benchmark = require('benchmark')
+
+const suite = new Benchmark.Suite()
+
+const fns = [
+  x => 2 * x,
+  x => x + 1,
+  x => x * x,
+  x => Math.log(x),
+  x => Math.exp(x)
+]
+
+const pipe1 = fns => {
+  return function(x) {
+    let y = x
+    for (const fn of fns) {
+      y = fn.call(this, y)
+    }
+    return y
+  }
+}
+
+const pipe2 = fns => {
+  const len = fns.length - 1
+  return function(x) {
+    let y = x
+    for (let i = 0; i <= len; i++) {
+      y = fns[i].call(this, y)
+    }
+    return y
+  }
+}
+
+const pipe3 = fns => {
+  return x => fns.reduce((acc, fn) => fn.call(this, acc), x)
+}
+
+const f1 = pipe1(fns)
+const f2 = pipe2(fns)
+const f3 = pipe3(fns)
+
+suite
+  .add('pipe1', function() {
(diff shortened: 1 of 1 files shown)


gcanti:
Hi,

the following tests both the `pipe(n)` (*) execution and the execution of the resulting function

```js
var Benchmark = require('benchmark')

const fns = [x => 2 * x, x => x + 1, x => x * x, x => Math.log(x), x => Math.exp(x)]

const pipe1 = fns => {
  return function(x) {
    let y = x
    for (const fn of fns) {
      y = fn.call(this, y)
    }
    return y
  }
}

const pipe2 = fns => {
  const len = fns.length - 1
  return function(x) {
    let y = x
    for (let i = 0; i <= len; i++) {
      y = fns[i].call(this, y)
    }
    return y
  }
}

const pipe3 = fns => {
  return x => fns.reduce((acc, fn) => fn.call(this, acc), x)
}

const f1 = pipe1(fns)
const f2 = pipe2(fns)
const f3 = pipe3(fns)

suite
  .add('pipe1', function() {
    pipe1(fns)
  })
  .add('pipe2', function() {
    pipe2(fns)
  })
  .add('pipe3', function() {
    pipe3(fns)
  })
  .add('f1', function() {
    f1(1)
  })
  .add('f2', function() {
    f2(1)
  })
  .add('f3', function() {
    f3(1)
  })
  .on('cycle', function(event) {
  })
  .on('complete', function() {
  })
  .run({ async: true })
```

Results on my machine

```
pipe1 x 728,374,015 ops/sec ±0.67% (89 runs sampled)
pipe2 x 731,761,745 ops/sec ±0.46% (87 runs sampled)
pipe3 x 730,961,620 ops/sec ±0.57% (88 runs sampled)
f1 x 8,768,588 ops/sec ±0.58% (87 runs sampled)
f2 x 9,453,816 ops/sec ±0.56% (86 runs sampled)
f3 x 9,090,939 ops/sec ±0.64% (86 runs sampled)
```

(*) note that my `pipe3` implementation is slightly different

gcanti:
> Would you mind to add the perf tests to the codebase for future reference?

Sure, thanks. Could you please squash the commits?

PR #698, opened 2019-02-02
feat(build): implement rollup builds

baetheus:
- Terser minify bundles
- Provide es6 modules bundle and commonjs bundle
- Provide sourcemaps for each
- Output ts declaration files to lib


baetheus:
It looks like definitelytyped-header-parser is a dependency of dtslint but is not published to npm. Instead it is referenced as a git dependency. I don't have any issues with the tests locally, but I'm guessing travis is not resolving that dep correctly. @gcanti It looks like you've already forked dtslint and are looking into it. Let me know if you want further help.

gcanti:
@baetheus thanks for this PR but I'm not acquainted with rollup so I'm not sure what's the goal / benefit here.

Looks like this build will output only two `.js` files

- `lib/index.js`
- `lib/index.es.js`

However all `fp-ts` modules are supposed to be (also) consumed individually

```ts
import { Option } from 'fp-ts/lib/Option'
```

baetheus:
The goals I had with this PR were:

1. Provide a tree shakeable es6 bundle in addition to the current commonjs target, since es6 module exports [are the spec](https://www.ecma-international.org/ecma-262/6.0/#sec-exports) and showing up in [node](https://github.com/nodejs/node-eps/blob/master/002-es-modules.md) and [browsers](https://caniuse.com/#feat=es6-module).
2. Provide sourcemaps so debugging production apps is easier to reason about.
3. Ensure that the commonjs build is backwards compatible with previous releases.

gcanti:
I think this is covered by https://github.com/gcanti/fp-ts/pull/814, closing for now

PR #705, opened 2019-02-06
fix compile issue with typescript@next (fixes #704)

### LINKED ISSUE #704: Code base cannot compile with typescript@next
Something to do with type checking.
I'll provide a PR

sledorze:
That should make it work.

Note: Sorry for the CI noise; I'm still having some issue with the instersection of well-known-but-I-dont-remember issue with running tests because of Localstorage and I-m-out-of-time-before-lunch-and-using-the-CI-optimistically.


gcanti:
@sledorze I'm fine with this change, is just a type annotation, but I would try to find a way to not run the tests against `typescript@next` as soon as possible, which means either

- replace dtslint with something else (what?)
- configure dtslint (how?)
- switch to my custom fork

gcanti:
@sledorze master is fixed now, thanks

PR #714, opened 2019-02-09
array.indexes

user471:


gcanti:
@user471 thanks for this PR, looks like this use case is already covered by `range`

```ts
import { range } from 'fp-ts/lib/Array'

range(0, 4) // => [ 0, 1, 2, 3, 4 ]
```

user471:
@gcanti yes, but when I need indexes and have some variable like length I have to do
`range(0, length -1)`
`indexes(length)` would be better and most of the time I need `indexes` instead of `range`, so  `fp-ts` `range` doesn't very useful for me and I have to use lodash `_.range(length)`

PR #715, opened 2019-02-10
Add Map module & tests

joshburgess:
This PR does the following:

- Adds a new Map module for working with ES6 Maps
- Adds a getMapSemigroup function to the Semigroup module
- Adds a getMapMonoid function to the Monoid module
- Adds new tests for all of the above things

I want to help work on adding a String module next. :)


package.json
```diff
@@ -9,8 +9,8 @@
   "typings": "lib/index.d.ts",
   "scripts": {
     "lint": "tslint -p tsconfig.tslint.json src/**/*.ts test/**/*.ts",
-    "jest": "jest --ci",
-    "jest-coverage": "jest --ci --coverage",
+    "jest": "jest --testURL=\"http://localhost\" --ci",
```
texastoland: Intentional?

joshburgess: I had to add this because of an error I was getting with Jest:

`SecurityError: localStorage is not available for opaque origins at Window.get localStorage [as localStorage]`

See here: https://stackoverflow.com/questions/51554366/jest-securityerror-localstorage-is-not-available-for-opaque-origins

jethrolarson: I think you can put that in the jest config section rather than via cli arg

src/Map.ts
```diff
@@ -0,0 +1,565 @@
...
+
+/**
+ * Insert or replace a key/value pair in a Map
+ *
+ * @since 1.14.0
+ */
+export function insert<KS, K extends KS, A>(k: K, a: A, d: Map<KS, A>): Map<KS | K, A>
+export function insert<K, A>(k: K, a: A, d: Map<K, A>): Map<K, A> {
+  const m = new Map(d)
+  m.set(k, a)
```
sledorze: Please try to be conservative and return the original Map if the operation is a noop (value already present at that key, tested using strict equality)

src/Map.ts
```diff
@@ -0,0 +1,565 @@
...
+
+/**
+ * Delete a key and value from a map
+ *
+ * @since 1.14.0
+ */
+export function remove<KS, K extends KS, A>(k: K, d: Map<KS, A>): Map<KS extends K ? KS : Exclude<KS, K>, A>
+export function remove<K, A>(k: K, d: Map<K, A>): Map<K, A> {
+  const m = new Map(d)
+  m.delete(k)
```
sledorze: Same here, let be conservative in case of a noop

src/Map.ts
```diff
@@ -0,0 +1,565 @@
...
+      m.set(key, optionB.value)
+    }
+  }
+  return m
+}
+
+/**
+ * @since 1.14.0
+ */
+export function filterWithIndex<K, A>(fa: Map<K, A>, p: (key: K, a: A) => boolean): Map<K, A> {
```
sledorze: would also be awesome to be conservation regarding noop.

src/Map.ts
```diff
@@ -0,0 +1,565 @@
...
+ * @since 1.14.0
+ */
+export const keys = <K, A>(d: Map<K, A>): Array<K> => {
+  return Array.from(d.keys())
+}
+
+/**
+ * @since 1.14.0
+ */
+export const unsafeLookup = <K, A>(k: K, d: Map<K, A>): A => {
```
giogonzo: I'd avoid exporting this unsafe function?

gcanti:
@joshburgess thanks for this PR

Not sure I understand what's the problem with jest (`--testURL=\"http://localhost\"`), I don't get any error from jest with the current configuration.

I propose the following changes:

- drop `getMapSemigroup` in `Semigroup`
- move `getMapMonoid` from `Monoid` to `Map`, needs a `Setoid` constraint for `K`
- add `member` function (needs a `Setoid` constraint), see https://github.com/purescript/purescript-ordered-collections/blob/67a0263f8b1009a26397a811da300f0ef16813c4/src/Data/Map/Internal.purs#L403
- the following functions need a `Ord` constraint for `K`
  - `keys`
  - `collect`
  - `toArray`
- the following functions need a `Setoid` constraint for `K`
  - `insert`
  - `remove`
  - `pop`
  - `isSubmap`
  - `getSetoid`
  - `lookup`
  - `fromFoldable`
- missing overloadings in `toUnfoldable`
- why the overloadings in `insert` and `remove`?
- missing overloading in `filter`
- check if `unsafeLookup` is still useful after all the proposed changes

Also looks like some functions could be optimized using an iterator instead of `keys` + `for ... of`.

Example

```ts
export const compact = <K, A>(fa: Map<K, Option<A>>): Map<K, A> => {
  const m = new Map<K, A>()
  const entries = fa.entries()
  let e: IteratorResult<[K, Option<A>]>
  while (!(e = entries.next()).done) {
    const [k, oa] = e.value
    if (oa.isSome()) {
      m.set(k, oa.value)
    }
  }
  return m
}
```

joshburgess:
> @joshburgess thanks for this PR
> 
> Not sure I understand what's the problem with jest (`--testURL=\"http://localhost\"`), I don't get any error from jest with the current configuration.

I get 68 errors saying `SecurityError: localStorage is not available for opaque origins`. Do you have a jest config file that locally that's in the .gitignore?

gcanti:
@joshburgess no

src/Map.ts
joshburgess: Hmmm. So, you want me to always perform a `get` first to check the value?

joshburgess:
> @joshburgess no

Hmm, I'm not sure what the problem is. If you google for it, there are many issues about it on GitHub. They al suggest adding this `testURL` config option in a Jest config file or in the CLI command.

I can't seem to run the tests without it.

https://github.com/facebook/jest/issues/6769

https://github.com/facebook/jest/issues/6766

I'll make the other suggested changes tomorrow.

src/Map.ts
sledorze: @joshburgess yes

texastoland:
@joshburgess Can you check your `jsdom` version is `11.10.0` (unaffected) from `package-lock.json`. facebook/jest#6792 fixed `testURL` and released it as `jes@23.5.0`.

joshburgess:
> @joshburgess Can you check your `jsdom` version is `11.10.0` (unaffected) from `package-lock.json`. [facebook/jest#6792](https://github.com/facebook/jest/pull/6792) fixed `testURL` and released it as `jest@23.5.0`.

References to `jsdom` in the package-lock.json:

```
"jest-environment-jsdom": "^22.4.1",
```

```
"jest-config": {
              "version": "22.4.4",
              "resolved": "https://registry.npmjs.org/jest-config/-/jest-config-22.4.4.tgz",
              ...
              "requires": {
                ...
                "jest-environment-jsdom": "^22.4.1",
                ...
              }
            },
```

```
"jest-environment-jsdom": {
      "version": "22.4.3",
      "resolved": "https://registry.npmjs.org/jest-environment-jsdom/-/jest-environment-jsdom-22.4.3.tgz",
      "integrity": "sha512-FviwfR+VyT3Datf13+ULjIMO5CSeajlayhhYQwpzgunswoaLIPutdbrnfUHEMyJCwvqQFaVtTmn9+Y8WCt6n1w==",
      "dev": true,
      "requires": {
        "jest-mock": "^22.4.3",
        "jest-util": "^22.4.3",
        "jsdom": "^11.5.1"
      }
    },
```

```
"jsdom": {
      "version": "11.10.0",
      "resolved": "https://registry.npmjs.org/jsdom/-/jsdom-11.10.0.tgz",
      ...
}
```

gcanti:
@joshburgess have you tried `jest@23.6.0`? If it solves the problem we could upgrade

joshburgess:
> @joshburgess have you tried `jest@23.6.0`? If it solves the problem we could upgrade

Just tried it. Yes, this fixed it.

joshburgess:
> * drop `getMapSemigroup` in `Semigroup`
> * move `getMapMonoid` from `Monoid` to `Map`, needs a `Setoid` constraint for `K`

Sure, I can do this, but can you explain why? There is both a `getDictionarySemigroup` and `getDictionaryMonoid`, and they are in the `Semigroup` and `Monoid` modules, not the `Record` module.

> * why the overloadings in `insert` and `remove`?

I'm not 100% sure, to be honest. I was just following what had already been done for `insert` and `remove` in the `Record` module.

gcanti:
@joshburgess Legacy. They predate the `Record` module and, as pointed out by @giogonzo, their names are not consistent with the convention used in TypeScript, so I'm going to deprecate them before releasing 1.14.0

joshburgess:
@gcanti Can you explain how I should be using the `Setoid<K>` for `insert` and `remove`? Can I not use the native `.has` to check for the existence of the key?

gcanti:
@joshburgess native `has` uses reference equality (`===`) which makes `Map` a lot less useful. You can take the `Set` module as a reference.

So for example let's see `lookup` (not tested)

```ts
/**
 * Lookup the value for a key in a `Map`.
 * If the result is a `Some`, the existing key is also returned.
 */
export function lookupWithKey<K>(S: Setoid<K>): <A>(k: K, m: Map<K, A>) => Option<[K, A]> {
  return <A>(k: K, m: Map<K, A>) => {
    const entries = m.entries()
    let e: IteratorResult<[K, A]>
    while (!(e = entries.next()).done) {
      const [ka, a] = e.value
      if (S.equals(ka, k)) {
        return some(tuple(k, a))
      }
    }
    return none
  }
}

/**
 * Lookup the value for a key in a `Map`.
 */
export function lookup<K>(S: Setoid<K>): <A>(k: K, m: Map<K, A>) => Option<A> {
  const lookupWithKeyS = lookupWithKey(S)
  return (k, m) => lookupWithKeyS(k, m).map(([_, a]) => a)
}
```

and ~`insert`~ `insertAt` (for consistency with `Array.insertAt`) (not tested)

```ts
export function insertAt<K>(S: Setoid<K>): <A>(k: K, a: A, m: Map<K, A>) => Map<K, A> {
  const lookupS = lookupWithKey(S)
  return (k, a, m) => {
    const found = lookupS(k, m)
    if (found.isNone()) {
      const r = new Map(m)
      // add the value
      r.set(k, a)
      return r
    } else if (found.value[1] !== a) {
      const r = new Map(m)
      // overwrite the existing key
      r.set(found.value[0], a)
      return r
    }
    return m
  }
}
```

@sledorze @giogonzo do you agree?

joshburgess:
@gcanti I think I addressed all of your feedback.

The only thing I didn't change that you suggested was adding overloads to `toUnfoldable`. I don't see any examples of other modules having overloads for this function. What would they be?

Also, I had to add an `Ord` constraint for `K` to `toUnfoldable` so that I could pass it through to `toArray` within it.

Also, I removed `unsafeLookup`.

Could you look over the changes to see whether or not they look good before I go to update the tests?

gcanti:
> Could you look over the changes

Sure, I need some time though. In the meanwhile..

> adding overloads to `toUnfoldable`. I don't see any examples

see [this PR](https://github.com/gcanti/fp-ts/pull/717)

gcanti:
@joshburgess

Renaming

- rename `member` to `has` (since the behavior is aligned with the native method)
- rename `insertAt` to `insert` (sorry I changed my mind, the naming should be consistent with `Record` not `Array`)
- rename `deleteAt` to `remove` (idem)

Group constraints

- toUnfoldable signature: `export const toUnfoldable = <F, K>(unfoldable: Unfoldable<F>, O: Ord<K>) => <A>(d: Map<K, A>): HKT<F, [K, A]> => { ... }`
- getSetoid signature: `export const getSetoid = <K, A>(SK: Setoid<K>, SA: Setoid<A>): Setoid<Map<K, A>> => { ... }`
- getMonoid signature: `export const getMonoid = <K, A>(SK: Setoid<K>, SA: Semigroup<A>): Monoid<Map<K, A>> => { ... }`

Optimizations

There are many low hanging fruits like this:

Current

```ts
export const lookup = <K>(S: Setoid<K>) => <A>(k: K, m: Map<K, A>): Option<A> => {
  const lookupWithKeyS = lookupWithKey(S) // <= created each time
  return lookupWithKeyS(k, m).map(([_, a]) => a)
}
```

Better

```ts
export function lookup<K>(S: Setoid<K>): <A>(k: K, m: Map<K, A>) => Option<A> {
  const lookupWithKeyS = lookupWithKey(S) // <= created once
  return (k, m) => lookupWithKeyS(k, m).map(([_, a]) => a)
}
```

joshburgess:
> * rename `member` to `has` (since the behavior is aligned with the native method)

This seems weird to me. The Array module doesn't currently do this. It uses `member` despite the native function being called `Array.includes`.

I think I would prefer `member` since this relies on `Setoid<K>`, unlike the native `map.has`.

joshburgess:
Okay, I made all of those changes now except for renaming `member`. I'm going to wait until you reply about that one, as I think `member` makes more sense.

This is looking pretty good now though. I will update the types soon... possibly tonight if I have enough time.

gcanti:
@joshburgess A function named `member` would rather test whether a **value** is a member of a map

Map.member

```ts
<A>(S: Setoid<A>) => (a: A, m: Map<K, A>): boolean
```

Array.member

```ts
<A>(S: Setoid<A>) => (as: Array<A>, a: A): boolean
```

Set.member

```ts
<A>(S: Setoid<A>) => (x: Set<A>) => (a: A): boolean
```

Foldable2v.member

```ts
(S: Setoid<A>, F: Foldable2v<F>): (a: A, fa: HKT<F, A>) => boolean
```

gcanti:
Actually we could take this opportunity to make the signatures more consistent, I'll send a PR

joshburgess:
> @joshburgess A function named `member` would rather test whether a **value** is a member of a map

We could change it to take in the value and do a comparison if you want?

Something like:

```ts
export const member = <K, A>(SK: Setoid<K>, SA: Setoid<A>): ((k: K, a: A, m: Map<K, A>) => boolean) => {
  const lookupSK = lookup(SK)
  return (k, a, m) => {
    const o = lookupSK(k, m)
    return o.isSome() && SA.equals(a, o.value)
  }
}
```

joshburgess:
How about this?

```
/**
 * Test whether a key exists in a map
 *
 * @since 1.14.0
 */
export const has = <K>(S: Setoid<K>): (<A>(k: K, m: Map<K, A>) => boolean) => {
  const lookupS = lookup(S)
  return (k, m) => lookupS(k, m).isSome()
}

/**
 * Test whether a key/value pair is a member of a map
 *
 * @since 1.14.0
 */
export const member = <K, A>(SK: Setoid<K>, SA: Setoid<A>): ((k: K, a: A, m: Map<K, A>) => boolean) => {
  const lookupSK = lookup(SK)
  return (k, a, m) => {
    const o = lookupSK(k, m)
    return o.isSome() && SA.equals(a, o.value)
  }
}
```

joshburgess:
Okay, I updated it to include both `has` and `member`:

```ts
/**
 * Test whether or not a key exists in a map
 *
 * @since 1.14.0
 */
export const has = <K>(S: Setoid<K>): (<A>(k: K, m: Map<K, A>) => boolean) => {
  const lookupS = lookup(S)
  return (k, m) => lookupS(k, m).isSome()
}

/**
 * Test whether or not a key/value pair is a member of a map
 *
 * @since 1.14.0
 */
export const member = <K, A>(SK: Setoid<K>, SA: Setoid<A>): ((k: K, a: A, m: Map<K, A>) => boolean) => {
  const lookupSK = lookup(SK)
  return (k, a, m) => {
    const o = lookupSK(k, m)
    return o.isSome() && SA.equals(a, o.value)
  }
}
```

Is that ok?

gcanti:
`member` looks weird to me (I'm used to `Foldable2v.member`)

joshburgess:
> `member` looks weird to me (I'm used to `Foldable2v.member`)

You think it should only check for the existence of the value and not the key?

joshburgess:
It looks like Haskell also only checks for the existence of the key for its `member` functions for Maps: https://www.stackage.org/haddock/lts-13.2/containers-0.6.0.1/Data-Map-Strict.html#v:member

gcanti:
> it should only check for the existence of the value and not the key?

Yes, it would be aligned with all the other `member`s functions (well, `isMember` if we consider https://github.com/gcanti/fp-ts/pull/729).

The point is that it should be theoretically derivable from `Foldable2v.member`

```ts
import { member } from 'fp-ts/lib/Foldable2v'
import { array } from 'fp-ts/lib/Array'
import { tree } from 'fp-ts/lib/Tree'
import { Setoid } from 'fp-ts/lib/Setoid'

// const isArrayMember: <A>(S: Setoid<A>) => (a: A, fa: A[]) => boolean
const isArrayMember = <A>(S: Setoid<A>) => member(S, array)

// const isTreeMember: <A>(S: Setoid<A>) => (a: A, fa: Tree<A>) => boolean
const isTreeMember = <A>(S: Setoid<A>) => member(S, tree)

// etc...
```

joshburgess:
I don't think Record or Map currently implement Foldable/Foldable2v though?

How should I proceed?

Should I just delete `member` and leave only `has`? Or should I copy what you did there with `isMember`? (iterate over the values and just look for the existence of only the passed in value)

gcanti:
> I don't think Record or Map currently implement Foldable/Foldable2v though?

Right, that's why I said "theoretically".

> Should I just delete `member` and leave only `has`?

It's up to you. My only advice is: **IF** we add a function named `isMember`, then it would be nice if its signature is aligned with all the others.

joshburgess:
- renamed `member` to `isMember` & changed its implementation
- adjusted `reduce`, `reduceWithKey`, `foldr`, & `foldrWithKey` to take in an `Ord<K>` so that they can sort correctly (These functions use `.sort()` in the Record module)
- added an alternate `keys_` function which returns a Set instead of an Array (similar to `keys` from PureScript's Data.Map https://pursuit.purescript.org/packages/purescript-ordered-collections/1.5.0/docs/Data.Map#v:keys)
- fixed a logic mistake that made it into `remove` earlier that affected `remove` & `pop`
- rearranged the order of the tests to match the order of the exports in the Map module
- added tests for all of the things that previously weren't being tested


Other than that, I have one quick comment about `isMember`... I believe that you changed this function's name from `member` to `isMember` to try to disambiguate between it and `member` in PureScript/Haskell... However, I'm not sure that the name is different enough. I think a better name would be something like `contains` or `isValueOf` to more clearly convey the difference. `isMember` still sort of implies the same idea as `member`, and `member` only checks for keys in PureScript/Haskell.

But maybe this is okay & just feels weird because it is sort of inconsistent in Haskell/PureScript? For other data structures, `member` does check the value, but it only checks the `key` for Maps...

gcanti:
> added an alternate `keys_`

The constraint `Ord` doesn't seem required, isn't just

```ts
export const keys_ = <A>(m: Map<K, A>): Set<K> => {
  return new Set(m.keys())
}
```

Perhaps we can name them: `keys` (Array) / `keysSet` (Set)

> and `member` only checks for keys in PureScript/Haskell

`member` is named `elem` in `Foldable` (and renamed to `member` in `Foldable2v`)

- `elem`/`member` are a porting of [`elem` in `purescript-foldable-traversable`](https://github.com/purescript/purescript-foldable-traversable/
blob/29d5b873cc86f17e0082d777629819a96bdbc7a1/src/Data/Foldable.purs#L334) which tests the value
- [`member` in `purescript-ordered-collections` (Set)](https://github.com/purescript/purescript-ordered-collections/blob/master/src/Data/Set.purs#L117) tests the value
- [`member` in `purescript-ordered-collections` (Map)](https://github.com/purescript/purescript-ordered-collections/blob/master/src/Data/Map/Internal.purs#L403) **tests the key though**

I'm absolutely fine to change all the `isMember` to `contains` or `includes` or something else, I'm more interested in signature consistency in this case.

joshburgess:
> > added an alternate `keys_`
> 
> The constraint `Ord` doesn't seem required, isn't just
> 
> ```ts
> export const keys_ = <A>(m: Map<K, A>): Set<K> => {
>   return new Set(m.keys())
> }
> ```

You're right. I'll update it. 

> Perhaps we can name them: `keys` (Array) / `keysSet` (Set)

Sounds good. I'll rename `keys_` to `keysSet`. 

> Perhaps we can name them: `keys` (Array) / `keysSet` (Set)
> 
> > and `member` only checks for keys in PureScript/Haskell
> 
> `member` is named `elem` in `Foldable` (and renamed to `member` in `Foldable2v`)
> 
> * `elem`/`member` are a porting of [`elem` in `purescript-foldable-traversable`](https://github.com/purescript/purescript-foldable-traversable/
>   blob/29d5b873cc86f17e0082d777629819a96bdbc7a1/src/Data/Foldable.purs#L334) which tests the value
> * [`member` in `purescript-ordered-collections` (Set)](https://github.com/purescript/purescript-ordered-collections/blob/master/src/Data/Set.purs#L117) tests the value
> * [`member` in `purescript-ordered-collections` (Map)](https://github.com/purescript/purescript-ordered-collections/blob/master/src/Data/Map/Internal.purs#L403) **tests the key though**
> 
> I'm absolutely fine to change all the `isMember` to `contains` or `includes` or something else, I'm more interested in signature consistency in this case.

Hmmm, I see.

I think I'd prefer `has` to be named `member` and `isMember` to be named something else, but I'm not sure what the best name would be. Anyone else have any thoughts? @texastoland @sledorze @giogonzo 

`elem` would make sense to me if we could make `Record` and `Map` have `Foldable`/`Foldable2v` instances, but it would probably be weird/confusing to use `elem` without them... I'm not sure.

src/Map.ts
```diff
@@ -46,7 +46,7 @@ export const isMember = <A>(SA: Setoid<A>): (<K>(a: A, m: Map<K, A>) => boolean)
     let out: boolean = false
     const as = m.values()
     let e: IteratorResult<A>
-    while (!(e = as.next()).done) {
+    while (!(e = as.next()).done && !out) {
```
gcanti: I think that it could just `return` here

```ts
const as = m.values()
let e: IteratorResult<A>
while (!(e = as.next()).done) {
  if (SA.equals(a, e.value)) {
    return true
  }
}
return false
```

joshburgess:
@gcanti Sorry, fixed.

joshburgess:
@texastoland @sledorze @giogonzo @jethrolarson I'm still curious about your thoughts on the potential renaming of `isMember` above if you have any.

However, @gcanti and I spoke on Slack a bit about possibly registering this Map module as a HKT like how the Array module currently does (and Set and Record currently do _not_).

We haven't fully decided, so I'm curious what you think. 

It was not done for Record, because augmenting the native Object/ObjectInstance TS types would be required, and that would be very intrusive.

However, Map & Set could be registered as HKTs like Array currently is without too much intrusion. The minor issue with registering these native, unwrapped types as HKTs is this:

```ts
const x = new Map()

// these don't exist at runtime, but TS thinks they do
x._A
x._URI
```

However, this is currently how Array works, and it doesn't seem to be too big of a problem. What do you think?

For `Map`, it would work like this:

```ts
declare global {
  interface Map<K, V> {
    _URI: URI // phantom
    _A: V // phantom
  }
}

declare module './HKT' {
  interface URI2HKT2<L, A> {
    Map: Map<L, A>
  }
}

export const URI = 'Map'

export type URI = typeof URI
```

However, the extra `Ord` & `Setoid` constraints means that many of the functions in this module don't have the right signatures to satisfy some of the type class requirements. For type classes like: `Foldable`/`Foldable2v`/`FoldableWithIndex`, `FilterableWithIndex`, `Witherable`, etc., we wouldn't be able to statically export the instances. Instead, we could create functions that would return those instances when provided with an `Ord` constraint.

For example:

```ts
export const getFoldable = <K>(O: Ord<K>): Foldable2v2C<URI, K> => {
  return {
    // type class requirements here
  }
}
```

What do you all think? Is it worth the trouble? It does add a lot of complexity, but I think it might be worth it.

giogonzo:
@joshburgess 

> I'm still curious about your thoughts on the potential renaming of `isMember` above if you have any.

Sorry I don't - not really familiar with any other fp "stdlib" other than fp-ts. As far as I can tell, two guiding principles up to now have been 1) stick to the original purescript names, if consistent 2) respect TS ecosystem lingo (e.g. the case for `Record`: even if not the proper name, TS people understand that). If neither of the above apply here (it appears to me that the PS counterpart is not 100% consistent?), I guess it's just a matter of taste, and I don't know what to add to the discussion.

> However, this is currently how Array works, and it doesn't seem to be too big of a problem. What do you think?

😲 wow never really noticed this (`[]._URI` being valid TS) - so I guess not a problem in practice, even more so for `Set` or `Map`

> It was not done for `Record`, because augmenting the native `Object`/`ObjectInstance` TS types would be required, and that would be very intrusive.

Why would this be worse exactly? I guess it would apply not only to `Record`s but also to any other JS object?

gcanti:
@giogonzo after talking with @joshburgess I tried to drop the module augmentation in `Array` and, except for finding a "bug" in `Whiterable` (https://github.com/gcanti/fp-ts/pull/738) (which is a very good thing), everything seems to just work. 

Keep in mind that the current encoding is pretty old ([developed 2 year ago](https://github.com/gcanti/fp-ts/blob/63644f5fc1f2fc7e3a345a5e4ffe76b7c07e5453/src/Arr.ts#L19), with [TypeScript 2.2](https://github.com/gcanti/fp-ts/blob/63644f5fc1f2fc7e3a345a5e4ffe76b7c07e5453/package.json#L36)), so the conditions that led me to it may be _slightly_ changed in the meanwhile..

joshburgess:
@gcanti what do you mean? you're saying Array doesn't require the registering/augmentation now?

gcanti:
> you're saying Array doesn't require the registering/augmentation now?

@joshburgess yes, I put up a branch (`drop-phantom-fields`) where

- `Array` modue augmentation is removed
- all "phantom fields" are removed, i.e

```diff
export class State<S, A> {
-  readonly _A!: A
-  readonly _L!: S
-  readonly _URI!: URI
  constructor(readonly run: (s: S) => [A, S]) {}
```

everything still seems to work., i.e.

- build
- tests
- type-level tests 

(but we would need a test on a real codebase)

joshburgess:
@gcanti Ok, I rewrote a lot of the module & tests to switch to those `get_` functions that return the type class instances. Can you look at what I did to make sure it looks good?

Also, to write the instances for TraversableWithIndex, I had to add the following types:

`TraverseWithIndex2`
`TraverseWithIndex2C`
`TraversableWithIndex2`
`TraversableWithIndex2C`

All the tests passed. So, I think I wrote them correctly, but can you look them over to make sure?

I think this should be just about ready to merge now.

gcanti:
@joshburgess Looks like the tests are not running with ` --clearCache`

A general observation: the tests would be more solid if `K` is not a primitive, for example

```ts
interface User {
  id: number
}

const setoidUser = contramap((u: User) => u.id, setoidNumber)

it('remove', () => {
  const m = new Map<User, number>([[{ id: 1 }, 1], [{ id: 2 }, 2]])
  assert.deepStrictEqual(M.remove(setoidUser)({ id: 1 }, m), new Map<User, number>([[{ id: 2 }, 2]]))
  // m shouldn't change
  assert.deepStrictEqual(m, new Map<User, number>([[{ id: 1 }, 1], [{ id: 2 }, 2]]))
})
```

so we can see the constraints (`Setoid` and `Ord`) in action

src/Map.ts
```diff
@@ -0,0 +1,825 @@
...
+ * @since 1.14.0
+ */
+export const collect = <K>(O: Ord<K>): (<A, B>(m: Map<K, A>, f: (k: K, a: A) => B) => Array<B>) => {
+  const lookupWithKeyO = lookupWithKey(O)
+  const keysO = keys(O)
+  return <A, B>(m: Map<K, A>, f: (k: K, a: A) => B): Array<B> => {
+    const out: Array<B> = []
+    const ks = keysO(m)
+    for (let key of ks) {
+      const o = lookupWithKeyO(key, m)
```
gcanti: this looks O(n^2), maybe we can leverage `get`

```diff
-const o = lookupWithKeyO(key, m)
-if (o.isSome()) {
-  out.push(f(key, o.value[1]))
-}
+out.push(f(key, m.get(key)!))
```

src/Map.ts
```diff
@@ -0,0 +1,825 @@
...
+ * @since 1.14.0
+ */
+export const lookupWithKey = <K>(S: Setoid<K>): (<A>(k: K, m: Map<K, A>) => Option<[K, A]>) => {
+  return <A>(k: K, m: Map<K, A>) => {
+    const entries = m.entries()
+    let e: IteratorResult<[K, A]>
+    while (!(e = entries.next()).done) {
+      const [ka, a] = e.value
+      if (S.equals(ka, k)) {
+        return some(tuple(k, a))
```
gcanti: typo

```diff
-return some(tuple(k, a))
+return some(tuple(ka, a))
```

src/Map.ts
```diff
@@ -0,0 +1,825 @@
...
+ *
+ * @since 1.14.0
+ */
+export const remove = <K>(S: Setoid<K>): (<A>(k: K, m: Map<K, A>) => Map<K, A>) => {
+  const lookupS = lookupWithKey(S)
+  return (k, m) => {
+    const found = lookupS(k, m)
+    if (found.isSome()) {
+      const r = new Map(m)
+      m.delete(k)
```
gcanti: shouldn't change `m` and should use the existing key

```diff
-      m.delete(k)
+      r.delete(found.value[0])
```

src/Map.ts
```diff
@@ -0,0 +1,825 @@
...
+ * @since 1.14.0
+ */
+export const isSubmap = <K, A>(SK: Setoid<K>, SA: Setoid<A>): ((d1: Map<K, A>, d2: Map<K, A>) => boolean) => {
+  const lookupWithKeyS = lookupWithKey(SK)
+  return (d1: Map<K, A>, d2: Map<K, A>): boolean => {
+    const entries = d1.entries()
+    let e: IteratorResult<[K, A]>
+    while (!(e = entries.next()).done) {
+      const [k, a] = e.value
+      const d1OptA = lookupWithKeyS(k, d1)
```
gcanti: I think that this lookup can be removed, you can already read `[k, a]`

src/Map.ts
```diff
@@ -0,0 +1,825 @@
...
+ */
+export const getMonoid = <K, A>(SK: Setoid<K>, SA: Semigroup<A>): Monoid<Map<K, A>> => {
+  return {
+    concat: (mx, my) => {
+      const r = new Map(mx)
+      const entries = my.entries()
+      let e: IteratorResult<[K, A]>
+      while (!(e = entries.next()).done) {
+        const [k, a] = e.value
+        const lookupWithKeyS = lookupWithKey(SK)
```
gcanti: this can be moved outside the `return`

src/Map.ts
```diff
@@ -0,0 +1,825 @@
...
+        const mxOptA = lookupWithKeyS(k, mx)
+        const myOptA = lookupWithKeyS(k, my)
```
gcanti: this can be removed, you can read `[k, a]`

src/Map.ts
```diff
@@ -0,0 +1,825 @@
...
+      return r
+    },
+    empty
+  }
+}
+
+/**
+ * @since 1.14.0
+ */
+function filter<K, A, B extends A>(fa: Map<K, A>, p: Refinement<A, B>): Map<K, B>
```
gcanti: overloadings can be removed (is not exported)

src/Map.ts
```diff
@@ -0,0 +1,825 @@
...
+/**
+ * @since 1.14.0
+ */
+const map_ = <K, A, B>(fa: Map<K, A>, f: (a: A) => B): Map<K, B> => mapWithKey(fa, (_, a) => f(a))
+
+/**
+ * @since 1.14.0
+ */
+const reduce = <K>(O: Ord<K>): (<A, B>(fa: Map<K, A>, b: B, f: (b: B, a: A) => B) => B) => {
+  return (fa, b, f) => reduceWithKey(O)(fa, b, (_, b, a) => f(b, a))
```
gcanti: `reduceWithKey(O)` can be moved outside

src/Map.ts
```diff
@@ -0,0 +1,825 @@
...
+}
+
+/**
+ * @since 1.14.0
+ */
+const foldMap = <K>(O: Ord<K>): (<M>(M: Monoid<M>) => <A>(fa: Map<K, A>, f: (a: A) => M) => M) => M => (fa, f) =>
+  foldMapWithKey(O)(M)(fa, (_, a) => f(a))
```
gcanti: `foldMapWithKey(O)` can be moved outside

src/Map.ts
```diff
@@ -0,0 +1,825 @@
...
+
+/**
+ * @since 1.14.0
+ */
+const foldr = <K>(O: Ord<K>): (<A, B>(fa: Map<K, A>, b: B, f: (a: A, b: B) => B) => B) => {
+  return (fa, b, f) => foldrWithKey(O)(fa, b, (_, a, b) => f(a, b))
```
gcanti: `foldrWithKey(O)` can be moved outside

src/Map.ts
```diff
@@ -0,0 +1,825 @@
...
+ * @since 1.14.0
+ */
+export const singleton = <K, A>(k: K, a: A): Map<K, A> => {
+  return new Map<K, A>([[k, a]])
+}
+
+/**
+ * @since 1.14.0
+ */
+function traverseWithKey<F extends URIS3>(
```
gcanti: overloadings can be removed

src/Map.ts
```diff
@@ -0,0 +1,825 @@
...
+      fm = F.ap(F.map(fm, m => (b: B) => new Map(m).set(key, b)), f(key, a))
+    }
+    return fm
+  }
+}
+
+/**
+ * @since 1.14.0
+ */
+function traverse<F extends URIS3>(
```
gcanti: overloadings can be removed

src/Map.ts
```diff
@@ -0,0 +1,825 @@
...
+function traverse<F>(F: Applicative<F>): <K, A, B>(ta: Map<K, A>, f: (a: A) => HKT<F, B>) => HKT<F, Map<K, B>>
+function traverse<F>(F: Applicative<F>): <K, A, B>(ta: Map<K, A>, f: (a: A) => HKT<F, B>) => HKT<F, Map<K, B>> {
+  const traverseWithKeyF = traverseWithKey(F)
+  return (ta, f) => traverseWithKeyF(ta, (_, a) => f(a))
+}
+
+/**
+ * @since 1.14.0
+ */
+function sequence<F extends URIS3>(
```
gcanti: overloadings can be removed

src/Map.ts
```diff
@@ -0,0 +1,825 @@
...
+  return {
+    left,
+    right
+  }
+}
+
+/**
+ * @since 1.14.0
+ */
+function wither<F extends URIS3>(
```
gcanti: overloadings can be removed

src/Map.ts
```diff
@@ -0,0 +1,825 @@
...
+function wither<F>(F: Applicative<F>): (<K, A, B>(wa: Map<K, A>, f: (a: A) => HKT<F, Option<B>>) => HKT<F, Map<K, B>>)
+function wither<F>(F: Applicative<F>): (<K, A, B>(wa: Map<K, A>, f: (a: A) => HKT<F, Option<B>>) => HKT<F, Map<K, B>>) {
+  const traverseF = traverse(F)
+  return (wa, f) => F.map(traverseF(wa, f), compact)
+}
+
+/**
+ * @since 1.14.0
+ */
+function wilt<F extends URIS3>(
```
gcanti: overloadings can be removed

src/Map.ts
```diff
@@ -0,0 +1,825 @@
...
+  F: Foldable1<F>
+): <A>(ta: Type<F, [K, A]>, f: (existing: A, a: A) => A) => Map<K, A>
+export function fromFoldable<K, F>(
+  S: Setoid<K>,
+  // tslint:disable-next-line: deprecation
+  F: Foldable<F>
+): <A>(ta: HKT<F, [K, A]>, f: (existing: A, a: A) => A) => Map<K, A>
+export function fromFoldable<K, F>(
+  S: Setoid<K>,
+  // tslint:disable-next-line: deprecation
```
gcanti: I think that `fromFoldable` should use `Foldable2v`

src/Map.ts
```diff
@@ -0,0 +1,825 @@
...
+  F: Foldable<F>
+): <A>(ta: HKT<F, [K, A]>, f: (existing: A, a: A) => A) => Map<K, A> {
+  return <A>(ta: HKT<F, [K, A]>, f: (existing: A, a: A) => A) => {
+    return F.reduce<[K, A], Map<K, A>>(ta, new Map<K, A>(), (b, [k, a]) => {
+      const bOpt = lookupWithKey(S)(k, b)
```
gcanti: `lookupWithKey(S)` can be moved outside

src/Map.ts
```diff
@@ -0,0 +1,825 @@
...
+      b.set(k, bOpt.isSome() ? f(bOpt.value[1], a) : a)
```
gcanti: should re-use the existing key here

src/Map.ts
```diff
@@ -0,0 +1,825 @@
...
+    _L: phantom,
+    ...functor,
+    mapWithIndex: mapWithKey
+  }
+}
+
+/**
+ * @since 1.14.0
+ */
+export const getFoldable = <K>(O: Ord<K>): Foldable2v2C<URI, K> => {
```
gcanti: `getFoldable`, `getFoldableWithIndex`, `getTraversable`, `getTraversableWithIndex` can be merged into `getTraversableWithIndex`

texastoland:
> Instead, we could create functions that would return those instances when provided with an `Ord` constraint.

This makes sense to me. Were you considering an alternative?

> `member` is named `elem` in `Foldable` (and renamed to `member` in `Foldable2v`)

🤯

> * [`member` in `purescript-ordered-collections` (Set)](https://github.com/purescript/purescript-ordered-collections/blob/master/src/Data/Set.purs#L117) tests the value

Would it be a stretch to consider values to also be keys in `Set`?

> I'm more interested in signature consistency in this case.

👍🏽

> However, I'm not sure that the name is different enough. I think a better name would be something like `includes`, `contains`, or maybe `isValueOf` or something to more clearly convey the difference.

Agree about `isMember`. Maybe `elemKey :: v -> Map k v -> Maybe k` (like PS `Array.elemIndex`)?

joshburgess:
> @joshburgess Looks like the tests are not running with ` --clearCache`
> 
> A general observation: the tests would be more solid if `K` is not a primitive, for example

Yeah, I noticed this. However, it wasn't like this before when I first made the change. It was still showing me the reporting correctly before, but you've made various commits since then. I think something else is going on now?

joshburgess:
> A general observation: the tests would be more solid if `K` is not a primitive, for example

Yeah... I was just being lazy here. 😅 I'll change them.

src/Map.ts
joshburgess: I did this so that that `SK` (the `Setoid<K>`) would test both sets of keys for equality using the Setoid. I think it needs to stay like this?

src/Map.ts
joshburgess: Same as before. I was making sure the `Setoid`'s `equals` was acting on both Map's K keys.

src/Map.ts
joshburgess: Can you explain this? It still does get exported within the instances, just not as a standalone function. Why does this eliminate the need for the overloads?

src/Map.ts
joshburgess: Same as before, can you explain why? Are they only needed when treated as standalone functions & don't exist within a type class instance?

src/Map.ts
joshburgess: Same as before

src/Map.ts
joshburgess: Same as before

src/Map.ts
joshburgess: Same as before

src/Map.ts
joshburgess: Same as before

src/Map.ts
joshburgess: I was just following what was done in `Record` here. I can make this change, but, then, we should also change it in `Record`.

src/Map.ts
joshburgess: Like before, I did this to ensure that the `Setoid<K>` gets used for equality testing.

joshburgess:
> This makes sense to me. Were you considering an alternative?

@texastoland They were all just going to be standalone functions until we went down this route, but I agree and have already made that change.

src/Map.ts
joshburgess: I can't do this, because:

- `getTraversable` is also needed for `getWitherable`
- `getFoldable` is needed by both `getTraversable` and `getFoldableWithIndex`

src/Map.ts
joshburgess: I made this change. Marking it as resolved.

src/Map.ts
joshburgess: I made this change. Marking it as resolved.

src/Map.ts
joshburgess: I made this change. Marking it as resolved.

src/Map.ts
joshburgess: I made this change. Marking it as resolved.

src/Map.ts
joshburgess: I made this change. Marking it as resolved.

src/Map.ts
joshburgess: I made this change. Marking it as resolved.

src/Map.ts
joshburgess: I made this change. Marking it as resolved.

src/Map.ts
joshburgess: I made this change. Marking it as resolved.

src/Map.ts
joshburgess: I changed this. Marking it as resolved.

src/Map.ts
joshburgess: I changed this. Marking it as resolved.

src/Map.ts
joshburgess: I'm not sure about this one. Do you mean like this?

```
return F.reduce<[K, A], Map<K, A>>(ta, new Map<K, A>(), (b, [k, a]) => {
  b.set(k, f(b.get(k)!, a))
  return b
})
```

joshburgess: I'm not sure what you mean here.

src/Map.ts
joshburgess: I made this change. Marking as resolved.

joshburgess:
Not sure why that started happening with `clearCache`, but I moved it into a separate npm command, and now it works again.

```
// gets called by itself first
"jest-clear-cache": "jest --clearCache",
// gets called separately
"jest": "jest --ci --coverage",
```

joshburgess:
@gcanti I was at 100% test coverage before, and now I'm not after changing what gets exported & reorganizing the tests (following what other modules with these `get_` functions that return instances did), but I don't understand why. All of the old tests still exist. What needs to change here?

src/Map.ts
joshburgess: I made this change. Marking as resolved.

src/Map.ts
joshburgess: I made this change. Marking as resolved.

src/Map.ts
joshburgess: I made this change. Marking as resolved.

src/Map.ts
joshburgess: I made this change. Marking as resolved.

src/Map.ts
gcanti: > we should also change it in `Record`

I can't, it would be a breaking change since Foldable2v requires more operations

src/Map.ts
gcanti: The following test is failing with the current implementation

```ts
interface User {
  id: number
}

const setoidUser: Setoid<User> = fromEquals((x, y) => x.id % 2 === y.id % 2)

assert.deepStrictEqual(
  M.fromFoldable(setoidUser, array)([[{ id: 1 }, 1], [{ id: 3 }, 3]], (existing, _) => existing),
  new Map<User, number>([[{ id: 1 }, 1]])
)
```

keys are duplicated

```
Expected value to deeply and strictly equal to:
      Map {{"id": 3} => 1}
Received:
  Map {{"id": 1} => 1, {"id": 3} => 1}
```

I think should be

```diff
+if (bOpt.isSome()) {
+  b.set(bOpt.value[0], f(bOpt.value[1], a))
+} else {
+  b.set(k, a)
+}
-b.set(k, bOpt.isSome() ? f(bOpt.value[1], a) : a)
```

src/Map.ts
gcanti: Yeah, I misspoke, I mean more "let's keep the public API surface as small as possible", we can export only `getTraversableWithIndex`

gcanti:
@joshburgess you can check what's not covered by opening `./coverage/lcov-report/Map.ts.html`

gcanti:
@joshburgess this is the test file I'm using to check the implementation if it may help

https://gist.github.com/gcanti/8ccee8a84d8244e4f8c3cfa8b6180685

- `getMonoid` test is failing
- tests tagged with `.skip` are not modified (yet)

src/Map.ts
joshburgess: I fixed this. Marking as resolved.

src/Map.ts
joshburgess: Ok, I minimized it as much as I could. Things like `getWitherable` and `getFilterableWithIndex` still must be public though.

joshburgess:
@gcanti Okay, I've said this a few times now, but I think this is _finally_ actually ready to be merged. 😛 

I rewrote the tests to use objects as keys in the Maps to better test the `Ord`/`Setoid` functionality.

This uncovered a bug with that `keysSet` function. My previous implementation was correct. You can't do `new Set(myMap.keys())`, because ES6 Sets only use strict equality checking.

```ts
new Set([{ a: 1 }, { b: 2 }, { a: 1 }])

// => Set(3) [[Entries]]: Array(3) [{ a: 1 }, { b: 2 }, { a: 1 }]
```

So, it must use `fromArray` from the `Set` module.

Anyway, there is now 100% test coverage, and the tests are better than they were before.

I think the only thing left to do before merging is to decide on whether or not `has` and `isMember` should be renamed. Personally, I like `member` and `elem`.

texastoland:
@joshburgess did you see this? It seems more helpful than returning a boolean.

> Agree about `isMember`. Maybe `elemKey :: v -> Map k v -> Maybe k` (like PS `Array.elemIndex`)?

gcanti:
> You can't do `new Set(myMap.keys())`

@joshburgess it depends, whether the input `m` is assumed "valid" with respect to a particular `Setoid<K>` (e.g built with `insert` or `fromFoldable`) or not.

However the current signature is safer so :+1:

I would avoid a dependency on `Set` for just `fromArray` though (if possible), we are already paying a dependency on `Array`, maybe we could use `uniq`

```ts
export const keysSet = <K>(S: Setoid<K>): (<A>(m: Map<K, A>) => Set<K>) => {
  const uniqS = uniq(S)
  return m => new Set(uniqS(Array.from(m.keys())))
}
```

joshburgess:
@gcanti Have you considered a bool tool like `rollup` to get treeshaking capabilities? Non-referenced functions could get stripped away.

Alternatively, I could just write my own `fromArray`/`uniq` in this file to avoid the dependency if you want?

joshburgess:
> @joshburgess did you see this? It seems more helpful than returning a boolean.
> 
> > Agree about `isMember`. Maybe `elemKey :: v -> Map k v -> Maybe k` (like PS `Array.elemIndex`)?

So, it looks for a value, and, if found, it returns a Some containing the key or None? That seems interesting & useful, but it would require changing all of the other current `isMember` functions as well. I think it would be best to add it as an additional function rather than replace `isMember` with it. We could always follow up this PR with additional functions at a later time, since additions wouldn't be breaking changes.

gcanti:
> I could just write my own `sort` & `fromArray`/`uniq` in this file to avoid the dependency

@joshburgess good idea, we can get rid of `sort` pretty easily

```ts
export const keys = <K>(O: Ord<K>): (<A>(m: Map<K, A>) => Array<K>) => {
  return m => Array.from(m.keys()).sort(O.compare)
}
```

getting rid of `uniq` is more annoying but it would be worth it

> Have you considered a build tool like `rollup` to get treeshaking capabilities? Non-referenced functions could get stripped away

No, how would that work? Let's talk about this in a separate issue / PR though.

joshburgess:
> @joshburgess good idea, we can get rid of `sort` pretty easily

> getting rid of `uniq` is more annoying but it would be worth it

I made both of these changes. There are no longer any dependencies from Array or Set included in this Map module.

> No, how would that work? Let's talk about this in a separate issue / PR though.

Actually, I think I may have misspoke. I'll think about this a bit and get back to you. I'll open an issue and/or PR if I think there's a way.

joshburgess:
@gcanti @texastoland List of new changes other than the above elimination of dependencies:

- For the sake of making some progress on undecided things, I renamed `has` back to `member` to match the PureScript/Haskell terminology and `isMember` to `elem` to match the idea of `elem` for foldables. I am open to reverting/changing these, but these would be my preferences. I am also open to simply dropping `isMember`/`elem` since it is now derivable by getting ahold of the foldable instance via `getTraversableWithIndex`.
- Added `elemWithKey`. This is essentially the same thing as `lookupWithKey`, but performs the lookup via the value rather than the key.
- Added `elemKey`, based on @texastoland 's suggestion. This uses `elemWithKey` internally, but maps over its returned Option to turn the pair into just a key, similar to how how `lookup` uses `lookupWithKey` internally
- Refactored `isMember` (now called `elem`) to use `elemWithKey` internally + the `.isSome` boolean check, similar to what `has` (now called `member`) does with `lookup`

NOTE: The only odd thing about `elemWithKey` and `elemKey` is that they return only the first occurrence of an equal value. This is normal with lists, but is weird when working with maps? I think it's okay, but just figured it was worth pointing out.

gcanti:
> I renamed `has` back to `member` to match the PureScript/Haskell terminology and `isMember` to `elem` to match the idea of `elem` for foldables

:+1:

@joshburgess to recap

```
member =        <K>(S: Setoid<K>): (<A>(k: K, m: Map<K, A>) => boolean)
lookup =        <K>(S: Setoid<K>): (<A>(k: K, m: Map<K, A>) => Option<A>)
lookupWithKey = <K>(S: Setoid<K>): (<A>(k: K, m: Map<K, A>) => Option<[K, A]>)

elem =        <A>(S: Setoid<A>): (<K>(a: A, m: Map<K, A>) => boolean)
elemKey =     <A>(S: Setoid<A>): (<K>(a: A, m: Map<K, A>) => Option<K>)
elemWithKey = <A>(S: Setoid<A>): (<K>(a: A, m: Map<K, A>) => Option<[K, A]>)
```

shouldn't be?

```
elem =               <A>(S: Setoid<A>): (<K>(a: A, m: Map<K, A>) => boolean)
lookupKey (*) =      <A>(S: Setoid<A>): (<K>(a: A, m: Map<K, A>) => Option<K>)
lookupKeyWithValue = <A>(S: Setoid<A>): (<K>(a: A, m: Map<K, A>) => Option<[K, A]>)
```

(*) `lookupKey` is a temporary name

> they act on only the first occurrence of an equal value

It's not a problem if the map is "valid" with respect to the passed `Setoid`.

I think that we should establish once for all whether this module manages only "valid" maps (<= I strongly vote for this) or not.

`keysSet` and `valuesSet` can be simplified a lot if we assume the former. Also, if we assume the latter, many signatures, like `lookup`, don't make much sense.

joshburgess:
Also, just added a new `toSet` function, similar to `keysSet` & `valuesSet`, but returning a `Set<K, A>`.

joshburgess:
> @joshburgess to recap
> 
> ```
> member =        <K>(S: Setoid<K>): (<A>(k: K, m: Map<K, A>) => boolean)
> lookup =        <K>(S: Setoid<K>): (<A>(k: K, m: Map<K, A>) => Option<A>)
> lookupWithKey = <K>(S: Setoid<K>): (<A>(k: K, m: Map<K, A>) => Option<[K, A]>)
> 
> elem =        <A>(S: Setoid<A>): (<K>(a: A, m: Map<K, A>) => boolean)
> elemKey =     <A>(S: Setoid<A>): (<K>(a: A, m: Map<K, A>) => Option<K>)
> elemWithKey = <A>(S: Setoid<A>): (<K>(a: A, m: Map<K, A>) => Option<[K, A]>)
> ```
> shouldn't be?
> 
> ```
> elem =               <A>(S: Setoid<A>): (<K>(a: A, m: Map<K, A>) => boolean)
> lookupKey (*) =      <A>(S: Setoid<A>): (<K>(a: A, m: Map<K, A>) => Option<K>)
> lookupKeyWithValue = <A>(S: Setoid<A>): (<K>(a: A, m: Map<K, A>) => Option<[K, A]>)
> ```
> (*) `lookupKey` is a temporary name

Hmm, I'm not sure. I was thinking that the `lookup_` variations would always perform the lookup using the key + a `Setoid<K>`, while the `elem_` variations would ignore the key and "lookup" the `A` value via a `Setoid<A>`.

I think it's okay that some of them would share the same type signature, because the implementation is different (a lookup based on the key vs a "lookup" based on the value).

In other words, I was tying the name `lookup` to the behavior of "looking at" the key and the name `elem` to the behavior of "looking at" the value, not their return types.

> It's not a problem if the map is "valid" with respect to the passed `Setoid`.
> 
> I think that we should establish once for all whether this module manages only "valid" maps (<= I strongly vote for this) or not.
> 
> `keysSet` and `valuesSet` can be simplified a lot if we assume the former. Also, if we assume the latter, many signatures, like `lookup`, don't make much sense.

Can you further elaborate on this? Sorry, I don't fully understand what you mean by "valid" here.

gcanti:
**Naming**

> In other words, I was tying the name `lookup` to the behavior of "looking at" the key and the name `elem` to the behavior of "looking at" the value, not their return types

Ok, I was referring to the name-pattern `X` / `XWithValue`

```ts
X =          <A>(S: Setoid<A>): (<K>(a: A, m: Map<K, A>) => Option<K>)
XWithValue = <A>(S: Setoid<A>): (<K>(a: A, m: Map<K, A>) => Option<[K, A]>)
```

So if we choose `X = elemKey` for the former then we could use `XWithValue = elemKeyWithValue` for the latter

---

**Constraints and general assumptions**

> what you mean by "valid" here

By "valid" I mean there's no duplicated keys. 

Let's consider `lookup`:

Current signature:

```ts
lookup = <K>(S: Setoid<K>): (<A>(k: K, m: Map<K, A>) => Option<A>)
```

Now if we don't assume that there's no duplicate keys in `m` with respect to `Setoid<K>` the signature should be

```ts
lookup = <K>(S: Setoid<K>): (<A>(k: K, m: Map<K, A>) => Array<A>)
```

since it may return `0`, `1` or more values for a single key.

I think that the general assumption we should make for this module is the following

> maps cannot contain duplicated keys

As a consequence `keysSet` can be simplified to `keysSet = <A>(m: Map<K, A>) => Set<K>`.

---

**Drop `valuesSet`**

Thinking twice, I'm not sure this function makes sense: values, as opposed to keys, may be duplicated. Let's drop it.

gcanti:
> The only odd thing about `elemWithKey` and `elemKey` is that they act on only the first occurrence of an equal value

@joshburgess ah that's right.. both `elemKey` and `elemWithKey` don't make much sense for a map. Let's drop them.

joshburgess:
> @joshburgess ah that's right.. both `elemKey` and `elemWithKey` don't make much sense for a map. Let's drop them.
> 
> EDIT: just noticed a `toSet` function. Same as above, let's drop it.

Why do you say that? Because of the fact that the key-value pairs are unordered?

joshburgess:
> **Drop `valuesSet`**
> 
> Thinking twice, I'm not sure this function makes sense: values, as opposed to keys, may be duplicated. Let's drop it.

This means `keysSet` should also be dropped then, doesn't it?

joshburgess:
Ok, I deleted those functions & their tests, but the build is failing now for some reason: `Error: Cannot find module 'definitelytyped-header-parser'`

gcanti:
@joshburgess Thanks!

PR #725, opened 2019-02-10
Record: optimise filterWithIndex (conservative) and traverseWithKey (less object creations) (fixes #711)

### LINKED ISSUE #711: `insert` of `Record` should be a `noop` if the value at the key already exists (strict equals)
Note: We may also optimise `filterWithIndex` and `filterMapWithIndex`

sledorze:


sledorze:
@gcanti I still have the problem of running the tests locally (related to Jest 
 as depicted here https://github.com/gcanti/fp-ts/pull/715
).
So blindly guessing and using the CI to fix issues, sorry for the noise..

gcanti:
> I still have the problem of running the tests locally

Ah! Adding

```json
"jest": {
  "testURL": "http://localhost/"
}
```

in `package.json` does solve the problem?

gcanti:
@sledorze actually we could upgrade to `jest@23.6.0`

src/Record.ts
```diff
@@ -571,15 +580,25 @@ export function filterMapWithIndex<A, B>(
 export function filterWithIndex<K extends string, A>(fa: Record<K, A>, p: (key: K, a: A) => boolean): Record<string, A>
 export function filterWithIndex<A>(fa: Record<string, A>, p: (key: string, a: A) => boolean): Record<string, A>
 export function filterWithIndex<A>(fa: Record<string, A>, p: (key: string, a: A) => boolean): Record<string, A> {
```
gcanti: What about

```ts
export function filterWithIndex<A>(fa: Record<string, A>, p: (key: string, a: A) => boolean): Record<string, A> {
  const r: Record<string, A> = {}
  let changed = false
  for (const key in fa) {
    if (fa.hasOwnProperty(key)) {
      const a = fa[key]
      if (p(key, a)) {
        r[key] = a
      } else {
        changed = true
      }
    }
  }
  return changed ? r : fa
}
```

test/Record.ts
```diff
@@ -147,6 +149,12 @@ describe('Record', () => {
     const y: Record<string, string | number> = { a: 1, b: 'foo' }
     const actual = R.filter(y, isNumber)
     assert.deepStrictEqual(actual, { a: 1 })
+
+    const z: Record<string, string | number> = { b: 1, a: 'foo', c: 'bar' }
```
gcanti: This test doesn't look different from the previous one

sledorze:
Was to trigger another case for coverage

src/Record.ts
sledorze: It's doing all work in case of noop (obj création plus assigns (n log n))

gcanti: noop is unlikely in a `filter*` function so in the most common case looks like you are doing even more work

sledorze: I m fine with the simpler approach.
Afk this morning (on phone) so expect a change later.

sledorze:
@gcanti yes adding

```
"jest": {
  "testURL": "http://localhost/"
}
```
Solves the problem (I have another one popping, however)

gcanti:
FYI we are upgrading jest in the other PR https://github.com/gcanti/fp-ts/pull/715#issuecomment-462247271

sledorze:
@gcanti not sure how I can fix that coverage issue

gcanti:
@sledorze did you try `open coverage/lcov-report/index.html`?

src/Record.ts
```diff
@@ -572,14 +581,18 @@ export function filterWithIndex<K extends string, A>(fa: Record<K, A>, p: (key:
...
   const r: Record<string, A> = {}
-  const keys = Object.keys(fa)
-  for (const key of keys) {
-    const a = fa[key]
-    if (p(key, a)) {
-      r[key] = a
+  let changed = false
+  for (const key in fa) {
+    if (fa.hasOwnProperty(key)) {
```
gcanti: ah sorry, do you mean this line right?

gcanti:
this seems to work

```ts
const x = Object.assign(Object.create({ c: true }), { a: 1, b: 'foo' })
assert.deepStrictEqual(R.filter(x, isNumber), { a: 1 })
```

src/Record.ts
sledorze: yes

sledorze:
thanks @gcanti

PR #726, opened 2019-02-11
Handle unwanted behaviour when an IO throws an error.

bwlt:
This will prevent a `Task` created with `fromIO` to throw instead of return a Promise with a rejection.


gcanti:
@bwlt thanks for this PR but an `IO`, like any other code, shouldn't throw in the first place

PR #738, opened 2019-02-15
fix Wilt* definitions

gcanti:


giogonzo:
@gcanti just for knowledge sharing, what's the fix here exactly?

gcanti:
Repro

```ts
import { array } from '../src/Array'
import { Either, left, right } from '../src/Either'
import * as I from '../src/Identity'

const wiltIdentity = array.wilt(I.identity)
const p = (n: number) => n > 2
const f = (n: number): I.Identity<Either<number, number>> => new I.Identity(p(n) ? right(n + 1) : left(n - 1))
const x = wiltIdentity([1, 3], f)
```

Now

```ts
const x: I.Identity<Separated<HKT<"Array", number>, number[]>>
```

After this change

```ts
const x: I.Identity<Separated<number[], number[]>>
```

PR #749, opened 2019-02-26
Fix link to new API Reference docs

joshburgess:
Fixes the API Reference link to point to the new docs @ https://gcanti.github.io/fp-ts


gcanti:
Thanks @joshburgess

PR #759, opened 2019-03-03
Add Ruby dependencies for local docs development

grossbart:
It's unfortunate to have to require Ruby as a dependency in this project, but in order to work with the documentation and extend it for GitHub Pages this makes it much easier to contribute because changes can be previewed locally.

To sort of keep the Ruby dependency contained, I added this to the /docs folder, but there would be nothing stopping us from adding this to the root, which would be a more usual place to look for the Gemfile and Gemfile.lock files.

Note that this PR is not essential to accept into fp-ts and I would completely understand if it wasn't accepted – it's just foreign tech in a way –, but because it simplifies contributions to the documentation it might be worthwile to look into.

At the moment I don't see a strong focus on Jekyll template customization yet, but at one point this could also become a topic so that we can customize the currently purple color of Just-the-docs to a more TypeScript-y blue and create a more distinctive look. First things first, though with more content, I would say.


gcanti:
Thanks @grossbart but I prefer to keep it simple, the good thing of a markdown documentation is that makes contributions easier, without any special setup.

> because changes can be previewed locally

You could setup `gh-pages` in your `fp-ts` fork:

Settings -> GitHub Pages -> choose "master branch /docs folder"

grossbart:
Yeah, I agree that it's too much ceremony! Some things I like to experiment with locally, but for that I have the tools locally, for demo-ing I setup GitHub Pages on my branch, but it's a bit of a pity that it only works on the master branch – that is not ideal for showing unreleased stuff. But oh well, there are ways around that, too 🤷‍♂️

PR #760, opened 2019-03-03
Show how Nix could be used for using the Rub env

grossbart:
It's clear to me that [Nix](https://nixos.org/nix/) is an exotic choice and mostly a personal preference of mine (I haven't used Homebrew for years thanks to Nix). This is not for everyone; however, this example shows how Nix could be used to easily create a reproducible Ruby environment, which can be used for writing documentation with Jekyll and GitHub Pages.

I've been successfully using this setup for several years now, so I wanted to suggest it as a possibility. But I do see this as maybe a bit too exotic for this repository and am torn whether it should be accepted or not – probably not, so feel free to close. The main issue would be that this creates more confusion ("do I need Nix to contribute?"), which could be counter productive.

I'm happy to answer questions on this approach, however, if you have any :)


grossbart:
Closing this because the focus is to not require any special tools but edit Markdown directly.

PR #762, opened 2019-03-08
Adjust the documentation colors

Wraul:
The default purple color used by the documentation theme bothers me a bit as fp-ts always had more of a blue feel to me.
The purple combined with the blue fp-ts logo bothered me enough to learn a bit about Jekyll and to create this PR.

I have adjusted the link and button colors using the fp-ts logo as a guide. To me this gives the documentation a vastly more coherent look.

There is one small problem though. The current version of the theme is broken with regards to customisation. This forced me into switching to a forked version of the theme. 

There exists a PR attempting to solve the issue pmarsceill/just-the-docs#97
So I'm thinking that this PR, if accepted, should probably not be merged until the above is fixed and we can use the official theme again.

Anyway I though I'll leave this PR here in hope that it might somehow be useful.

Ohh, and thank you for the awesome library. It makes TypeScript so much more bearable 👍

### DIFF
--- a/docs/_config.yml
+++ b/docs/_config.yml
@@ -1,4 +1,4 @@
-remote_theme: pmarsceill/just-the-docs
+remote_theme: fluffyemily/just-the-docs
 
 # Enable or disable the site search
 search_enabled: true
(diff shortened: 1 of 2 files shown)


grossbart:
Ah, this is nice! I almost thought that it is not possible to override theme vars unless you add the whole theme to the repo, which would have been too much – glad I was wrong!

PR #765, opened 2019-03-10
Add overloads to sequenceT to allow more arguments (up to 7)

cdimitroulas:
This PR adds a couple of overloads to allow passing an increased number of arguments to `sequenceT` (up to 7 instead of just 5).

I had some problem running `npm test` in the repo (the `npm run prettier` command fails for some reason which I couldn't figure out) but the tests all pass.

### DIFF
--- a/src/Apply.ts
+++ b/src/Apply.ts
@@ -231,173 +231,173 @@ export function getSemigroup<F, A>(F: Apply<F>, S: Semigroup<A>): () => Semigrou
   })
 }
 
-export interface SequenceT3<H extends URIS3> {
-  <U, L, A>(a: Type3<H, U, L, A>): Type3<H, U, L, [A]>
-  <U, L, A, B>(a: Type3<H, U, L, A>, b: Type3<H, U, L, B>): Type3<H, U, L, [A, B]>
-  <U, L, A, B, C>(a: Type3<H, U, L, A>, b: Type3<H, U, L, B>, c: Type3<H, U, L, C>): Type3<H, U, L, [A, B, C]>
-  <U, L, A, B, C, D>(a: Type3<H, U, L, A>, b: Type3<H, U, L, B>, c: Type3<H, U, L, C>, d: Type3<H, U, L, D>): Type3<
-    H,
+export interface SequenceT3<F extends URIS3> {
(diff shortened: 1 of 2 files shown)


raveclassic:
Hi, @cdimitroulas, thanks for the PR!

Is there any reason for changing the name of `F` type argument to `H`? `F` is a common naming convention for a HKT in a scope.

gcanti:
@cdimitroulas you can run `npm run fix-prettier`

cdimitroulas:
@gcanti ah thanks, I missed that script. I've ran `fix-prettier` now and everything seems to be ok.

@raveclassic actually I wasn't aware of this convention - I just assumed `F` had been used as it was the next letter after `E` which was the last letter used for the arguments. So when I wanted to add two new arguments I wanted to name them `F` and `G` so I renamed the `F` type argument to `H` to follow the same logic.

Would you prefer if I rename it back to `F` and then use `G` and `H` for the new overload type arguments?

raveclassic:
@cdimitroulas Yeah, usually `F` is skipped in such cases

cdimitroulas:
OK @raveclassic I've made that change + I added an extra overload for 8 arguments to `sequenceT` for good measure :)

src/Apply.ts
```diff
@@ -266,6 +293,33 @@ export interface SequenceT3C<F extends URIS3, U, L> {
...
+  ): Type3<F, U, L, [A, B, C, D, E, G, H]>
+  <A, B, C, D, E, G, H, I>(
+    a: Type3<F, U, L, A>,
+    b: Type3<F, U, L, B>,
+    c: Type3<F, U, L, C>,
+    d: Type3<F, U, L, D>,
+    e: Type3<F, U, L, E>,
+    g: Type3<F, U, L, G>,
+    h: Type3<F, U, L, H>,
+    i: Type3<F, U, L, H>
```
gcanti: typo

cdimitroulas: good catch, thanks

raveclassic:
@cdimitroulas Nice, thanks!

gcanti:
Thanks @cdimitroulas

Issue #520, opened 2018-07-06
Suggestion: Make instances of existing ADTs extendable


raveclassic:
Currently instance of typeclass for an ADT is defined as constant in the module of such ADT. This makes it impossible to extend existing ADT instance with additional typeclasses because TS doesn't allow merging of interfaces for constants in module augmentation.

I suggest to move typeclass methods to class body and to still export constant which would be an instance of that class. Here's an example:

```typescript
// Array.ts
export class ArrayInstance
  implements
    Monad1<URI>,
    Foldable1<URI>,
    Unfoldable1<URI>,
    Traversable1<URI>,
    Alternative1<URI>,
    Plus1<URI>,
    Extend1<URI>,
    Compactable1<URI>,
    Filterable1<URI>,
    Witherable1<URI> {
  readonly URI = URI

  readonly map = map
  readonly compact = compact
  readonly separate = separate
  readonly filter = filter
  readonly filterMap = filterMap
  readonly partition = partition
  readonly partitionMap = partitionMap
  readonly of = of
  readonly ap = ap
  readonly chain = chain
  readonly reduce = reduce
  readonly unfoldr = unfoldr
  readonly traverse = traverse
  readonly zero = zero
  readonly alt = alt
  readonly extend = extend
  readonly wither = wither
  readonly wilt = wilt
}

export const array = new ArrayInstance() //non-breaking

// Awesome.ts
import { URI, ArrayInstance } from './Array'
import { HKT, URIS, Type } from './HKT'

// typeclass
export interface Awesome<F> {
  readonly URI: F
  readonly makeItAwesome: <A>(fa: HKT<F, A>) => void
}
export interface Awesome1<F extends URIS> {
  readonly URI: F
  readonly makeItAwesome: <A>(fa: Type<F, A>) => void
}

// generic version
export function makeItAwesome<F extends URIS>(F: Awesome1<F>): <A>(fa: Type<F, A>) => void
export function makeItAwesome<F>(F: Awesome<F>): <A>(fa: HKT<F, A>) => void {
  return fa => F.makeItAwesome(fa)
}

// instance for Array
const awesomeArray: Awesome1<URI> = {
  URI,
  makeItAwesome: <A>(as: Array<A>) => 'Yeah!'
}
declare module './Array' {
  interface ArrayInstance extends Awesome1<URI> {}
}
Object.assign(ArrayInstance.prototype, awesomeArray)

// test.ts
import { makeItAwesome } from './Awesome'
[...]

PR #772, opened 2019-03-12
Add Align and Semialign type classes

Wraul:
I thought that we should get some more milage out of the These data type. So I decided to have a go at implementing the `Align` and `Semialign` type classes, that's part of the [_these_ Haskell package](http://hackage.haskell.org/package/these-0.8/docs/Data-Align.html).

I'm not sure if this is something that you want included in fp-ts.  
But if you do, I would appreciate some assistance regarding a couple of things that I'm a bit uncertain about:
1. The implementation of `Align`. It only defines the constant `nil` and I'm not sure how that should interact with the HKT emulation.
2. In the documentation, have I translated the laws correctly, and do they have names?

Also feel free to point out any other misstakes and inefficiencies. I'm partly doing this to learn, so that would be helpful.

### DIFF
--- a/dtslint/ts3.0/index.ts
+++ b/dtslint/ts3.0/index.ts
@@ -234,6 +234,15 @@ declare const fromFoldableF1: Fo.Foldable<'Test'>
 declare const fromFoldableInput1: H.HKT<'Test', ['a' | 'b', number]>
 const fromFoldable1 = R.fromFoldable(fromFoldableF1)(fromFoldableInput1, a => a) // $ExpectType Record<"a" | "b", number>
 
+declare const d2: Record<string, string>
+declare const r2: Record<'a', number>
+declare const r3: Record<'b', string>
+R.align(d1, d2) // $ExpectType Record<string, These<number, string>>
+R.align(r2, r3) // $ExpectType Record<"a" | "b", These<number, string>>
+
+R.alignWith(d1, d2, (x: Th.These<number, string>) => 'Test') // $ExpectType Record<string, string>
+R.alignWith(r2, r3, (x: Th.These<number, string>) => 'Test') // $ExpectType Record<"a" | "b", string>
+
 //
 // Semigroup
 //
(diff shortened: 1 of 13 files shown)


gcanti:
@Wraul Nice! thanks for this PR

> I'm not sure if this is something that you want included in fp-ts

I think that for this kind of gerachy an external library would be a better choice (mainly to avoid a runtime dependency on `These` in core modules like `Array`, `Option`, etc...). Maybe [fp-ts-contrib](https://github.com/gcanti/fp-ts-contrib)?

> The implementation of Align. It only defines the constant nil and I'm not sure how that should interact with the HKT emulation

What about this? (prior art: [Category's id](https://github.com/gcanti/fp-ts/blob/2d0d1e232adeea46d4741ddeb97897f04a600f6d/src/Category.ts#L8))

```ts
export interface Align<F> extends Semialign<F> {
  readonly nil: <A>() => HKT<F, A>
}

export interface Align1<F extends URIS> extends Semialign1<F> {
  readonly nil: <A>() => Type<F, A>
}

export interface Align2<F extends URIS2> extends Semialign2<F> {
  readonly nil: <L, A>() => Type2<F, L, A>
}

// etc...
```

> In the documentation, have I translated the laws correctly?

AFAIK yes

> and do they have names?

not sure

Wraul:
Thank you for the feedback, greatly appreciated.

I have changed `Align` to what you proposed, and it appears to be working great 👍 

> I think that for this kind of gerachy an external library would be a better choice (mainly to avoid a runtime dependency on These in core modules like Array, Option, etc...).

I had a feeling that this was the case. I though about the fact that this adds `These` as an additional runtime dependency while I was implementing it. And I felt like this might not be worth it for what  amounts to some niche functionality.
So I agree with your assessment.

> Maybe fp-ts-contrib?

I'm wiling to give this a go.
Though I think I will have to figure out how to externally add type class instances to existing types.
I'm looking at what you wrote [here](https://github.com/gcanti/fp-ts/issues/520#issuecomment-402986198), and thinking that this might be the approach. But I'll have to try it out to better grasp the details.

I'm closing this PR for now.

gcanti:
> I'm wiling to give this a go

Great, looking forward to your PR

PR #773, opened 2019-03-13
add Selective typeclass

giogonzo:
I'm doing this mainly for learning purposes (encoding of type classes and instances in fp-ts),
not sure if this is something that could be part of `fp-ts` core or not

It's a porting of https://github.com/snowleopard/selective

disclaimer: super WIP! If there's interest and what I started somewhat makes sense, I'll invest more time in polish and tests

missing:
- [ ] docs
- [ ] `Over` and `Under`
- [ ] more `selectM` instances

### DIFF
--- a/src/Option.ts
+++ b/src/Option.ts
@@ -98,6 +98,7 @@ import { Semigroup } from './Semigroup'
 import { Setoid, fromEquals } from './Setoid'
 import { Traversable2v1 } from './Traversable2v'
 import { Witherable1 } from './Witherable'
+import { Selective1 } from './Selective'
 
 declare module './HKT' {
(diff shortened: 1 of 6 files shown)


gcanti:
@giogonzo thanks for this PR but selective functors, while interesting, AFAIK are still experimental. Would you consider to send a PR to [fp-ts-contrib](https://github.com/gcanti/fp-ts-contrib) instead?

giogonzo:
@gcanti sure, I'll move the PR to the other repo later today

giogonzo:
moved to https://github.com/gcanti/fp-ts-contrib/pull/6

PR #776, opened 2019-03-14
Add flatTraverse operation

julienrf:


julienrf:
I’m not sure what exactly is the problem. I have the following output:

~~~
> prettier --list-different "./{src,test,examples,tutorial}/**/*.ts"

src/Traversable2v.ts
test/Traversable2v.ts
npm ERR! code ELIFECYCLE
npm ERR! errno 1
npm ERR! fp-ts@1.14.4 prettier: `prettier --list-different "./{src,test,examples,tutorial}/**/*.ts"`
npm ERR! Exit status 1
~~~

src/Traversable2v.ts
```diff
@@ -297,3 +298,16 @@ export function getTraversableComposition<F, G>(
...
+ * A traverse followed by flattening the inner result.
+ *
+ * @since 1.14.5
+ */
+export function flatTraverse<F extends URIS3, G extends URIS3>(F: Applicative3<F>, G: Traversable2v3<G> & Chain3<G>): <U, L, A, B>(ta: Type3<G, U, L, A>, f: (a: A) => Type3<F, U, L, Type3<G, U, L, B>>) => Type3<F, U, L, Type3<G, U, L, B>>
+export function flatTraverse<F extends URIS2, G extends URIS2>(F: Applicative2<F>, G: Traversable2v2<G> & Chain2<G>): <L, A, B>(ta: Type2<G, L, A>, f: (a: A) => Type2<F, L, Type2<G, L, B>>) => Type2<F, L, Type2<G, L, B>>
+export function flatTraverse<F extends URIS, G extends URIS>(F: Applicative1<F>, G: Traversable2v1<G> & Chain1<G>): <A, B>(ta: Type<G, A>, f: (a: A) => Type<F, Type<G, B>>) => Type<F, Type<G, B>>
+export function flatTraverse<F, G>(F: Applicative<F>, G: Traversable2v<G> & Chain<G>): <A, B>(ta: HKT<G, A>, f: (a: A) => HKT<F, HKT<G, B>>) => HKT<F, HKT<G, B>>
+export function flatTraverse<F, G>(F: Applicative<F>, G: Traversable2v<G> & Chain<G>): <A, B>(ta: HKT<G, A>, f: (a: A) => HKT<F, HKT<G, B>>) => HKT<F, HKT<G, B>> {
+  return (ta, f) => F.map(G.traverse(F)(ta, f), flatten(G))
```
gcanti: I propose to remove the runtime dependency on the `Chain` module by replacing `flatten` with its expanded counterpart. Also `G.traverse(F)` can be moved outside the returned function as an optimization

```ts
  const traverseF = G.traverse(F)
  return (ta, f) => F.map(traverseF(ta, f), ggb => G.chain(ggb, gb => gb))
```

gcanti:
@julienrf you can run `npm i fix-prettier` to fix the CI error

src/Traversable2v.ts
julienrf: We still need to depend on the `Chain` module to be able to call `G.chain`, no? Or maybe by using the `flatten` function some module optimizations become impossible?

gcanti: Yes but in my version we are only importing types, so there's no runtime dependency. You can double check by inspecting the `lib/Traversable2v.js`file (`var Chain_1 = require("./Chain");` isn't there anymore)

julienrf:
Thanks for the review @gcanti. I want to add one more test case, turning an `Either<string, Option<Either<string, number>>>` into an `Option<Either<string, number>>`. The problem is that calling `flatTraverse(option, either)` doesn’t compile because the kinds of `Either` and `Option` are not aligned. What should I do?

gcanti:
@julienrf Alas when combining two type classes such `Applicative` and `Traversable` there's a combinatorial explosion of overloads.

5 `Traversable`s variants x 5 `Applicative`s variante = 25 overloads

Actually, thinking twice, I don't think it worths it, thanks for ths PR though.

Issue #766, opened 2019-03-11
sequenceT and mapped tuples


gcanti:
@sledorze I'm trying to refactor `sequenceT` with mapped tuples, but I get the same "defect" of [`sequenceS`](https://github.com/gcanti/fp-ts-contrib/blob/9215b6fcd04212afe2b220562cd6e688b069118b/src/sequenceS.ts), i.e. TypeScript infers a union for the left part

```ts
import { Apply2 } from 'fp-ts/lib/Apply'
import { Either, either } from 'fp-ts/lib/Either'
import { Type2, URIS2 } from 'fp-ts/lib/HKT'

export declare function sequenceT<F extends URIS2>(
  F: Apply2<F>
): <T extends Array<Type2<F, any, any>>>(
  ...t: T
) => Type2<
  F,
  T[number] extends Type2<F, infer L, any> ? L : never,
  { [K in keyof T]: T[K] extends Type2<F, any, infer A> ? A : never }
>

declare const fa: Either<string, number>
declare const fb: Either<string, boolean>
declare const fc: Either<string, string>

const f = sequenceT(either)
// x: Either<string, [number, boolean, string]>
const x = f(fa, fb, fc)

declare const fd: Either<boolean, string>

// y: Either<string | boolean, [number, boolean, string, string]>
const y = f(fa, fb, fc, fd)
```

Any idea how to fix this?

Or would it be acceptable?

EDIT: forgot to say that the naive version doesn't work either

```ts
import { URIS2, Type2 } from 'fp-ts/lib/HKT'
import { Apply2 } from 'fp-ts/lib/Apply'
import { Either, either } from 'fp-ts/lib/Either'

export declare function sequenceT<F extends URIS2>(
  F: Apply2<F>
): <L, T extends Array<Type2<F, L, any>>>(
  ...t: T
) => Type2<F, L, { [K in keyof T]: T[K] extends Type2<F, L, infer A> ? A : never }>

declare const fa: Either<string, number>
declare const fb: Either<string, boolean>
declare const fc: Either<string, string>

const f = sequenceT(either)
// x: Either<{}, [number, boolean, string]>
const x = f(fa, fb, fc)
```

`L` is inferred as `{}`

mattiamanzati:
This seems like an hack, but may work. I don't know much about the sequenceT unfortunately :/

```typescript
import { URIS2, Type2 } from 'fp-ts/lib/HKT'
import { Apply2 } from 'fp-ts/lib/Apply'
[...]

PR #781, opened 2019-03-18
New Feature: make Apply.sequenceT variadic using mapped types, closes…

gcanti:
… #766

@mattiamanzati thanks a lot for the trick you suggested, deleting all those overloads feels so good


mattiamanzati:
Thanks for your library @gcanti ! :D

PR #783, opened 2019-03-23
Add new Function alias

ta2gch:
This is a feature request to add new alias for Function.

`Function<A,B>` accepts functions has arguments which is any length. `A` is a tuple of arguments type such as `[number, number]` and `B` is the return type such as `number`.

```typescript
const add: Function<[number, number], number> = (a, b) => a+b;
``` 

Futhermore, `Function` which has no type arguments is equal to `(...args: any[])=>any`.

### DIFF
--- a/docs/modules/function.ts.md
+++ b/docs/modules/function.ts.md
@@ -19,6 +19,7 @@ parent: Modules
 - [Curried8 (type alias)](#curried8-type-alias)
 - [Curried9 (type alias)](#curried9-type-alias)
 - [Endomorphism (type alias)](#endomorphism-type-alias)
+- [Function (type alias)](#function-type-alias)
 - [Function1 (type alias)](#function1-type-alias)
 - [Function2 (type alias)](#function2-type-alias)
 - [Function3 (type alias)](#function3-type-alias)
@@ -152,6 +153,14 @@ export type Curried9<A, B, C, D, E, F, G, H, I, J> = (
 
 ```ts
 export type Endomorphism<A> = (a: A) => A
+```
+
+# Function (type alias)
+
+**Signature**
+
+```ts
+export type Function<A extends Array<any> = Array<any>, B = any> = (...args: A) => B
 ```
 
 # Function1 (type alias)
--- a/docs/modules/function.ts.md
+++ b/docs/modules/function.ts.md
@@ -19,7 +19,6 @@ parent: Modules
 - [Curried8 (type alias)](#curried8-type-alias)
 - [Curried9 (type alias)](#curried9-type-alias)
 - [Endomorphism (type alias)](#endomorphism-type-alias)
-- [Function (type alias)](#function-type-alias)
 - [Function1 (type alias)](#function1-type-alias)
 - [Function2 (type alias)](#function2-type-alias)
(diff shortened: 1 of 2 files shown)


gcanti:
@ta2gch thanks for this PR

> Function which has no type arguments is equal to (...args: any[])=>any

What's the use case? (I'm not comfortable with such a broad signature..)

ta2gch:
It makes compatible with original Function interface of TypeScript.
According to the `lib.es5.d.ts`,
```typescript
interface Function {
    /**
      * Calls the function, substituting the specified object for the this value of the function, and the specified array for the arguments of the function.
      * @param thisArg The object to be used as the this object.
      * @param argArray A set of arguments to be passed to the function.
      */
    apply(this: Function, thisArg: any, argArray?: any): any;


    /**
      * Calls a method of an object, substituting another object for the current object.
      * @param thisArg The object to be used as the current object.
      * @param argArray A list of arguments to be passed to the method.
      */
    call(this: Function, thisArg: any, ...argArray: any[]): any;


    /**
      * For a given function, creates a bound function that has the same body as the original function.
      * The this object of the bound function is associated with the specified object, and has the specified initial parameters.
      * @param thisArg An object to which the this keyword can refer inside the new function.
      * @param argArray A list of arguments to be passed to the new function.
      */
    bind(this: Function, thisArg: any, ...argArray: any[]): any;


    /** Returns a string representation of a function. */
    toString(): string;


    prototype: any;
    readonly length: number;


    // Non-standard extensions
    arguments: any;
    caller: Function;
}
```
https://github.com/Microsoft/TypeScript/blob/3f3444be80dc939eb20db52fa98a206dba31e077/lib/lib.es5.d.ts#L272-L304

For instance, `tsc` can compile the both codes

```typescript
// no imports
const f: Function = (a: number) => a+1;
```

```typescript
import  { Function } from 'fp-ts/lib/function'
const f: Function = (a: number) => a+1;
```

gcanti:
I'd prefer to not provide such an unsafe typing by default and make the arguments required

ta2gch:
I understand what you want. Do you say that we should use `Function<[any], any>` instead of `Function`, don't you? By the way, We cannot get `ReturnType<Function>` if we use builtin's `Function` interface, on the other hand, we could get `ReturnType<Function>` if we use fp-ts's.
So the `Function` improves TypeScript as **Typed** Function, I want tell you this `Function` alias is not **unsafe** and makes builtin types be **safe**.

However, I can update my request if you want so again. Please consider it.

gcanti:
> Do you say that we should use Function<[any], any>

No, I mean that `any` shouldn't be used *at all*

```ts
const f: Function<[any], any> = (a: number) => a + 1
declare function g(f: (a: string) => string): void
g(f) // no error
```

so providing a fallback to `any` is not advised

ta2gch:
I see that the alias `Function` lost some information about types because of the type `any` in your example. Since I agree you said, I'll remove the default value in the definition, AND I'm going to change the name of this type to `FunctionN` because `Function` is preserved at `lib.es5.d.ts` and it has no compatibility with this request. Would do mind it?

gcanti:
@ta2gch thanks, I agree on both changes

ta2gch:
I updated it.

PR #784, opened 2019-03-23
Add Promise.all like interface for Task

ta2gch:
I added `all` to `Task` and `TaskEither` same as `Promise.all`.
i.e., `all` convert `Array<Task<A>>` to `Task<Array<A>>`
```typescript 
const tasks = [new Task(async () => 1), new Task(async () => 1)]
all(tasks).run().then(console.log)
 ```

### DIFF
--- a/docs/modules/Task.ts.md
+++ b/docs/modules/Task.ts.md
@@ -26,6 +26,7 @@ If you want to represent an asynchronous computation that may fail, please see `
 - [URI (constant)](#uri-constant)
 - [task (constant)](#task-constant)
 - [taskSeq (constant)](#taskseq-constant)
+- [all (function)](#all-function)
 - [delay (function)](#delay-function)
 - [fromIO (function)](#fromio-function)
 - [getMonoid (function)](#getmonoid-function)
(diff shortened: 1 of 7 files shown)

### CODE REVIEW

**Comment by gcanti** [comment]
@ta2gch thanks for this PR but this is just `sequence` (an operation of the `Traversable` type class) where the `Traversable` instance is `array` and the `Applicative` instance is `task` / `taskEither`

```ts
import { task } from 'fp-ts/lib/Task'
import { array } from 'fp-ts/lib/Array'

const tasks = [task.of(1), task.of(1)]
// tslint:disable-next-line: no-floating-promises
array
  .sequence(task)(tasks)
  .run()
  .then(console.log) // [ 1, 1 ]
```

ta2gch:
Wow, thank you for telling that awesome way.
I agree that it is natural. I'll close this PR.

grossbart:
@ta2gch I added a recipe that shows how you can solve the problem you had here: https://github.com/gcanti/fp-ts/pull/787 If you have time, please have a look to see whether you think this helps clarify how to do what you wanted. Let me know if you need more examples.

ta2gch:
Thank you for updating that document. It is pretty good helps to understand how to solve that kinds of issue.

PR #785, opened 2019-03-23
Perf: add monad stack comparison benchmark

gcanti:
@sledorze I was trying to get an idea of how much we are loosing by stacking monads and the result is surprising (at least for `TaskEither`), is the benchmark code flawed?

TaskEither x 1,507,865 ops/sec ±5.95% (57 runs sampled)
native x 1,153,964 ops/sec ±7.83% (57 runs sampled)
ReaderTaskEither x 405,216 ops/sec ±9.67% (26 runs sampled)
Fastest is **TaskEither**

### DIFF
--- a/perf/native-vs-task-taskEither-vs-readerTaskEither.ts
+++ b/perf/native-vs-task-taskEither-vs-readerTaskEither.ts
@@ -1,0 +1,52 @@
+import * as Benchmark from 'benchmark'
+
+const suite = new Benchmark.Suite()
+
+import { taskEither } from 'fp-ts/lib/TaskEither'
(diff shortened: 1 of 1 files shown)


gcanti:
Update

```
TaskEither x 1,446,429 ops/sec ±8.57% (59 runs sampled)
native x 1,562,634 ops/sec ±9.28% (52 runs sampled)
ReaderTaskEither x 363,574 ops/sec ±9.01% (55 runs sampled)
```

sledorze:
@gcanti Ok thanks for benchmarking; ReaderTaskEither is very time consuming and I dunno how it can be improved (beside 'Effect rotation').

PR #786, opened 2019-03-24
New recipes: equality, ordering, numbers

grossbart:
I'm still trying to find the right tone, content, and structure, so these might be adjusted in the future. However, they've shown me some things that would need to be covered in a tutorial, which I might start writing next.

Because these are "recipes", I tried to keep the explanations to a minimum: the code should be copy-pasteable and not more at this point. Explanations will follow in different sections of the docs.

Some examples are already well documented in, say, the `Option` documentation page, but for people not knowing where to look those would be unfindable. This is why recipes provide an alternative entry point. Let's see where this goes …

### DIFF
--- a/docs/recipes/equality.md
+++ b/docs/recipes/equality.md
@@ -1,0 +1,141 @@
+---
+title: Equality
+parent: Recipes
+nav_order: 1
+---
+
+# How to determine if two things are equal
+
+With `fp-ts` you can test whether two values are equal using a `Setoid`. You can also compose equality functions to test deep structures and create your own definitions of equality.
+
+We show the most common usages here, but if you need more ways to check for equality, be sure to read the [Setoid](../modules/Setoid.ts) documentation page.
+
+## Primitive equality
+
+```ts
+import { setoidBoolean,  setoidDate, setoidNumber, setoidString } from 'fp-ts/lib/Setoid'
+
+setoidBoolean.equals(true, true) // true
+setoidDate.equals(new Date('1984-01-27'), new Date('1984-01-27')) // true
+setoidNumber.equals(3, 3) // true
+setoidString.equals('Cyndi', 'Cyndi') // true
+```
+
+## Compare structures
+
+```ts
+import { Setoid, getStructSetoid, setoidNumber } from 'fp-ts/lib/Setoid'
+
+type Point = {
+  x: number
+  y: number
+}
+
+const setoidPoint: Setoid<Point> = getStructSetoid({
+  x: setoidNumber,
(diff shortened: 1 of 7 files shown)


docs/recipes/equality.md
```diff
@@ -0,0 +1,141 @@
...
+setoidUserId.equals({ userId: 1, name: 'Giulio' }, { userId: 1, name: 'Giulio Canti' }) // true
+setoidUserId.equals({ userId: 1, name: 'Giulio' }, { userId: 2, name: 'Giulio' }) // false
+```
+
+## More Setoid instances
+
+Many data types provide `Setoid` instances. Here's [Option](../modules/Option.ts):
+
+```ts
+import { getSetoid, none, some } from '../lib/Option'
```
gcanti: typo: `../lib` -> `fp-ts/lib`

docs/recipes/equality.md
```diff
@@ -0,0 +1,141 @@
...
+import { setoidNumber } from '../lib/Setoid'
```
gcanti: idem

docs/recipes/equality.md
```diff
@@ -0,0 +1,141 @@
...
+
+O.equals(some(3), some(3)) // true
+O.equals(none, some(4)) // false
+O.equals(none, none) // true
+```
+
+It works similarly for [Either](../modules/Either.ts) and other types where it is possible to determine equality:
+
+```ts
+import { getSetoid, left, right } from '../lib/Either'
```
gcanti: idem

docs/recipes/equality.md
```diff
@@ -0,0 +1,141 @@
...
+import { setoidNumber, setoidString } from '../lib/Setoid'
```
gcanti: idem

docs/recipes/ordering.md
```diff
@@ -0,0 +1,156 @@
...
+
+const O = getOrd(ordNumber)
+O.compare(none, none) // 0
+O.compare(none, some(1)) // -1
+O.compare(some(1), none) // 1
+O.compare(some(1), some(2)) // -1
+O.compare(some(1), some(1)) // 0
+```
+
+It works similarly for [Tuple](../modules/Tuple.ts)`s and other types where it is possible to determine order:
```
gcanti: It's me or the diff display is weird here, maybe that additional "`"?

docs/recipes/equality.md
grossbart: Thanks, last minute copy-pasting …

PR #789, opened 2019-03-26
getUnionSetoid, setoidAll, setoidNull and setoidUndefined added

mlegenhausen:
This pull requests adds `getUnionSetoid`, `setoidAll`, `setoidNull` and `setoidUndefined` to the `Setoid` module.

`getUnionSetoid` allows the combination of multiple setoids to one. My use case is the creation of new setoids that union types. Example I want to compare two values of type `string | undefined`.

```ts
const a: string | undefined;
const b: string | undefined;

setoidString.equals(a, b) // TypeScript Error

const S = getUnionSetoid(setoidString, setoidUndefined);
S.equals(a, b); // Works and typechecks correctly
```

`setoidAll` is a helper to ignore certain properties in a struct type cause it always returns `true`.

Example

```ts
interface Foo {
  a: string;
  b: number;
}

const setoidFoo = getStructSetoid<Foo>({
  a: setoidString,
  b: setoidAll
});

setoidFoo.equals({ a: 'a', b: 0 }, { a: 'a', b: 1 }) === true
```

### DIFF
--- a/docs/modules/Setoid.ts.md
+++ b/docs/modules/Setoid.ts.md
@@ -4,27 +4,18 @@ nav_order: 78
 parent: Modules
 ---
 
-# Overview
-
-The `Setoid` type class represents types which support decidable equality.
-
-Instances must satisfy the following laws:
-
-1. Reflexivity: `S.equals(a, a) === true`
-2. Symmetry: `S.equals(a, b) === S.equals(b, a)`
-3. Transitivity: if `S.equals(a, b) === true` and `S.equals(b, c) === true`, then `S.equals(a, c) === true`
-
-See [Getting started with fp-ts: Setoid](https://dev.to/gcanti/getting-started-with-fp-ts-setoid-39f3)
-
 ---
 
 <h2 class="text-delta">Table of contents</h2>
 
 - [Setoid (interface)](#setoid-interface)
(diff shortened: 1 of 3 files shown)


gcanti:
@mlegenhausen thanks for this PR but in general `getUnionSetoid` doesn't produce a lawful `Setoid` (the transitivity law doesn't hold).

A counterexample:

```ts
interface Counterexample {
  a: string
  b: number
}

const S1 = contramap((c: Counterexample) => c.a, setoidString)
const S2 = contramap((c: Counterexample) => c.b, setoidNumber)
const S: Setoid<Counterexample> = getUnionSetoid(S1, S2)

const x: Counterexample = { a: 'foo', b: 1 }
const y: Counterexample = { a: 'foo', b: 2 }
const z: Counterexample = { a: 'bar', b: 2 }

console.log(S.equals(x, y)) // true
console.log(S.equals(y, z)) // true
// however the transitivity law doesn't hold 
console.log(S.equals(x, z)) // false
```

Also the implementation is unsafe since each involved setoid potentially receives invalid values.

mlegenhausen:
@gcanti sadly but true... I will reconsider my main problems regarding `null` and `undefined` and tagged union comparison.

PR #814, opened 2019-04-04
Add es6 module step to build for tree-shaking support

### LINKED ISSUE #813: Ship ES6 module syntax along with CJS modules

Feature

I'm using bits and pieces of fp-ts in a library I'm building, but since fp-ts only ships CJS modules, a lot of code that goes unused by me ends up being bundled with my library anyway (since bundlers can't perform tree-shaking on CJS modules).

Ideally my library should only contain the parts of fp-ts that I'm using in my code.

In theory this should be a matter of running tsc a second time with `--module=es6` and a separate outDir like `--outDir=esm`. You can then point bundlers that support es6 modules (e.g. webpack and rollup) to the es6 entry point by setting `"module": "esm/index.js"` in package.json.

If the codebase is free of side effects at import time (is it?), we can additionally tell bundlers that performing tree-shaking on this library is safe by setting `"sideEffects": false` in package.json ([read more](https://webpack.js.org/guides/tree-shaking/#mark-the-file-as-side-effect-free)).

If it turns out that these are the only steps needed in practice as well, I might prepare a PR soon :-)

FruitieX:
This PR resolves #813 

Adds another step to the build script which outputs a bundle into `/es6` using es6 module syntax. Users can optionally switch to importing from `fp-ts/es6/*` rather than `fp-ts/lib/*` in order to make use of tree-shaking.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,7 @@
 *.log
 node_modules
 lib
+es6
 dev
 .idea
 coverage
--- a/README.md
+++ b/README.md
@@ -57,6 +57,24 @@ npm install fp-ts
 
 Make sure to always have a single version of `fp-ts` installed in your project. Multiple versions are known to cause `tsc` to hang during compilation. You can check the versions currently installed using `npm ls fp-ts` (make sure there's a single version and all the others are marked as `deduped`).
 
+## Consuming fp-ts
+
+Most examples will use the following import syntax:
+
+```
+import { Option, some, none } from 'fp-ts/lib/Option'
+```
+
+This will give you the widest support across tools, as you will be importing CommonJS modules.
+
+If you use a bundler such as webpack or Rollup that supports tree-shaking, you can take advantage of this by opting to import ECMAScript modules instead:
+
+```
+import { Option, some, none } from 'fp-ts/es6/Option'
+```
+
+Note that there are caveats such as some tools (e.g. Jest) not supporting ES6 module syntax natively yet.
+
 ## TypeScript compatibility
 
 **Strictness** – This library is conceived, tested and is supposed to be consumed by TypeScript with the `strict` flag turned on.
--- a/package.json
+++ b/package.json
@@ -3,10 +3,13 @@
   "version": "1.15.0",
   "description": "Functional programming in TypeScript",
   "files": [
-    "lib"
+    "lib",
+    "es6"
   ],
   "main": "lib/index.js",
+  "module": "es6/index.js",
   "typings": "lib/index.d.ts",
+  "sideEffects": false,
   "scripts": {
     "lint": "tslint -p tsconfig.tslint.json src/**/*.ts test/**/*.ts",
     "jest-clear-cache": "jest --clearCache",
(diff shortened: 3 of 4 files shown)


gcanti:
Thanks @FruitieX 

> Users can optionally switch to importing from fp-ts/es6/* rather than fp-ts/lib/* in order to make use of tree-shaking

if I don't ask you too much, could you please add a small section to the README explaining `how to consume fp-ts` / `how to setup a project` in order to enable tree-shaking? It would be very useful (for me too)

FruitieX:
> if I don't ask you too much, could you please add a small section to the README explaining `how to consume fp-ts` / `how to setup a project` in order to enable tree-shaking? It would be very useful (for me too)

I'll look into this later today, along with a small repro comparing the two.

FruitieX:
For easy experimenting, I published an npm package called `fp-ts-es6` which is the result of running `npm publish` on this PR branch.

Quick setup using create-react-app, due to it having tree-shaking support out of the box:

```
npx create-react-app fp-ts-test --typescript
cd fp-ts-test
yarn add fp-ts-es6
```

### With CJS modules:

Append to src/index.tsx:

```
import { empty } from 'fp-ts-es6/lib/Array';

console.log(empty);
```

Result of running `yarn build`:

```
$ react-scripts build
Creating an optimized production build...
Compiled successfully.

File sizes after gzip:

  42.13 KB  build/static/js/2.49b89b21.chunk.js
  764 B     build/static/js/runtime~main.c5541365.js
  739 B     build/static/js/main.2a17070f.chunk.js
  540 B     build/static/css/main.4eb7766e.chunk.css
```

### With ES6 modules:

Append to src/index.tsx:

```
import { empty } from 'fp-ts-es6/es6/Array';

console.log(empty);
```

Result of running `yarn build`:

```
$ react-scripts build
Creating an optimized production build...
Compiled successfully.

File sizes after gzip:

  37.4 KB (-4.73 KB)  build/static/js/2.458b9b6e.chunk.js
  764 B               build/static/js/runtime~main.c5541365.js
  734 B (-5 B)        build/static/js/main.113909e4.chunk.js
  540 B               build/static/css/main.4eb7766e.chunk.css
```

As you can see, we managed to shave off 4.73 KB (after gzip!) from the production app vendor bundle. While not super significant or anything, the savings add up when you import from more modules, especially if using only small bits and pieces of these modules as well.

FruitieX:
Some more observations:

Due to the fact that we're importing from `fp-ts/(lib|es6)/Array` instead of the entry-point `fp-ts`, bundlers and tools like Jest can't automatically select between the CJS and ES6 versions of the library. It also means that users have to change their fp-ts import paths in order to use one bundle format over the other.

This wouldn't matter if it weren't for e.g. Jest not supporting ES6 syntax under `node_modules` (without additional configuration on the user's side). Normally Jest would look inside the library's package.json at the `main` and `module` fields and pick the one which it supports (`main` in this case), but it can only do that for imports directly from the entry-point:

image

So I noticed fp-ts does expose its modules in the entry-point as well, and tested using those:

```
import { array } from 'fp-ts-es6';

console.log(array.empty);
```

```
$ react-scripts build
Creating an optimized production build...
Compiled successfully.

File sizes after gzip:

  40.78 KB (+3.38 KB)  build/static/js/2.0402c198.chunk.js
  764 B                build/static/js/runtime~main.c5541365.js
  739 B (+5 B)         build/static/js/main.1fa0363b.chunk.js
  540 B                build/static/css/main.4eb7766e.chunk.css
```

Slightly larger result vs. importing directly from the module, but still better than in the CJS module test case. Seems that the build now contains all of Array's functions, indicating that webpack was not able to perform tree-shaking on `src/Array.ts`.

I speculate the difference in size compared to the CJS test case is that whatever modules [`src/Array.ts`](https://github.com/gcanti/fp-ts/blob/master/src/Array.ts) is importing do get tree-shaken correctly.

At least Jest is happy:
image

So it seems this isn't as simple as I was hoping from the start, but maybe we still want this change if we document carefully what the downsides are.

gcanti:
Thanks @FruitieX !

PR #817, opened 2019-04-06
New Feature: add Record.fromFoldableMap

gcanti:


src/Record.ts
```diff
@@ -624,32 +624,98 @@ export function filterWithKey<A>(fa: Record<string, A>, p: (key: string, a: A) =
...
+  ta: Type3<F, U, L, A>,
+  f: (a: A) => [K, B],
+  concat: (existing: B, a: B) => B
+) => Record<K, B>
+export function fromFoldableMap<F extends URIS2>(
+  F: Foldable2<F>
+): <L, A, K extends string, B>(
+  ta: Type2<F, L, A>,
+  f: (a: A) => [K, B],
+  concat: (existing: B, a: B) => B
```
Wenqer: I just curious, why we require here concat function instead of semigroup?

gcanti: @Wenqer mmmh admitely the name `concat` is misleading in this context, I'll change it to `onConflict`

> why we require here concat function instead of semigroup?

I'll answer with the opposite question: since requiring a `Semigroup` would add an additional constraint, why resolving conflicts should be associative? Is it possible to justify such a constraint?

raveclassic: What about introducing `Magma`s? Afaics they don't require binary operation to be associative. This would make instance parametrization consistent across whole codebase and also make possible to parametrize `fromFoldableMap` with existing `Semigroup` instances.

raveclassic: related https://github.com/gcanti/fp-ts/pull/431

gcanti: @raveclassic LGTM

raveclassic: @gcanti I'll submit a PR then.

PR #820, opened 2019-04-10
Fix typo.

DenisFrezzato:


gcanti:
@DenisFrezzato that's not a typo, it means that you can prove to the type checker that the array is actually not empty

DenisFrezzato:
I just fixed the word, I didn't delete it.

gcanti:
What I mean is that there's nothing to fix, "provably" (= "in way that can be shown as true") is already the right word for the constraint `Array<A> & { 0: A }`

DenisFrezzato:
Oh, ok, sorry then :)

PR #821, opened 2019-04-10
Create Zipper function fromNonEmptyArray2v.

DenisFrezzato:
This PR adds a function which create a `Zipper` from a `NonEmptyArray2v`.


src/Zipper.ts
```diff
@@ -193,6 +194,13 @@ export const fromNonEmptyArray = <A>(nea: NonEmptyArray<A>): Zipper<A> => {
   return new Zipper(empty, nea.head, nea.tail)
 }
 
+/**
+ * @since 1.9.0
```
gcanti: I guess this will be released in `1.17.0`

gcanti:
Thanks @DenisFrezzato

PR #827, opened 2019-04-13
Docs Ramda

grossbart:
Added more similarities to Ramda and fixed some broken links.


gcanti:
Nice, thank you @grossbart

PR #828, opened 2019-04-13
Document Magma

grossbart:
I added Magma to the type classes diagram and to the index. I _think_ it is in the right place in the diagram, but a second pair of eyes would certainly help :)


gcanti:
Thanks @grossbart

PR #830, opened 2019-04-15
New Feature: add Show type class

gcanti:


grossbart:
To resolve the issue from #408, would it make sense to log something like "Please use the Show typeclass to convert this type to a string" instead of just letting the `inspect` call happen? Maybe remove `inspect` completely (for v2)?

gcanti:
> Maybe remove inspect completely (for v2)?

Yes, that's my intention, `inspect` and `toString` will be removed

PR #846, opened 2019-04-18
Breaking Change: change type parameters of Either.left and Either.right, closes #543

### LINKED ISSUE #543: Revise 'left' and 'right' typings for typesafety
Today's typing for 'left' and 'right' are unsafe due to the way typescript does it's inference and causing us some trouble.

Example:
```typescript
export declare const left: <L, A>(l: L) => Either<L, A>;
export declare const right: <L, A>(a: A) => Either<L, A>;

const res = (cond: boolean) => (cond ? left(42) : right('what?')) // return type: Left<number, {}> | Right<number, {}> | Left<{}, string> | Right<{}, string>
const v = res(true)
if (v.isRight()) {
  v.value // string | {} (!! WRONG)
} else {
  v.value // number | {} (!! WRONG)
}
```

Thanks to the way union behave with never
```typescript
type T = number | never // Type is number
```

With these typings inference works as expected
```typescript
export declare const left: <L>(l: L) => Either<L, never>;
export declare const right: <A>(a: A) => Either<never, A>;

const res = (cond: boolean) => (cond ? left(42) : right('what?')) // return type: Left<number, never> | Right<number, never> | Left<never, string> | Right<never, string>
const v = res(true)
if (v.isRight()) {
  v.value // string (!! OK)
} else {
  v.value // number (!! OK)
}

```

I know this would be a breaking change,
so here's a backward compatible non breaking solution:

```typescript
export declare const left: <L, A = never>(l: L) => Either<L, A>
export declare const right: <L = never, A = any>(a: A) => Either<L, typeof a>
```

Do you see any issue with it? (Being typesafe is priceless)

gcanti:
/cc @sledorze

### DIFF
--- a/dtslint/ts3.4/index.ts
+++ b/dtslint/ts3.4/index.ts
@@ -279,8 +279,8 @@ R.sequence(O.option)(ro1) // $ExpectType Option<Record<"a" | "b", number>>
 
 R.record.compact(do1) // $ExpectType Record<string, number>
 
-R.partitionMapWithKey(d1, (_k: string, n) => E.right<string, number>(n)) // $ExpectType Separated<Record<string, string>, Record<string, number>>
-R.partitionMapWithKey(r1, (_k: 'a' | 'b', n) => E.right<string, number>(n)) // $ExpectType Separated<Record<string, string>, Record<string, number>>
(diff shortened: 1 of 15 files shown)


sledorze:
@gcanti That will be so useful, thanks!
Eager to test the V2

gcanti:
There are other modules where we can apply the same pattern, I'll send a PR for them too

PR #850, opened 2019-04-23
remove classes + other breaking changes

gcanti:


src/Either.ts
```diff
@@ -20,8 +20,10 @@
  * operations like `map`, `chain`, ... return the `Left` value unchanged:
  *
  * ```ts
- * right(12).map(double) // right(24)
- * left(23).map(double)  // left(23)
+ * import { either } from 'fp-ts/lib/Either'
+ *
+ * either.map(right(12, double) // right(24)
+ * either.map(left(23, double)  // left(23)
```
giogonzo: missing `)` after `12` and `23`

src/Array.ts
```diff
@@ -968,15 +960,15 @@ export const reverse = <A>(as: Array<A>): Array<A> => {
...
+ * @since 2.0.0
  */
 export const rights = <L, A>(as: Array<Either<L, A>>): Array<A> => {
   const r: Array<A> = []
   const len = as.length
   for (let i = 0; i < len; i++) {
     const a = as[i]
-    if (a.isRight()) {
-      r.push(a.value)
+    if (a._tag === 'Right') {
```
giogonzo: is this for perf reasons (instead of using `isRight()`)?

gcanti: to avoid a runtime dependency

src/TaskEither.ts
```diff
@@ -34,246 +29,141 @@ export const URI = 'TaskEither'
...
-  return new TaskEither(fl.map(eitherLeft))
+export function orElse<L, A, M>(ma: TaskEither<L, A>, f: (l: L) => TaskEither<M, A>): TaskEither<M, A> {
+  return task.chain(ma, e => E.fold<L, A, Task<E.Either<M, A>>>(e, f, eitherT.of))
 }
 
 /**
- * @since 1.0.0
+ * @since 2.0.0
  */
-export const fromEither = <L, A>(fa: Either<L, A>): TaskEither<L, A> => {
```
giogonzo: this is now `taskEither.fromEither` from `MonadThrow`, correct?
It seems a bit weird to only have this removed (and not e.g. `fromLeft` which would be already available as `taskEither.throwError`)

src/Option.ts
```diff
@@ -712,71 +530,62 @@ export const isNone = <A>(fa: Option<A>): fa is None<A> => {
...
+    map(fa, e => ({
+      left: getLeft(e),
+      right: getRight(e)
+    })),
+    defaultSeparate
+  )
 }
 
-const filter = <A>(fa: Option<A>, p: Predicate<A>): Option<A> => fa.filter(p)
+const filter = <A>(fa: Option<A>, p: Predicate<A>): Option<A> => (isNone(fa) ? fa : p(fa.value) ? fa : none)
```
giogonzo: @gcanti I think we still need also the equivalent of the old `.filter` method, accepting a `Refinement<A, B>`. This is some code that doesn't compile anymore:
```ts
import { option, some } from 'fp-ts/lib/Option';

declare const a: string | number;
declare function isString(x: any): x is string;

const b: string = option.filter(some(a), isString);
```

src/TaskEither.ts
gcanti: @giogonzo `fromLeft` and `taskEither.throwError` have different signatures (i.e. `never` in the former).

We can re-add a redundant `fromEither` later, I just want a clean baseline to start from.

src/Option.ts
gcanti: @giogonzo I think that we can handle refinements (in `filter` and `partition`) once for all by adding proper overloads to the `Filterable` type class. This is beneficial for v1 too, I'll send a PR for v.1.17.2

semyon2105:
Could you explain why you started replacing const arrow functions with regular functions? Are they faster/better in some way?

gcanti:
@semyon2105 no it's just a matter of style. The first docs generator I wrote had some problems but the current one can handle `function`s just fine, even with overloads.

PR #852, opened 2019-04-25
Breaking Change: rename Setoid to Eq

gcanti:


mattgrande:
Can I ask why? Most of my knowledge of this stuff comes from [fantasy-land](https://github.com/fantasyland/fantasy-land#setoid), and it looks like they use Setoid for this as well. Is there a difference between Setoid and Eq?

gcanti:
@mattgrande looks like "Setoid" is [not popular](https://github.com/gcanti/fp-ts/issues/823#issuecomment-486146335) amongst `fp-ts` users

PR #853, opened 2019-04-26
Polish: add refinement overloads to filter / partition (Filterable ty…

gcanti:
…pe class)

This change makes obsolete to define specific overloads of `filter` / `partition` in order to handle refinements (e.g. in the `Array` module) by adding overloads to the `Filterable` type class.

This change is also beneficial to the data types which don't admit a unconstrained `Filterable` instance (e.g. `Either`)

```ts
import * as E from 'fp-ts/lib/Either'
import { monoidAll } from 'fp-ts/lib/Monoid'

declare function isString(x: unknown): x is string

const filterableEither = E.getFilterable(monoidAll)

filterableEither.filter(E.right<boolean, string | number>(1), isString)
// before: E.Either<boolean, string | number>
// after: E.Either<boolean, string>

filterableEither.partition(E.right<boolean, string | number>(1), isString)
// before: Separated<Either<boolean, string | number>, Either<boolean, string | number>>
// after: Separated<Either<boolean, string | number>, Either<boolean, string>>
```

/cc @giogonzo

### DIFF
--- a/docs/modules/Array.ts.md
+++ b/docs/modules/Array.ts.md
@@ -28,7 +28,7 @@ Adapted from https://github.com/purescript/purescript-arrays
 - [dropEnd (function)](#dropend-function)
 - [dropWhile (function)](#dropwhile-function)
 - [elem (function)](#elem-function)
-- [filter (function)](#filter-function)
+- [~~filter~~ (function)](#filter-function)
 - [findFirst (function)](#findfirst-function)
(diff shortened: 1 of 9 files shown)


giogonzo:
@gcanti should we also `@deprecated` the instance methods in `Option` etc.?

gcanti:
@giogonzo no, I'd wait a final decision on v2 for that.

However we can deprecate:

- `Array.filter`
- `Array.partition`

Issue #531, opened 2018-07-19
Suggestion: introductory documentation


piersmacdonald:
This is a suggestion and perhaps I can help but the documentation for this repo is very hard to read for someone without a background in FP.

I think typescript's  evolving type system and JS roots makes it a good cross over language where OO developers can learn functional techniques, myself included. Therefore this library has a tonne of potential to be an introduction. 

For example:

- I know function binding/chaining is a basic part of FP. I _think_ I have to use the `Chain` module to do that but I'm not sure and there's no documentation showing that.

- The docs that do exist have very little grounding in application settings. Examples which say something like `fab -> fba` or `HKT<A,B>` provide little reference for what's actually being attempted. If I don't know what `HKT` is or what `A` or `B` represent then I have little understand of what something does.

I'd love to help improve docs as I learn FP techniques myself. What's the preferred format to contribute? Readmes?

baldegit:
hi piersmacdonald

I totally agree with you. I am a Java programmer and am currently working on a project typeScript where we use fp-ts and io-ts. and I must confess that I have a lot of trouble.

grossbart:
I started an effort to make the documentation more accessible to beginners: https://github.com/gcanti/fp-ts/pull/779 It's not where it should be yet (by a far margin), but I plan to work on this over the next weeks. I'll leave this issue open for now until this use case is covered properly 😄

cyberixae:
I would appreciate more practical usage examples. The documentation at the beginning of Options.ts chapter is good. Number one problem at the moment is figuring out how to import the things you need since there are very few code examples with import statements showing how that works.

cyberixae:
I have some background in functional programming, however I'm struggling to deal with objects when using fp-ts. Maybe this is something that the documentation could address. For example I'm not sure what is the idiomatic fp-ts way of transforming...
```
{
  a: Either<Error,A>,
  b: Either<Error,B>,
}
```
...to the following...
```
Either<Error,{
  a: A,
  b: b,
}>
```
[...]

PR #854, opened 2019-04-27
Add code examples to IO Overview

cyberixae:
This PR adds beginner friendly code examples from #531 to IO Overview. Some of the sequence code examples might be useful for other types with slight modifications. However, I think we may need to repeat some of the more generic concepts in several parts of the documentation since people who are reading the documentation do not know where to look for the more generic documentation. They only know that they are right now working with `IO` and wondering how to deal with problems that arise from dealing with `IO`. The fact that there are other types like `Either` which are similar and support similar operations do not help them at that point. However, as they are reading through several pages of documentation they will hopefully notice the pattern between the two.

### DIFF
--- a/src/IO.ts
+++ b/src/IO.ts
@@ -1,6 +1,94 @@
 /**
- * @file `IO<A>` represents a synchronous computation that yields a value of type `A` and **never fails**.
- * If you want to represent a synchronous computation that may fail, please see `IOEither`.
+ * @file `IO<A>` represents a non-deterministic synchronous computation that can cause side effects, yields a value of
+ * type `A` and **never fails**. If you want to represent a synchronous computation that may fail, please see
+ * `IOEither`.
+ *
+ * IO actions are terminated by calling their `run()` method that executes the computation and returns the result.
+ * Ideally each application should call `run()` only once for a root value of type `Task` or `IO` that represents the entire
+ * application. However, this might vary a bit depending on how you construct your application.  An application
+ * framework with fp-ts types might take care of calling `run()` for you, while another application framework without
+ * fp-ts typing might force you to call `run()` at multiple locations whenever the framework demands less strictly typed
+ * values as inputs for its method calls.
+ *
+ * Below are some basic examples of how you can wrap values and function calls with `IO`.
+ *
+ * ```ts
+ * import { IO, io } from 'fp-ts/lib/IO'
+ *
+ * const constant: IO<number> = io.of(123);
+ * constant.run()  // returns 123
+ *
+ * const random: IO<number> = new IO(() => Math.random())
+ * random.run()  // returns a random number
+ * random.run()  // returns another random number
+ *
+ * const log = (...args): IO<void> => new IO(() => console.log(...args));
+ * log('hello world').run()  // returns undefined and outputs "hello world" to console
+ * ```
+ *
+ * In the example above we implemented type safe alternatives for `Math.random()` and `console.log()`. The main
+ * motivation was to explain how you can wrap values. However, fp-ts provides type safe alternatives for such basic
+ * tools through `Console` and `Random` modules. So you don't need to constantly redefine them.
+ *
+ * The next code snippet below is an example of a case where type safety affects the end result. Using `console.log()`
+ * directly would break the code, resulting in both logging actions being executed when the value is not `null`.  You
+ * can confirm this by removing `.run()` from the end of the example code and replacing calls to `log()` with
+ * standard`console.log()`.
+ *
+ * ```ts
+ * import { IO } from 'fp-ts/lib/IO'
+ * import { fromNullable } from 'fp-ts/lib/Option'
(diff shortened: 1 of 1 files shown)


src/IO.ts
```diff
@@ -1,6 +1,74 @@
 /**
- * @file `IO<A>` represents a synchronous computation that yields a value of type `A` and **never fails**.
- * If you want to represent a synchronous computation that may fail, please see `IOEither`.
+ * @file `IO<A>` represents a synchronous computation that yields a value of type `A` and **never fails**.  If you want
+ * to represent a synchronous computation that may fail, please see `IOEither`.  IO actions are terminated by calling
+ * their `run()` method that executes the computation and returns the result.
```
giogonzo: Maybe a note regarding `run()`  should be called sparingly or even only once at the "end of the program" in order to maintain the referential transparency and composability advantages?

src/IO.ts
```diff
@@ -1,6 +1,74 @@
...
+ *
+ * const safeLog = (...args): IO<void> => new IO(() => console.log(...args));
+ *
+ * const logger = (input: number|null): IO<void> => fromNullable(input).fold(
+ *   safeLog('Received null'),
+ *   (value) => safeLog(`Received ${value}`),
+ * );
+ *
+ * logger(123).run();
+ * ```
```
giogonzo: maybe add a reference to https://github.com/gcanti/fp-ts/blob/master/src/Console.ts ?

src/IO.ts
```diff
@@ -1,6 +1,74 @@
...
+ *
+ * It is often convenient to combine several `IO` actions into one for sequential execution.
+ * You can use `sequence_(io, array)` to do this.
```
giogonzo: I would clarify that `sequence_` is helpful if you want to discard the IO results

src/IO.ts
grossbart: I think it would also make sense to update this to say "… represents a synchronous effectful computation", because `IO` lets us describe side-effects and non-deterministic computations.

src/IO.ts
```diff
@@ -1,6 +1,74 @@
...
+ * to represent a synchronous computation that may fail, please see `IOEither`.  IO actions are terminated by calling
+ * their `run()` method that executes the computation and returns the result.
+ *
+ * ```ts
+ * import { IO, io } from 'fp-ts/lib/IO'
+ *
+ * const constant: IO<number> = io.of(123);
+ * constant.run()  // returns 123
+ *
+ * const random: IO<number> = new IO(() => Math.random())
```
grossbart: Similar to @giogonzo's comment on `Console.ts`, you could mention [Random](https://github.com/gcanti/fp-ts/blob/master/src/Random.ts). Maybe the mention of these two would make sense after the examples or individually just before you introduce them.

src/IO.ts
```diff
@@ -1,6 +1,74 @@
...
+ *   foo: number,
+ *   bar: string,
+ * }
+ *
+ * const computations: { [K in keyof Result]: IO<Result[K]>} = {
+ *   foo: io.of(123),
+ *   bar: io.of('example')
+ * }
+ *
+ * const computation = sequenceS(io)(computations)
```
grossbart: At the moment the type `Result` is not really used, I would either add it here to show what's happening:

```ts
const computation: IO<Result> = sequenceS(io)(computations)
```

Or remove the types completely (they can be inferred) because explaining how sequence works might be better done in the `Apply` docs.

src/IO.ts
```diff
@@ -1,6 +1,74 @@
...
+ * It is often convenient to combine several `IO` actions into one for sequential execution.
+ * You can use `sequence_(io, array)` to do this.
+ * ```ts
+ * import { IO, io } from 'fp-ts/lib/IO'
+ * import { array } from 'fp-ts/lib/Array'
+ * import { sequence_ } from 'fp-ts/lib/Foldable2v'
+ *
+ * const safeLog = (...args): IO<void> => new IO(() => console.log(...args));
+ *
+ * const logFoo: IO<void> = safeLog('foo');
```
grossbart: If you can think of a more realistic example instead of "foo" and "bar" that would be nice (some people don't know what "foo" means). If not, that is fine.

cyberixae:
I updated the PR based on feedback. I replaced foo and bar with children's book style examples. Not sure if this is ideal but perhaps the examples are easier to visualise even thought these are still far from real world usage examples.

grossbart:
Thanks for your effort, @cyberixae!

cyberixae:
Merging the PR did not update the actual documentation. How does the release process work? Was I supposed to include a prebuilt document in the pull request?

grossbart:
Ah yes, sorry, you need to run `npm run docs` for that. I'll do this now.

grossbart:
It's there now: https://gcanti.github.io/fp-ts/modules/IO.ts.html

PR #857, opened 2019-05-02
Defect Fix: remove `reverse` (mutable) from `NonEmptyArray2v` interface

gcanti:


src/NonEmptyArray2v.ts
```diff
@@ -33,7 +33,6 @@ export interface NonEmptyArray<A> extends Array<A> {
   0: A
   map: <B>(f: (a: A, index: number, nea: NonEmptyArray<A>) => B) => NonEmptyArray<B>
   concat: (as: Array<A>) => NonEmptyArray<A>
-  reverse: () => NonEmptyArray<A>
```
raveclassic: What about extending `ReadonlyArray<A>`? This will drop all mutable methods.

gcanti: @raveclassic we can't:

- raises a compiler error (`Interface 'NonEmptyArray<A>' incorrectly extends interface 'readonly A[]'...`)
- would be a breaking change

Also handling `ReadonlyArray`s is just annoying without providing real benefits over a lint rule

PR #859, opened 2019-05-07
New Feature: add absurd function, closes #847

### LINKED ISSUE #847: feature proposal: absurd function for never type

fearure

---

It might be helpful for us to have `absurd` function somewhere (perhaps in `function.ts`?), especially when folding `Either<never, A>` to `A`.

```typescript
function absurd<A>(_: never): A {
    throw new Error('absurd');
}
```

gcanti:


test/function.ts
```diff
@@ -325,4 +326,8 @@ describe('function', () => {
     // tslint:disable-next-line: deprecation
     assert.deepStrictEqual(tupleCurried(1)('a'), [1, 'a'])
   })
+
+  it('absurd', () => {
+    assert.throws(() => absurd<string>((null as any) as never))
```
jethrolarson: Seems like a hole in the language that you can cast to `never` :)

src/function.ts
```diff
@@ -418,3 +418,10 @@ export const increment = (n: number): number => {
 export const decrement = (n: number): number => {
   return n - 1
 }
+
+/**
+ * @since 1.18.0
+ */
+export function absurd<A>(_: never): A {
+  throw new Error('absurd')
```
jethrolarson: Could put a more descriptive error here. "Called `absurd` function which should be uncallable".

PR #861, opened 2019-05-08
optimize function.concat

gcanti:


jethrolarson:
lgtm

mattgrande:
I see this was merged a few days ago, and this isn't affecting me, but I thought I'd raise a potential concern.

This changes the behaviour slightly; Prior to this, you always got a brand new array. Now you (sometimes) get back the exact same array.

Consider something like this:

```
const ar1 = [1, 2, 3];
const ar2 = concat(ar1, []);
ar2.push(4);  // This will affect ar1
```

Again, this is _probably_ okay, but I wanted to make sure y'all were aware.

PR #862, opened 2019-05-08
Added some documentation to Record.ts

jethrolarson:

### DIFF
--- a/src/Record.ts
+++ b/src/Record.ts
@@ -43,7 +43,16 @@ export const isEmpty = <A>(d: Record<string, A>): boolean => {
 }
 
 /**
+ * Map a record into an array
  * @since 1.10.0
+ * @example
+ * import {collect} from 'fp-ts/lib/Record'
+ *
+ * const ob: {a: string, b: boolean} = {a: 'foo', b: false}
+ * assert.deepStrictEqual(
+ *   collect(ob, (key, val) => ({key: key, value: val})),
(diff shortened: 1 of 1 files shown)


gcanti:
Thanks @jethrolarson !

src/Record.ts
```diff
@@ -225,7 +236,16 @@ export function map<A, B>(fa: Record<string, A>, f: (a: A) => B): Record<string,
 }
 
 /**
+ * Reduce object by iterating over it's values.
+ *
  * @since 1.10.0
+ *
+ * Warning: If your operation isn't commutative and the record's keys don't have consistent order you may get inconsistent results
```
gcanti: > the record's keys don't have consistent order

What do you mean?

jethrolarson: see the example

jethrolarson: See also
https://github.com/ramda/ramda/issues/2046

gcanti: > `reduce({a: 1, b: 2}, f) != reduce({b: 2, a: 1}, f)`

@jethrolarson Actually they are equal since keys are sorted https://github.com/gcanti/fp-ts/blob/7c090c5d0ee7d1257f56035b87351626e19bfa73/src/Record.ts#L276

jethrolarson: I could swear I looked for that. Okay. I will revise

jethrolarson: Updated

PR #865, opened 2019-05-10
Don't set `target: es6` in tsconfig.es6.json

### LINKED ISSUE #863: Broken in browsers that don't support ES6 syntax natively

Bug

The `module` field in `package.json` points to a file that uses ES6 syntax, not just ES Module syntax. This causes syntax errors when trying to run the code in a browser that does not support ES6, such as IE11.

**If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem via https://codesandbox.io/ or similar.**

1. Bundle `fp-ts` with a bundler such as Webpack or Rollup (it is possible to configure bundlers to ignore the `module` field in `package.json`, but the default configuration gives it preference over the `main` field)
2. Try to run the resulting bundle in a browser that does not support ES6 syntax natively, such as IE11
3. Should result in a syntax error

According to https://github.com/rollup/rollup/wiki/pkg.module:

> pkg.module will point to a module that has ES2015 module syntax but otherwise only syntax features that the target environments support.

The `module` field should point a file that uses `import`/`export` syntax, but no other ES6 features.

More details here: https://github.com/rollup/rollup/wiki/pkg.module#wait-it-just-means-import-and-export--not-other-future-javascript-features

> If you're using things like arrow functions and classes (or more exotic features like decorators and object spread), you should transpile them if you want your library to work in environments that don't support those features. Otherwise, whoever uses your library will have to transpile it themselves, including configuring their transpiler to handle your code, which might involve esoteric transforms and conflicting versions of Babel.

Browsers: IE11 and any other browser that don't support ES6 syntax natively
Previous version: It worked in previous versions of `fp-ts` that did not expose an ES6 version (i.e. before this issue was resolved: https://github.com/gcanti/fp-ts/issues/813)

FruitieX:
Resolves #863

### DIFF
--- a/tsconfig.es6.json
+++ b/tsconfig.es6.json
@@ -2,7 +2,6 @@
   "extends": "./tsconfig.json",
   "compilerOptions": {
     "outDir": "./es6",
-    "target": "es6",
     "module": "es6"
   }
 }


gcanti:
Thanks @FruitieX

PR #867, opened 2019-05-10
export the Fluent types

leemhenson:
If they are not exported, writing a line of code in your app like this:

```ts
export const fluentEither = fluent(either);
```

gives:

`Exported variable 'fluentEither' has or is using name 'Fluent2' from external module "/Users/src/app/node_modules/fp-ts/lib/fluent" but cannot be named.ts(4023)`

### DIFF
--- a/src/fluent.ts
+++ b/src/fluent.ts
@@ -28,7 +28,7 @@ import { Eq } from './Eq'
 import { Show } from './Show'
 import { Semigroupoid, Semigroupoid2, Semigroupoid2C } from './Semigroupoid'
 
-interface Fluent2C<F extends URIS2, I, L, A> {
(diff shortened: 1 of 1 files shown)


gcanti:
@leemhenson ah right, thanks. Speaking of which, what about adding an official `fluent` export to all data types?

leemhenson:
Yeah I imagine at least the obvious ones `Option`, `Either`, `TaskEither` should be available.

gcanti:
Nevermind, thinking twice only the user can decide when it makes sense and/or augment the standard instance of a data type (for example merging an `Eq` instance for `Option`)

(just pushed this change to `v2-lib`)

PR #870, opened 2019-05-15
Use explicit concat function for getEndomorphismMonoid

mlegenhausen:
Problem:

```ts
import { fold, getEndomorphismMonoid } from 'fp-ts/lib/Monoid';

fold(getEndomorphismMonoid<number>())([a => a + 1, b => b * 2])(3)
```

throws an error

```
Uncaught TypeError: fns[i].call is not a function
    at function.js:100
```

The problem is that `compose` gets called with 4 instead of 2 parameters. The additional two are `index` and `array` from the `reduce` function call.

### DIFF
--- a/src/Monoid.ts
+++ b/src/Monoid.ts
@@ -1,5 +1,5 @@
 import { Bounded } from './Bounded'
-import { compose, Endomorphism, identity, concat } from './function'
+import { Endomorphism, identity, concat } from './function'
 import {
   fold as foldSemigroup,
   getDictionarySemigroup,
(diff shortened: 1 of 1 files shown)


gcanti:
@mlegenhausen IMO the fix is to not define an implicit variadic `concat` in `getEndomorphismMonoid`

```diff
export const getEndomorphismMonoid = <A = never>(): Monoid<Endomorphism<A>> => {
  return {
-   concat: compose,
+   concat: (x, y) => a => x(y(a)),
    empty: identity
  }
}
```

mlegenhausen:
@gcanti Done

gcanti:
Thanks @mlegenhausen

PR #880, opened 2019-06-05
Makes array chop easier to use

Stouffi:
This PR changes the signature of the chop function to use `NonEmptyArray` as the type of the argument of the function passed as second argument because it's called always with a `NonEmptyArray`.

This PR also adds a simple type predicate to allow narrowing down an `Array` to a `NonEmptyArray`, which is useful for while loop, if-else or ternary operator, but also in all functions where a type predicate has to be provided.


gcanti:
Thanks @Stouffi

PR #881, opened 2019-06-05
version 1.19.0 (backport)

gcanti:
Changelog:

# 1.19.0

The goal of this release is to make the migration to v2 easier.

Since in v2 data types are no more implemented with classes, chainable APIs will be deprecated (in v1.20.0).

As an alternative, a `pipe` function is provided, along with suitable data-last top level functions (one for each deprecated method).

**Example**

Before

```ts
import * as O from 'fp-ts/lib/Option'

O.some(1)
  .map(n => n * 2)
  .chain(n === 0 ? O.none : O.some(1 / n))
  .filter(n => n > 1)
  .foldL(() => 'ko', () => 'ok')
```

After

```ts
import * as O from 'fp-ts/lib/Option'
import { pipe } from 'fp-ts/lib/pipeable'

pipe(
  O.some(1),
  O.map(n => n * 2),
  O.chain(n === 0 ? O.none : O.some(1 / n)),
  O.filter(n => n > 1),
  O.fold(() => 'ko', () => 'ok')
)
```

## Custom tslint rule

In order to make easier to spot all the occurrences of chainable APIs without depending on `@deprecated`, which would force you to migrate in one shot, a custom tslint rule is provided (`@obsolete`).

**Configuration**

Add the following lines to your `tslint.json` to turn the `@obsolete` rule on:

```diff
{
+  "rulesDirectory": ["./node_modules/fp-ts/rules"],
   "rules": {
+    "obsolete": true
   }
}
```

- **New Feature**
  - add `Eq` module (@gcanti)
  - backport top level data-last functions from v2 (@gcanti)
  - backport `pipeable` module form v2 (@gcanti)
  - backport `pipe` function form v2 (@gcanti)
  - backport `flow` function form v2 (@gcanti)
  - `Array`
    - add `isNonEmpty` (@Stouffi)
  - `IOEither`
    - add `foldIO` and `foldIOEither` (@bwlt)
  - `Record`
    - add `record` instance (@gcanti)
    - add `mapWithIndex` (@gcanti)
    - add `reduceWithIndex` (@gcanti)
    - add `foldMapWithIndex` (@gcanti)
    - add `reduceRightWithIndex` (@gcanti)
    - add `hasOwnProperty` (@gcanti)
  - `Map`
    - backport `updateAt` (@gcanti)
    - backport `modifyAt` (@gcanti)
  - `Ord`
    - make `Ord` a contravariant functor (@gcanti)
- **Bug Fix**
[...]

PR #883, opened 2019-06-10
Add applyEither

mstn:
Function `applyEither` returns a function that applies `f` or `g` depending on the type of input. It is common in fp practice (although trivial), but I have not found an implementation in `fp-ts`. Sorry in advance if it is a dup.


gcanti:
@mstn thanks, there's an open PR for this, see https://github.com/gcanti/fp-ts/pull/881 (in particular this commit https://github.com/gcanti/fp-ts/pull/881/commits/90cdfd4353652182dec1dbc93ce4c1437519c1b9)

PR #884, opened 2019-06-14
Add kleisli-ts module

YBogomolov:
I've created a [module with Kleisli arrows](https://github.com/YBogomolov/kleisli-ts), a port of Scala ZIO's `KleisliIO` by John A. De Goes. It uses MonadThrow and Bifunctor instances from fp-ts.


gcanti:
Thanks @YBogomolov

Issue #903, opened 2019-07-10
migration to 2.0


dakom:
## 📖 Documentation

Just upgraded to 2.0 and hit with hundreds of errors like:

```
Property 'chain' does not exist on type 'Either<string, Slide>'
```

or, more generically:

```
TS2339: Property 'fold' does not exist on type 'Either<E, T>'.
  Property 'fold' does not exist on type 'Left<E>'.

TS2339: Property 'chain' does not exist on type 'Option<T[]>'.
  Property 'chain' does not exist on type 'None'.
```

Is there a migration guide somewhere?

(also seems StrMap is gone?)

qwert42:
It looks like you are supposed to do

```typescript
import {either} from "fp-ts/lib/Either";

either.map(something, x => y);
```

This brings me to another question -- how would you do X number of maps in a chain (e.g. `something.map(x => y).map(y => z)`)?

Lermex:
FYI, the main v2 discussion was happening in #823 and you can read it as a kind of migration guide.

Short answer: you're now supposed to use `import { pipe } from 'fp-ts/lib/pipeable';`

qwert42:
Ah. Thanks. That's what I figured...

TLDR: for the example `something.map(x => y).map(y => z)`, you would write

```typescript
import F from "fp-ts/lib/YourFavoriteFunctorInstance";
import {pipe} from "fp-ts/lib/pipeable";

const fz = pipe(something,
    F.map(x => y),
    F.map(y => z));
```

gcanti:
@dakom you can check out the release notes of [v1.19.0](https://github.com/gcanti/fp-ts/releases/tag/1.19.0), many features of v2 have been backported to v1 so you can experiment / upgrade your code with the new API before switching to `fp-ts@2.x`

TLDR

Since in v2 data types are no more implemented with classes, chainable APIs will be deprecated.

As an alternative, a `pipe` function is provided, along with suitable data-last top level functions (one for each deprecated method).

**Example**

Before

```ts
import * as O from 'fp-ts/lib/Option'

O.some(1)
  .map(n => n * 2)
  .chain(n => n === 0 ? O.none : O.some(1 / n))
  .filter(n => n > 1)
  .foldL(() => 'ko', () => 'ok')
```

After

```ts
import * as O from 'fp-ts/lib/Option'
import { pipe } from 'fp-ts/lib/pipeable'

pipe(
  O.some(1),
  O.map(n => n * 2),
[...]

PR #905, opened 2019-07-12
Add an upgrade guide for v2

grossbart:
I started an upgrade guide for the v1 -> v2 transition, I hope it helps. Depending on how many questions are asked, this guide might evolve.

I hoped to do this earlier, but time is sparse at the moment, sigh …

### DIFF
--- a/docs/introduction/upgrade-to-v2.md
+++ b/docs/introduction/upgrade-to-v2.md
@@ -1,0 +1,142 @@
+---
+title: Upgrade to fp-ts 2.x
+parent: Introduction
+nav_order: 5
+has_toc: false
+---
+
+# Upgrade to version 2.x
+
+`fp-ts@2.x` brings with it some major improvements, but also breaking changes and the removal of deprecated APIs. This document will help you understand what changed and how you can upgrade your existing codebase.
+{: .fs-6 .fw-300 }
+
+---
+
+The major changes in `fp-ts@2.x` are:
+
+- Requires TypeScript 3.5+
+- `fp-ts@1.19.x` has been released with backported 2.x features for a gradual upgrade path
+- Data types are no longer implemented as classes, resulting in a new API using `pipe`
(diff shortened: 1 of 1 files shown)


dakom:
Until it is merged - I guess it makes sense to add suggestions/questions here?

One that just bit me - seems there is no longer `O.getOrElseL()`, rather `O.getOrElse()` is _always_ lazy?

grossbart:
I'll add a note to the document, the discussion happened here: https://github.com/gcanti/fp-ts/issues/823#issuecomment-500301156

tl;dr having a single API is preferrable, and always lazy is more general.

docs/introduction/upgrade-to-v2.md
```diff
@@ -0,0 +1,142 @@
...
++  "rulesDirectory": ["./node_modules/fp-ts/rules"],
+   "rules": {
++    "obsolete": true
+   }
+}
+```
+
+### Dependencies
+
+Make sure to always have a single version of `fp-ts` installed in your project. Multiple versions are known to cause `tsc` to hang during compilation. You can check the versions currently installed using `npm ls fp-ts`: make sure there's a single version and all the others are marked as `deduped`.
```
giogonzo: I think this (tsc hanging) will be no longer the case with fp-ts >=2 ? I remember someone mentioning it in multiple occasions, but never verified myself. cc @gcanti

leemhenson: If _all_ the versions you have installed at >= 2 then I _think_ the issue will no longer occur.

grossbart: I think you're right, this was one of the motivations behind the removal of classes.

dakom:
Suggest mentioning (as mentioned in https://github.com/gcanti/fp-ts/issues/903) that `pipe()` is helpful, even with only one function, for the following reasons:

1. type inference
2. allows for easier migration since args don't need to be flipped around

PR #906, opened 2019-07-12
Explain TaskEither manipulation

cyberixae:
This pull request contains my attempt at explaining how to convert between different combinations of `Task`, `Either` and `TaskEither`. It might not be perfect. I think I properly understood some parts of it for the first time today. :-) See #531 for related discussion.

### DIFF
--- a/docs/modules/TaskEither.ts.md
+++ b/docs/modules/TaskEither.ts.md
@@ -9,6 +9,38 @@ parent: Modules
 `TaskEither<E, A>` represents an asynchronous computation that either yields a value of type `A` or fails yielding an
 error of type `E`. If you want to represent an asynchronous computation that never fails, please see `Task`.
 
+Typically `TaskEither` provides convenience over working with `Task` and `Either` separately. However, you may need
+to combine or separate them to interface with functions that use one or the other. Below are some recipes to get you
+out of a pickle.
+
+```ts
+import * as t from 'fp-ts/lib/Task'
+import * as e from 'fp-ts/lib/Either'
+import * as te from 'fp-ts/lib/TaskEither'
+
+function te_from_t<R>(input: t.Task<R>): te.TaskEither<never, R> {
+  return te.right(input)
+}
+
+function te_from_e<L, R>(input: e.Either<L, R>): te.TaskEither<L, R> {
+  return te.fromEither(input)
+}
+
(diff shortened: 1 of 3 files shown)


docs/modules/TaskEither.ts.md
```diff
@@ -9,6 +9,38 @@ parent: Modules
...
+  return te
+    .right<L, e.Either<L, R>>(input)
+    .chain(lr => lr.fold(l => te.left(t.task.of(l)), r => te.right(t.task.of(r))))
+}
+
+function t_e_from_te<L, R>(input: te.TaskEither<L, R>): t.Task<e.Either<L, R>> {
+  return input.value
+}
+```
+
```
anilanar: Isn't this doc generated from `EitherTask.ts`?

cyberixae: It is but the project is lacking a continuous integration mechanism that would update the documentation based on changes in the source code so we need to update it manually whenever we make changes to the source document,

gcanti:
@cyberixae thanks for this PR but except for `te_from_e_t` these functions are just alias of built-in functions.

Note 1. `te_from_t_e` can be simplified, basically it's the `TaskEither` constructor

Note 2. Based on the code I guess you are targeting the `1.x` branch rather than `master`

cyberixae:
My intention was not that people would use these functions, only to provide an index for how to do the transformations. This was the minimum level of documentation that I would have needed to survive the situation that I encountered. There is probably a better way to explain the same thing. I'll perhaps revisit the topic after I have made the 2.0 transition in my code base.

Is the `TaskEither` constructor implemented somewhere already or is it something that could hypothetically be added if it was written in a more concise manner?

docs/modules/TaskEither.ts.md
c089: @cyberixae that sounds like a thing I have the skills to contribute something :) couldn't find an issue for it, is there any info on why that doesn't exist yet? just noone prioritized it, or anything in the setup blocking implementing it?

cyberixae: I don't know really. The documentation seems to use GitHub Pages for hosting. I wonder how it gets deployed. Compiling the documentation seems to do type checking for code examples, which probably needs to happen before the code is accepted into master.

cyberixae: I only know this because last time I did not generate the docs and that needed to be done separately afterwards. See https://github.com/gcanti/fp-ts/pull/854#issuecomment-487375171

PR #926, opened 2019-08-08
Make /lib in deep import paths optional

slikts:
Enables importing modules as `import { Either } from 'fp-ts/Either'` instead of `'fp-ts/lib/Either'` while staying backward compatible. It publishes from `/lib` instead of the package root, copies and patches `package.json` to `/lib`, and copies all generated source files to `/lib/lib` to keep backward compatibility.

The changes are intended to be minimal, but one caveat is that `npm-link` also needs to run from the `/lib` directory to work, and that changes in the copied files (`/lib/lib` and `/lib/package.json`) won't be updated by using `tsc` watch mode.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,7 +1,6 @@
 *.log
 node_modules
 lib
-es6
 dev
 .idea
 coverage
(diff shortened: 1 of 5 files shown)


gcanti:
Closing in favor of https://github.com/gcanti/fp-ts/issues/1053

PR #929, opened 2019-08-12
fp-ts-ixjs to ecosystem added

mlegenhausen:
I have created bindings for [IxJS](https://github.com/ReactiveX/IxJS) and would like to add my lib to the ecosystems page.

Btw IxJS is a great library when you want FP style nodejs streams with back pressure handling.


gcanti:
Thanks @mlegenhausen

PR #932, opened 2019-08-15
Fix outdated documentation in IO.ts

ksaaskil:
Thanks for the awesome library! The part on sequencing IOs is outdated as `sequence_` is not exported from `Foldable` anymore, so I replaced that with `array.sequence(io)`. I'm not sure if this was the correct way to do replace the example as `array.sequence(io)` also allows accessing the results, and the current docs are talking about discarding the results. To an FP newbie like me an example that combines the actions without dropping the results seems a bit more useful (one can always discard the results with `io.map(() => (})`).


gcanti:
Thanks @ksaaskil

PR #933, opened 2019-08-16
Fix code snippet in IO.ts

cateland:
Hello and thank you for the great library.

Browsing the documentation i found this snippet of code was not working anymore with fp-ts@2.x.x


gcanti:
Thanks @cateland

Issue #925, opened 2019-08-06
Compositions & pipe


pbadenski:
I looked in both documentation and issues, but couldn't find any information on this...

Is it possible (right now or planned as a feature) to make composition "pipe-friendly" with a simple syntax?

Currently *Composition signatures cannot be nicely piped without introducing a function wrapper. For example in case of a OptionT monad transformer below:

```javascript
const optionArray = optionT.getOptionM(array.array);

pipe(
  optionArray.of(1),
  m => optionArray.map(m, v => v + 1)
  // how can you make below possible?
  // optionArray.map(v => v + 1)
)
```

gcanti:
```ts
import * as array from 'fp-ts/lib/Array'
import { Monad1 } from 'fp-ts/lib/Monad'
import { Option, some, none } from 'fp-ts/lib/Option'
import * as optionT from 'fp-ts/lib/OptionT'
import { pipe, pipeable } from 'fp-ts/lib/pipeable'

// a Monad instance for `Array<Option<A>>`...
declare module 'fp-ts/lib/HKT' {
  interface URItoKind<A> {
    ArrayOption: Array<Option<A>>
  }
}

const arrayOption: Monad1<'ArrayOption'> = {
  URI: 'ArrayOption',
  ...optionT.getOptionM(array.array)
}

// ...getting pipeable functions...
const { ap, apFirst, apSecond, chain, chainFirst, flatten, map } = pipeable(arrayOption)

// ...usage
pipe(
  arrayOption.of(1),
  map(v => v + 1),
  chain(n => (n > 0 ? [some(n + 1), some(n - 1)] : [none]))
)
```

pbadenski:
Thanks! I was worried that was the answer.. Out of curiosity do you think there's TypeScript features coming any time soon that could make it less verbose?

raveclassic:
@gcanti Shouldn't we reconsider changing typeclasses signatures to be always data-last? I have exactly the same problem right now trying to write tests for new `RemoteDataT` transformer and it's quite weird for tests to pollute the source with globally registered monad instance.

raveclassic:
Just played a bit, looks nice:

```typescript
import { HKT, Kind, URIS } from 'fp-ts/lib/HKT';
import { pipe } from 'fp-ts/lib/pipeable';

///////
export interface Functor<F> {
	readonly URI: F;
	readonly map: <A, B>(f: (a: A) => B) => (fa: HKT<F, A>) => HKT<F, B>;
}

[...]

PR #942, opened 2019-08-22
refactor: make Functor, Apply, FunctorComposition and ApplicativeComposition data-last

raveclassic:
So following https://github.com/gcanti/fp-ts/issues/925#issuecomment-523546371 this PR is a POC of a data-last `Functor`, `Apply`, `FunctorComposition` and `ApplicativeComposition` for seamless usage with `pipe`. 
Monadic data structures as well as their monad transformers now implement data-last typeclasses.

Previously (in the time of `fp-ts@0.x`) we didn't have `pipe` and `TS` couldn't infer the type of curried data-last argument. Now `pipe` allows to start the chain from that argument so that the types are correctly inferred. 

As a side effect this PR is to show that `pipeable` can be made redundant when (if) all typeclasses migrate to data-last

### DIFF
--- a/docs/modules/Applicative.ts.md
+++ b/docs/modules/Applicative.ts.md
@@ -126,7 +126,7 @@ Added in v2.0.0
 ```ts
 export interface ApplicativeComposition<F, G> extends FunctorComposition<F, G> {
   readonly of: <A>(a: A) => HKT<F, HKT<G, A>>
-  readonly ap: <A, B>(fgab: HKT<F, HKT<G, (a: A) => B>>, fga: HKT<F, HKT<G, A>>) => HKT<F, HKT<G, B>>
+  readonly ap: <A, B>(fgab: HKT<F, HKT<G, (a: A) => B>>) => (fga: HKT<F, HKT<G, A>>) => HKT<F, HKT<G, B>>
 }
 ```
 
@@ -139,7 +139,7 @@ Added in v2.0.0
 ```ts
 export interface ApplicativeComposition11<F extends URIS, G extends URIS> extends FunctorComposition11<F, G> {
   readonly of: <A>(a: A) => Kind<F, Kind<G, A>>
-  readonly ap: <A, B>(fgab: Kind<F, Kind<G, (a: A) => B>>, fga: Kind<F, Kind<G, A>>) => Kind<F, Kind<G, B>>
+  readonly ap: <A, B>(fgab: Kind<F, Kind<G, (a: A) => B>>) => (fga: Kind<F, Kind<G, A>>) => Kind<F, Kind<G, B>>
 }
 ```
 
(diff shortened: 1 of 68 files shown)


raveclassic:
@gcanti Do you have any thoughts on this? This PR _is_ a breaking change but it ultimately fixes some of the existing annoying issues with fp-ts:
  - API inconsistency:  `pipe` is an idiomatic way to use `fp-ts` however its abstract core - the typeclasses - have signatures not suitable for this idiomatic way out of the box and require some messy (redundant IMO) steps like calling `pipeable`, destructuring, reexporting and so on. When they actually could be used _as is_.
  - transformers: if I need a "pipeable" version (an idiomatic version actually) of some transformer I need to register its type as a separate monad on the `HKT` dictionary. This doesn't  make much sense because the type is already "bundled" in the transformer instance (for example the result of `getOptionM` already contains info about `Option` and passed `M`.
  - `pipeable` module itself is a mess with no sensible profit - it just fixes data-first typeclasses when they could be data-last by design completely eliminating the need for this module. 

PS. This reminds me https://github.com/gcanti/fp-ts/pull/202 actually :)

gcanti:
> This PR is a breaking change

@raveclassic Indeed, sorry but I'm afraid I can't accept this PR.

raveclassic:
Thanks for such an informative answer on the main topic of this _POC_.

giogonzo:
@raveclassic just adding my 2 cents here: the proposed API change has some advantages for sure (uniformity of the API and import/exports mainly) but, if I understand it correctly, at the cost of:
- dropping completely the support for e.g. `either.map(e, fn)` (which is something I find myself using many times when only a few operations are needed) (and as you already know this can't be achieved in a data-last fashion because of how TS inference works)
- being a breaking change: after all the effort and discussion spent on drafting and then releasing v2, and people still in the middle of upgrades on their projects, this library just can't have another major breaking change right now

raveclassic:
@giogonzo This is just a POC for a discussion. I'm sorry for not being more clear with that in the description. 
Thanks a lot for the discussion, regarding your points:
  - yep, but that's the minimisation of API surface. All the time I find myself explaining new people coming to our projects which way exactly to use in which situation.  I'd would be better to stick to _one idiomatic_ way once and for all - `pipe(e, map(f))`, TS even forces you to stick to it being unable to infer the type with direct application.
 - yeah, that's sad, really sad. I'm the one of those in the middle of upgrades. But that doesn't mean I suggest breaking the whole library once again right now. I'm just searching for a proper solution to the problems my teams face every day. That could be a separate package, or a subpackage, IDK

giogonzo:
> It would be better to stick to one idiomatic way once and for all

this could be achieved in steps and:
1. push on `pipe` in terms of documentation and examples for the time being
2. reevaluate in some time how people are using `pipe` vs data-first methods and, if it is the case, re-propose to drop support for data-first, passing through minor versions & deprecations. But again, this just can't happen right away.

> That could be a separate package, or a subpackage, IDK

Do you think I'd be possible to re-export fp-ts from e.g. an internal library of yours and "hide" data-first methods? That could lead to some good data points to contribute to this discussion

raveclassic:
> this could be achieved in steps

Agree. Long time ago there were `Fantasy<name>` typeclasses in `fp-ts` which followed `fantasy-land` however they were not usable with static approach and therefore were dropped. One of possible solutions would be to add `static-land` interfaces and update data structures with proper instances for them. This will however massively bloat API.

> re-export fp-ts

This will end in rewriting the major part of `fp-ts` including typeclasses and instances as well (some of implementations could be picked from pipeable output, but only non-constrained ones). That's why I opened this PR for discussion - maybe there's a better way that I don't see. Creating a maintained wrapper around `fp-ts` or even a replacement which follows `static-land` as close as possible _could_ be a way however we'd go that way in the worst case.

PR #944, opened 2019-08-24
Avoid IO void array in IO.ts example code

cyberixae:


gcanti:
Thanks @cyberixae but looks like the intention here was to specifically show `sequence`'s behavior .

Here's the commit https://github.com/gcanti/fp-ts/commit/3b6ab152b2b090f2635a9385ede8ce6d2330fc2c

/cc @ksaaskil

(Closing for now)

PR #945, opened 2019-08-29
getOrElse returns the right value or result of orElse

jdolle:
Minor change to `getOrElse` for `Either` and `Option` to allow the `onLeft/onNone` methods to return an alternate type

### DIFF
--- a/docs/modules/Either.ts.md
+++ b/docs/modules/Either.ts.md
@@ -268,7 +268,7 @@ Added in v2.0.0
 **Signature**
 
 ```ts
-export function getOrElse<E, A>(onLeft: (e: E) => A): (ma: Either<E, A>) => A { ... }
+export function getOrElse<E, A, B = A>(onLeft: (e: E) => B): (ma: Either<E, A>) => A | B { ... }
 ```
 
 Added in v2.0.0
--- a/docs/modules/Option.ts.md
+++ b/docs/modules/Option.ts.md
@@ -493,7 +493,7 @@ Extracts the value out of the structure, if it exists. Otherwise returns the giv
 **Signature**
 
 ```ts
-export function getOrElse<A>(onNone: () => A): (ma: Option<A>) => A { ... }
+export function getOrElse<A, B = A>(onNone: () => B): (ma: Option<A>) => A | B { ... }
 ```
 
 **Example**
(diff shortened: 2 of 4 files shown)


giogonzo:
Hi @jdolle , from my understanding this would be out of scope / undesired API since `getOrElse` is a just a shortcut for `fold(fn, identity)` and `fold` should mimic a pattern match on the sum type, where you _want_ to return the same type for all the cases.

If you _really_ want to return a union of different types, you can already manually unify the fold return type as follows:
```ts
pipe(
  some("str"),
  getOrElse<boolean | string>(constTrue)
);
```

jdolle:
@giogonzo Good point -- I hadn't thought about it like that. I'm pretty new to functional programming.

I was trying to resolve a left/none in `getOrElse` _without_ throwing an exception. i.e.
```
pipe(
  some("str"),
  getOrElse(() => { throw new NoValueError() }), // this is legal since the or else returns `never`, but feels bad and requires try/catch
)
```
But I (correct me if I am wrong) I should be using `fold` to create an `either` instead, i.e.
```
pipe(
  some("str"),
  option.fold(
    () => left(new Error("This is broken")),
    (r) => right(r),
  ),
)
```

giogonzo:
@jdolle In general, any function dealing with a failure (both `Option` and `Either` can be seen as representing some error condition), will either know how to recover from that failure (i.e. provide a default value) or not (and thus propagate the error)

> I was trying to resolve a left/none in getOrElse without throwing an exception

Correct, throwing makes it all pointless. If you know how to "resolve" the failure, it means you know how to create a value of type `A` (A in `Option<A>` or `Either<L, A>`), so my snippet in the previous comment should do it (except for the explicit type param in `getOrElse<A | B>` which is needed only if you want to end up with a value of type `A | B`, not the common case)

> But I (correct me if I am wrong) I should be using fold to create an either instead, i.e.

If your function doesn't know how to obtain an `A`, propagating the error is the only way to go and doing it using `Either` is definitely an option, one that I'd consider if the function in question is used together with other code working on `Either` (in this case also have a look at [`fromOption`](https://gcanti.github.io/fp-ts/modules/Either.ts.html#fromoption-export)) - otherwise you can just carry on with the `Option`

> I'm pretty new to functional programming.

me too, so please take this with a grain of salt 😅

Issue #947, opened 2019-09-10
Encoding an UpStar & DownStar


YBogomolov:
I'm trying to encode UpStar and DownStar in order to define profunctor instances for them, and I'm having a hard time. Right now I was able to do this:

```ts
export type UpStar<F, A, B> = (a: A) => HKT<F, B>;
export type UpStar1<F extends URIS, A, B> = (a: A) => Kind<F, B>;

export const URI = 'UpStar';
export type URI = typeof URI;

declare module 'fp-ts/lib/HKT' {
  interface URItoKind3<R, E, A> {
    UpStar: UpStar<R, E, A>;
  }
}

export function getUpStarPF<F extends URIS>(F: Functor1<F>): Profunctor3<URI> {
  return {
    URI,
    // Argument of type 'HKT<R, A>' is not assignable to parameter of type 'Kind<F, A>'.
    //                              ↓
    map: (era, ab) => (e) => F.map(era(e), ab),
    //                           and here: ↓
    promap: (era, de, ab) => (d) => F.map(era(de(d)), ab),
  };
}
```

The problem is that UpStar needs a `Functor<F>` instance in order to define a profunctor. `F` should have kind `* → *`, but UpStar itself should be `* → * → * → *`, if I understand everything correctly. I need somehow to specify that I would like to return an instance for `Profunctor<UpStar<F>>`.

@gcanti what do you think, is it possible to define such construct in TypeScript at all?

giogonzo:
@YBogomolov I spent a few minutes trying to grasp this, without much success. I'll just add my 2 cents here

Reading up at https://hackage.haskell.org/package/profunctors-3.3.0.1/docs/src/Data-Profunctor.html#UpStar, you are trying to encode e.g.
```hs
------------------------------------------------------------------------------
-- UpStar
------------------------------------------------------------------------------

-- | Lift a 'Functor' into a 'Profunctor' (forwards).
newtype UpStar f d c = UpStar { runUpStar :: d -> f c }

instance Functor f => Profunctor (UpStar f) where
  dimap ab cd (UpStar bfc) = UpStar (fmap cd . bfc . ab)
```
right?

I don't really get what's the actual meaning/usage of the tool, but speaking in fp-ts encoding:
`UpStar` is a newtype (not a typeclass), so I would skip the URI/HKT dance and encode it as:
```ts
import { HKT, URIS2, Kind2 } from "fp-ts/lib/HKT";

export interface UpStar<F> {
[...]

PR #950, opened 2019-09-12
Add curried Profunctor, Strong and Choice

YBogomolov:
Also add documentation about this technique.

Please see this issue (https://github.com/gcanti/fp-ts/issues/947) for details.

Also, this recipe will describe the approach: https://github.com/gcanti/fp-ts/blob/a4b7d3495ba7bf2ac36b6dc0995df298624c8dde/docs/recipes/profunctors.md

### DIFF
--- a/docs/modules/Choice.ts.md
+++ b/docs/modules/Choice.ts.md
@@ -40,7 +40,10 @@ Adapted from https://github.com/purescript/purescript-profunctor/blob/master/src
 - [Choice (interface)](#choice-interface)
 - [Choice2 (interface)](#choice2-interface)
 - [Choice3 (interface)](#choice3-interface)
+- [Choice3C (interface)](#choice3c-interface)
 - [Choice4 (interface)](#choice4-interface)
+- [Choice4C1 (interface)](#choice4c1-interface)
+- [Choice4C2 (interface)](#choice4c2-interface)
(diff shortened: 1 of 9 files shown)


YBogomolov:
Any feedback on this PR? Should I close it/improve/move to contrib/etc?

giogonzo:
@YBogomolov probably due to me missing fundamental concepts, I'm not sure how this addition would improve usage of the library. I went through https://github.com/gcanti/fp-ts/blob/a4b7d3495ba7bf2ac36b6dc0995df298624c8dde/docs/recipes/profunctors.md and I think I understand the technicalities, but it doesn't help to grasp the general picture.

Could you provide some examples of how this addition would allow for new/different techniques in fp-ts and how it would be better w.r.t. currently available alternatives? What problems in practical terms could you solve better with the tool?

grossbart:
I would be interested to have this functionality available. [This article](https://www.schoolofhaskell.com/school/to-infinity-and-beyond/pick-of-the-week/profunctors) got me interested in profunctors, but when I tried to use them, I didn't know where to start, so they're not useful to me _now_. But my hope is that with understanding I would be able to do something useful with them 😅 And then having them available would be great.

I don't have the technical understanding to comment on the technical implementation here, so I can only show my interest …

PR #959, opened 2019-10-01
add constrained Filterable instance to IOEither, TaskEither

giogonzo:
A few doubts:
- ~I'm not sure the tests add any value (if yes, I will add them for other functions too)~ EDIT: left a single one just to obtain the necessary jest coverage
- ~I wanted to add `getFilterable` to `Either` but then stopped and just reused `getWitherable` in order to avoid duplicating part of the logic. If you feel it would add clarity, even at the cost of some duplication, I will split them~ EDIT: discussed offline with @gcanti , it should be already aligned to other similar constrained instances, where a new function has been added only if the constraint changes


gcanti:
Thanks @giogonzo

PR #960, opened 2019-10-02
fix a few leftover "Setoid" references

giogonzo:


docs/recipes/equality.md
```diff
@@ -116,13 +116,13 @@ eqUserId.equals({ userId: 1, name: 'Giulio' }, { userId: 2, name: 'Giulio' }) //
```
grossbart: It would be good to rename `setoidArrayOfPoints` further up (lines 91–96) to, say, `arrayOfPoints` so it's less confusing now that the Setoid references are gone.

giogonzo: good catch! I updated it to `eqArrayOfPoints`

giogonzo:
@grossbart RE `docs/ramda`: the doc says explicitly "The following table refers to fp-ts@1.x" (and, altough we have `Eq` in `v1.19`, 1.x branch docs still refer to `Setoid` everywhere). So I would keep it as is for the moment

grossbart:
Ah, that makes sense. It's not ideal, but a different task. I'd say: merge this :)

PR #962, opened 2019-10-04
bracket behavior to return use error if both use and release fail

muradm:
Follow cats-effect in `bracket` behavior. [Docs](https://typelevel.org/cats-effect/datatypes/io.html#bracket):

> if the use action throws an error and then the release action throws an error as well, the reported error will be that of use, whereas the error thrown by release will just get logged (via System.err)


muradm:
closing in favor of #964

PR #968, opened 2019-10-11
Add endomorphic flow overload

vicrac:
Hi, I've added an overload to fix a problem I've run into when I started adding many endomorphic (i.e. `A => A`) steps to my data processing pipeline. It partially bypasses `flow`'s limitation to receive only up to 9 arguments.


GrzegorzKazana:
Seems like a great idea to me, however I think it's a good idea to add a test that is not using only functions of type `A => A`, so we are sure the new implementation still functions properly.

vicrac:
@gcanti Thanks for pointing this out, and sorry for bothering you :)

Issue #983, opened 2019-10-24
[Task] Maximum call stack size exceeded


mxth:
### Current Behavior
`array.sequence` a large number of `Task` (>5000 approx) would throw `RangeError`

### Expected behavior
Not to throw `RangeError`

### Reproducible example

```typescript
import { array, map, range } from 'fp-ts/lib/Array'
import { pipe } from 'fp-ts/lib/pipeable'
import { task } from 'fp-ts/lib/Task'

pipe(
  range(1, 6000),
  map(a => () => Promise.resolve(a)),
  array.sequence(task),
)()
```
would throw
```
 RangeError: Maximum call stack size exceeded

      at ../../node_modules/fp-ts/lib/Task.js:106:37
      at ma (../../node_modules/fp-ts/lib/Task.js:104:57)
      at mab (../../node_modules/fp-ts/lib/Task.js:106:71)
      at ma (../../node_modules/fp-ts/lib/Task.js:104:57)
      at mab (../../node_modules/fp-ts/lib/Task.js:106:71)
      at ma (../../node_modules/fp-ts/lib/Task.js:104:57)
      at mab (../../node_modules/fp-ts/lib/Task.js:106:71)
      at ma (../../node_modules/fp-ts/lib/Task.js:104:57)
      at mab (../../node_modules/fp-ts/lib/Task.js:106:71)
      at ma (../../node_modules/fp-ts/lib/Task.js:104:57)
```

### Your environment

Which versions of fp-ts are affected by this issue? Did this work in previous versions of fp-ts?

| Software   | Version(s) |
| ---------- | ---------- |
| fp-ts      |     2.0.5       |
| TypeScript |    3.5.3        |

giogonzo:
Changing the implementation of `task.ap` to:

```ts
ap: (mab, ma) => () => Promise.resolve().then(() => Promise.all([mab(), ma()]).then(([f, a]) => f(a)))
```

seems to "fix" this

Issue #907, opened 2019-07-13
Stack Safety


miguelferraro:
## 📖 Documentation

I'm trying to find more information about stack safety and fp-ts. It would be helpful if a section appeared in the documentation somewhere. In particular I'd like to find whether there is a trampolining implementation in the library?

colomboe:
I'm looking exactly for the same content right now... My specific case is about sequence/traverse. 

I'm trying to convert `Array<IO<Result>>` to `IO<Array<Result>>` but if the array length is too big I get a stack overflow error (on nodejs).

DenisFrezzato:
I had this problem too some time ago and I "solved" it just by chunking the array. `IO` right now is not stack safe and there is no trampolining implementation.

miguelferraro:
In general I think that, this is an important project, but needs to made more accessible to people less familiar with fp. Thanks for your response.

waynevanson:
This might help someone.

I discovered a functional style trampolining type class someone had written. It's not with fp-ts but it uses functional programming principles.

https://gist.github.com/tusharmath/f1eaa6e0bdb3cc2c37835497a85c0b60

PR #988, opened 2019-10-27
stack-safe IO

giogonzo:
A porting of https://github.com/purescript/purescript-effect/pull/12 by @safareli

related to #907 

### Pros
- makes operations on `IO` stack safe (currently I get a "Maximum call stack size exceeded" error with e.g. `array.sequence(io)(range(0, 10000).map(io.of))()`)
- non-breaking / doesn't change the underlying representation of an `IO<A>` (a thunk `() => A`)
- slightly better performance on a big number of operations on `IO`

### Cons
- private / obscure implementation details, makes the code harder to read
- slightly worse performance on small number of operations on `IO`

### Notes
- I didn't spend time making the implementation better in terms of type-safety / style (not so relevant since private anyway), but it can definitely be done (e.g. using a proper discriminated union for the operation type, etc.)
- Probably we can do something similar also for `Task` (see e.g. #983)?

### Benchmarks
<details>
 <summary>"big"</summary>

```ts
import * as Benchmark from 'benchmark'
import * as IO from '../src/IO'
import * as oldIO from '../src/old-IO'
import { array, range } from '../src/Array'

// old IO x 30.75 ops/sec ±0.81% (50 runs sampled)
// IO x 32.49 ops/sec ±0.61% (53 runs sampled)
// Fastest is IO

const arr = range(1, 6000)
suite
  .add('old IO', function() {
    array.sequence(oldIO.io)(arr.map(oldIO.of))()
  })
  .add('IO', function() {
    array.sequence(IO.io)(arr.map(IO.of))()
  })
  .on('cycle', function(event: any) {
    // tslint:disable-next-line: no-console
  })
  .on('complete', function(this: any) {
    // tslint:disable-next-line: no-console
  })
  .run({ async: true })
```
</details>

<details>
 <summary>"small"</summary>

```ts
import * as Benchmark from 'benchmark'
import * as IO from '../src/IO'
import * as oldIO from '../src/old-IO'
import { array, range } from '../src/Array'

// old IO x 884,610 ops/sec ±0.82% (89 runs sampled)
// IO x 712,194 ops/sec ±0.66% (89 runs sampled)
// Fastest is old IO

const arr = range(1, 10)
suite
  .add('old IO', function() {
    array.sequence(oldIO.io)(arr.map(oldIO.of))()
  })
  .add('IO', function() {
    array.sequence(IO.io)(arr.map(IO.of))()
  })
  .on('cycle', function(event: any) {
    // tslint:disable-next-line: no-console
  })
  .on('complete', function(this: any) {
    // tslint:disable-next-line: no-console
  })
  .run({ async: true })
```
</details>

### Opinions?

### DIFF
--- a/docs/modules/IO.ts.md
+++ b/docs/modules/IO.ts.md
@@ -184,7 +184,7 @@ Added in v2.0.0
 **Signature**
 
 ```ts
-export const of = <A>(a: A): IO<A> => () => ...
+export const of = <A>(a: A): IO<A> => ...
 ```
 
 Added in v2.0.0
--- a/src/IO.ts
+++ b/src/IO.ts
@@ -141,17 +141,17 @@ export function getMonoid<A>(M: Monoid<A>): Monoid<IO<A>> {
 /**
  * @since 2.0.0
  */
-export const of = <A>(a: A): IO<A> => () => a
+export const of = <A>(a: A): IO<A> => pureE(a)
 
 /**
  * @since 2.0.0
  */
 export const io: Monad1<URI> & MonadIO1<URI> = {
   URI,
-  map: (ma, f) => () => f(ma()),
+  map: mapE,
   of,
-  ap: (mab, ma) => () => mab()(ma()),
-  chain: (ma, f) => () => f(ma())(),
+  ap: applyE,
+  chain: bindE,
   fromIO: identity
 }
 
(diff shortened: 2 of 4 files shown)


safareli:
👍 Great to see this!

BTW using trivial implementation of traversable instance for array with other lazyish structures like Reader, State etc, will break with overflow when used on large array. it's possible to fix that for all such types like this https://github.com/sanctuary-js/sanctuary-type-classes/pull/19

giogonzo:
Interesting, that seems to be very similar to the implementation in https://github.com/purescript/purescript-foldable-traversable/blob/master/src/Data/Traversable.js

```ts
array.sequence(S.state)(range(0, N).map(S.state.of))(undefined)
```
Is indeed stack overflowing for me with N greater than 5000

safareli:
@giogonzo did you decide to fix Traversable instance instead?

giogonzo:
@safareli no, we didn't do nothing (that I'm aware of) in the end.

Even if stack-safety-related issues [come up](https://github.com/gcanti/fp-ts/issues/1177) from [time](https://github.com/gcanti/fp-ts/issues/983) to [time](https://github.com/gcanti/fp-ts/issues/907), it doesn't seem to be a high priority for fp-ts at the moment (and there are workarounds, i.e. batching operations)

Also, the community is expanding in terms of higher-level libraries, and e.g. https://github.com/Matechs-Garage/matechs-effect 's `Effect` is built on stack safe runtime (cc @mikearnaldi)

PR #991, opened 2019-10-29
Add init for NonEmptyArray

steida:


src/NonEmptyArray.ts
```diff
@@ -227,6 +227,21 @@ export function last<A>(nea: NonEmptyArray<A>): A {
...
+ *
+ * @example
+ * import { init } from 'fp-ts/lib/NonEmptyArray'
+ *
+ * assert.deepStrictEqual(init([1, 2]), [1]))
+ * assert.deepStrictEqual(init([1]), []))
+ *
+ * @since 2.0.0
+ */
+export function init<A>(as: Array<A>): Array<A> {
```
gcanti: This should be

```ts
export function init<A>(as: NonEmptyArray<A>): Array<A>
```

I guess

src/NonEmptyArray.ts
```diff
@@ -227,6 +227,21 @@ export function last<A>(nea: NonEmptyArray<A>): A {
   return nea[nea.length - 1]
 }
 
+/**
+ * Get all but the last element of an array, creating a new array.
+ *
+ * @example
```
gcanti: travis is failing because this example doesn't type check:

```ts
// too many parens ----------------------v
assert.deepStrictEqual(init([1, 2]), [1]))
// too many parens ------------------v
assert.deepStrictEqual(init([1]), []))
```

src/NonEmptyArray.ts
steida: Yep. Fixed.

src/NonEmptyArray.ts
steida: Fixed.

gcanti:
Thanks @steida

PR #997, opened 2019-11-04
add link to blog post about state monad

gcanti:
@pfgray I would love to link your blog post about the state monad, if it's okay with you (feel free to change / suggest the text)


pfgray:
@gcanti , sure, thanks!

Issue #963, opened 2019-10-04
bracket behavior


muradm:
According to documentation:

> Make sure that a resource is cleaned up in the event of an exception. The release action is called regardless of whether the body action throws or returns.

Consider the following playground for `TaskEither.bracket`:
```
import { constVoid } from 'fp-ts/lib/function';
import * as E from 'fp-ts/lib/Either';
import * as TE from 'fp-ts/lib/TaskEither';

describe('test', () => {

  type E = Error
  type A = string
  type B = number

  test.only('bracket', async () => {
    return await TE.bracket<E, A, B>(
      () =>
        new Promise<E.Either<E, A>>((resolve, reject) => {
          console.log('acquire connection...');
          resolve(E.right('connection'));
          // resolve(E.left(new Error('gracefully failed to acquire')));
          // reject(new Error('unexpectedly failed to acquire'));
        }),
      (res: A) => () =>
        new Promise<e.Either<E, B>>((resolve, reject) => {
          console.log('working on connection...');
          // resolve(E.right(res.length));
          // resolve(E.left(new Error('gracefully failed to compute')));
          reject(new E('unexpectedly failed to compute'));
        }),
      (res: A, result: E.Either<E, B>) => () =>
        new Promise<E.Either<E, void>>((resolve, reject) => {
          console.log('releasing connection...');
          resolve(E.right(constVoid()));
          // resolve(E.left(new Error('gracefully failed to release')));
          // reject(new Error('unexpectedly failed to release'));
        })
    )()
      .then(E.fold(
        err => console.log('graceful error:', err.message),
        result => console.log('result', result),
      ))
      .catch(err => console.log('unexpected error:', err.message));
  });
});
```
### Current Behavior

1. When `use` promise rejects, `release` is not called. However, when `use` resolves to `Left<E>` `release` is called. Of course one would say that `E` might not be of type `Error`, so promise rejected with some unexpected error which is not necessarily will match to second argument `result` of release part. Does that mean that user has to make sure that final `catch` to promise is handled? If so then documentation is misleading by saying `throws or returns`.

[...]

PR #1001, opened 2019-11-07
bug(TaskEither): adding try/catch safety to TaskEither.tryCatch

CygnusRoboticus:
`TaskEither.tryCatch` only handles promise failures, if the predicate fails it does so silently. This is inline with `Option/IOEither/Either.tryCatch`.

### DIFF
--- a/src/TaskEither.ts
+++ b/src/TaskEither.ts
@@ -140,7 +140,13 @@ export function getApplyMonoid<E, A>(M: Monoid<A>): Monoid<TaskEither<E, A>> {
  * @since 2.0.0
  */
 export function tryCatch<E, A>(f: Lazy<Promise<A>>, onRejected: (reason: unknown) => E): TaskEither<E, A> {
-  return () => f().then(a => E.right(a), reason => E.left(onRejected(reason)))
+  return () => {
+    try {
+      return f().then(a => E.right(a), reason => E.left(onRejected(reason)))
+    } catch (e) {
+      return Promise.resolve(E.left(onRejected(e)))
+    }
+  }
 }
 
 /**
--- a/test/TaskEither.ts
+++ b/test/TaskEither.ts
@@ -121,13 +121,6 @@ describe('TaskEither', () => {
     assert.deepStrictEqual(e, E.left(1))
   })
 
-  it('tryCatch', async () => {
-    const e1 = await _.tryCatch(() => Promise.resolve(1), () => 'error')()
-    assert.deepStrictEqual(e1, E.right(1))
-    const e2 = await _.tryCatch(() => Promise.reject(undefined), () => 'error')()
-    assert.deepStrictEqual(e2, E.left('error'))
-  })
-
   it('taskify', async () => {
     const api1 = (_path: string, callback: (err: Error | null | undefined, result?: string) => void): void => {
       callback(null, 'ok')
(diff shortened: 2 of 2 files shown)


giogonzo:
> TaskEither.tryCatch only handles promise failures

This is intended behavior, see my comment [here](https://github.com/gcanti/fp-ts/issues/963#issuecomment-538505872).

I'm not saying it is obvious at a first glance, and could possibly be better documented: given the name `tryCatch`, non-FP people will likely assume `try {} catch {}` is included

CygnusRoboticus:
I understand the adherence to FP, but I don't think swallowing an error from consumer code, necessitating drilling down into `fp-ts` internals to discover what the issue is, is good behavior.

Is the recommended usage to simply try/catch manually inside of `TaskEither.tryCatch`?
```
TaskEither.tryCatch(() => {
  try {
    const value = (null as any).errory_code;
    return Promise.resolve(value);
  } catch (e) {
    return Promise.reject(e);
  },
  ...
);
```

giogonzo:
> Is the recommended usage to simply try/catch manually inside of TaskEither.tryCatch?

Are you using an external library function that both returns a (possibly rejected) promise and can throw?

Otherwise I would suggest to use `Either.tryCatch` to wrap the possibly throwing code and then `TaskEither.tryCatch` to wrap the possibly promise-rejecting one, as in:

```ts
pipe(
  E.tryCatch(possiblyThrowingError, onError),
  TE.fromEither,
  TE.chain(v => TE.tryCatch(possiblyRejectingPromise(v), onError))
)
```

Note: for the specific example provided above, I also suggest to have a look at `Either/Option.fromNullable`

anilanar:
@giogonzo @gcanti

I vote for /reopen and do a minor breaking change here:

> fp-ts expects to work with pure function implementations - always. 

Not always. There's why `E.tryCatch` exists. `E.tryCatch` specifically takes a non-pure function. It's an escape hatch for effectful functions which are "believed" to have no side effects except for `throw`ing errors. They are `unsafeCoerce`, but more specialized for a specific use case. Why take an unpractical stance for unprincipled lawless escape hatches? 

Moreover, maybe we can make it more explicit by replacing `Lazy` with `IO` for the following. They represent the same type from the compiler's perspective but (perhaps?) have different meanings from a developer's perspective (Lazy is pure, IO is not?):

- `TE.tryCatch` to `<E, A>(io: IO.IO<Promise<A>>, onErr: (err: unknown) => E): TE.TaskEither<E, A>` 
- `IOE.tryCatch` to `<E, A>(io: IO.IO<A>, onErr: (err: unknown) => E): IOE.IOEither<E, A>` 

And change `TE.tryCatch` to catch synchronous errors too. It's still possible to catch synchronous errors separately by doing something like:

```
pipe(
  IOE.tryCatch(
    () => fetch('/hello'),
    () => 'could not fetch sync',
  ),
  TE.fromIOEither,
  TE.map(IO.of),
  TE.chain(eff => TE.tryCatch(eff, () => 'could not fetch async')),
);
```

mlegenhausen:
IMHO then we could write it even further and write `Task<A>` instead of `IO<Promise<A>>`.

Couldn't we just simply write `TE.tryCatch(async () => await fetch('/hello'), E.toError)` to get the try/catch?

gcanti:
Since the purpose of `tryCatch` is **interop with unsafe code** (and moreover the current behaviour seems a source of confusion for many users) I think that it make  sense to catch both (as per @CygnusRoboticus's suggestion)

OliverJAsh:
@gcanti In that case should we re-open this PR?

PR #1017, opened 2019-11-15
Add Promise comparison table (closes #1016)

### LINKED ISSUE #1016: Promise.race behavior
## 📖 Documentation

I asked some days before about `Promise.allSettled`, but I have the same question for `Promise.race`. 

Let's say I have an array of
```typescript
declare const myTEs: TE.TaskEither<E,A>[];
```

How would one implement a `Promise.race`? It should return `TE.TaskEither<E,A>`.

I think this example could go on this page: https://gcanti.github.io/fp-ts/recipes/async.html

giogonzo:


grossbart:
This is a great idea!

gcanti:
Thanks @giogonzo

PR #1026, opened 2019-11-24
Add list

DenisFrezzato:


src/LinkedList.ts
```diff
@@ -0,0 +1,606 @@
...
+  return reverse(out)
+}
+
+/**
+ * Sort the elements of a list in increasing order, where elements
+ * are compared using the specified ordering.
+ * @since 2.1.1
+ */
+export function sort<A>(O: Ord<A>): (fa: LinkedList<A>) => LinkedList<A> {
+  return flow(toArray, A.sort(O), fromArray)
```
DenisFrezzato: This thing walks the array twice (other than sort it). Perhaps makes sense to implement a custom algorithm?

raveclassic:
Hi, @DenisFrezzato! 
Thanks for the PR  but I'd like to know what is the motivation behind introducing this module into the core? What are the advantages of using `List` in comparison to `Array` and what new features are available in `List` also comparing to `Array`.

gcanti:
@DenisFrezzato @raveclassic personally I'm interested in a functional data structure like `List`: sometimes I need a cheap way to build a list by `cons`-ing elements, for example in [hyper-ts](https://github.com/gcanti/hyper-ts/blob/77fd886c1d54f986bd0b546096eaffd2a24c4041/src/express.ts#L82).

However I'm not interested in feature parity with `Array`, I'd just add those operations that are already optimised (or on par with `Array`) by the very nature of `List`, like

- `nil`
- `cons`
- `singleton`
- `length`
- `isNil`
- `isCons`
- `head`
- `tail`
- `foldLeft`
- `findIndex`
- `findLastIndex`
- `dropLeft`
- `dropLeftWhile`

instances like

- `Foldable`
- `Traversable`

and a couple of interop functions like

- `toArray`
- `toReversedArray`
- `fromArray`

raveclassic:
Then what would an idiomatic fp-ts way be?  When to use `List` and when `Array`? Wouldn't it lead to confusion?

sledorze:
@raveclassic different data structures, with different behaviour (persistence, etc.) the choice is more related to the problem at hand.

raveclassic:
Mmmm what about `NonEmptyList`? Should it also be added? I'm just trying to understand what benefits will the library gain from this new structure which will outweigh the increase of API surface, increase of documentation and maintenance cost.

sledorze:
@raveclassic sometimes, for example, while traversing a tree, you may like to keep the current path, and for that, an array is not optimal (array recreation can cost a lot) whereas a List in that case is very useful.
I think it's part of the data-structure often use.
But I understand your concerns and it may feel a bit fp-ts-contrib level.
Should the question be stated otherwise? for instance, would fp-ts itself benefit from a List type as of now?

raveclassic:
@sledorze I see, thanks. So it's a matter of performance, right? AFAIK performance is not a _primary_ goal for `fp-ts` so, keeping that in mind, similar solution could be implemented using existing `NonEmptyArray`.

> Should the question be stated otherwise? for instance, would fp-ts itself benefit from a List type as of now?

Indeed, sorry for confusion. Are there any places in `fp-ts` core that could be improved by using `List`? If not, I'd prefer to keep the core clean and concise and to leave the choice for the end user, wether to use minimal working solution based on the core or to use separate optimised structure from a separate package.

gcanti:
A quick benchmark

```ts
import * as Benchmark from 'benchmark'
import * as A from '../src/Array'

export interface Nil {
  readonly type: 'Nil'
}

export interface Cons<A> {
  readonly type: 'Cons'
  readonly head: A
  readonly tail: List<A>
}

export type List<A> = Nil | Cons<A>

export const nil: List<never> = { type: 'Nil' }

export function cons<A>(head: A, tail: List<A>): List<A> {
  return { type: 'Cons', head, tail }
}

const iterations = 100

suite
  .add('mutable Array', function() {
    const as: Array<number> = []
    for (let i = 0; i < iterations; i++) {
      as.unshift(i)
    }
  })
  .add('immutable Array', function() {
    let as: Array<number> = []
    for (let i = 0; i < iterations; i++) {
      as = A.cons(i, as)
    }
  })
  .add('List', function() {
    let as: List<number> = nil
    for (let i = 0; i < iterations; i++) {
      as = cons(i, as)
    }
  })
  .on('cycle', function(event: any) {
    // tslint:disable-next-line: no-console
  })
  .on('complete', function(this: any) {
    // tslint:disable-next-line: no-console
  })
  .run({ async: true })
```

**Results**

```
mutable Array x 154,748 ops/sec ±0.48% (87 runs sampled)
immutable Array x 85,758 ops/sec ±0.26% (87 runs sampled)
List x 1,175,634 ops/sec ±0.35% (88 runs sampled)
```

I'm ok with moving this to `fp-ts-contrib`, @DenisFrezzato ?

DenisFrezzato:
It looks to me that `List` is rather a standard data structure, that's why I created the PR here and not in `fp-ts-contrib`. Anyway, if it's not welcome here, I don't have any problem to move it there, I'll be happy to contribute to `fp-ts` ecosystem anyway.

raveclassic:
Thank you guys for understanding. It gets really _hard_ when it comes to explaining things to newcomers in `fp-ts`.

PR #1043, opened 2019-12-11
version 2.3.0

gcanti:


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "fp-ts",
-  "version": "2.2.0",
+  "version": "2.3.0",
```
DenisFrezzato: I think that the version should be bumped also in `package-lock.json`.

Issue #1045, opened 2019-12-13
`sortBy` will fail in runtime if empty array is passed as `ords`


vicrac:
### Current Behavior
```
sortBy([])([1, 42, 13])
```
will pass type checks and fail in runtime, with something like
```
TypeError: O is undefined
```
### Expected behavior

In my opinion, it would be reasonable to return initial data "as-is", without changes (i.e. "no sorting applied"). Image e.g. a user choosing sorting criteria from a dropdown. Until nothing is selected, data remains unsorted.

I would also accept passing `NonEmptyArray` to `sortBy`, but it requires changing interface.

### Suggested solution(s)

I will open a PR solving it in a minute

### Your environment

| Software   | Version(s) |
| ---------- | ---------- |
| fp-ts      |      2.0.1      |
| TypeScript |      3.7.3      |

PR #1046, opened 2019-12-13
Fix sortBy failing on empty ords passed

vicrac:
Fixes https://github.com/gcanti/fp-ts/issues/1045

### DIFF
--- a/src/Array.ts
+++ b/src/Array.ts
@@ -10,7 +10,7 @@ import { Extend1 } from './Extend'
 import { FilterableWithIndex1 } from './FilterableWithIndex'
 import { Foldable1 } from './Foldable'
 import { FoldableWithIndex1 } from './FoldableWithIndex'
-import { Predicate, Refinement } from './function'
+import { Predicate, Refinement, identity } from './function'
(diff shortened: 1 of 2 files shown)


src/Array.ts
```diff
@@ -1105,7 +1105,7 @@ export function uniq<A>(E: Eq<A>): (as: Array<A>) => Array<A> {
  * @since 2.0.0
  */
 export function sortBy<A>(ords: Array<Ord<A>>): (as: Array<A>) => Array<A> {
```
raveclassic: I think it would be better to require `NonEmptyArray<Ord<A>>` instead of noop on empty array.

vicrac: Well, I would rather incline towards handling this case by this function rather than forcing user to provide `NonEmptyArray`. Besides, won't it break compatibility?

gcanti: > won't it break compatibility?

Yes it will, @raveclassic this fix LGTM

raveclassic: @gcanti ok then, lgtm too

gcanti:
Thanks @vicrac, patch released https://github.com/gcanti/fp-ts/releases/tag/2.3.1

PR #1051, opened 2019-12-14
Add `getMonoid` for Ord

vicrac:
Following https://github.com/gcanti/fp-ts/pull/1046, I think it would be semantically consistent to add a `Monoid<Ord>` instance, with `empty` returning an `Ord` instance which always considers compared elements equal (Js now guarantees to not to change order of elements in such a case - "no sorting applied" - sorting in js is now guaranteed to be stable). Besides consistency with `sortBy` implementation which can now take empty array of ords as input (and then performs "no sorting at all" - data remains untouched) it honestly... makes `sortBy` unnecessary (because we can now always build a specific `Ord` first, using `fold`), which leaves us with nice, terse and consistent interface for sorting, made of one function only - `sort` itself.

### DIFF
--- a/src/Ord.ts
+++ b/src/Ord.ts
@@ -16,6 +16,7 @@ import { Semigroup } from './Semigroup'
 import { Eq } from './Eq'
 import { Contravariant1 } from './Contravariant'
 import { pipeable } from './pipeable'
+import { Monoid } from './Monoid'
 
 declare module './HKT' {
   interface URItoKind<A> {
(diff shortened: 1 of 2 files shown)


src/Ord.ts
```diff
@@ -226,6 +227,16 @@ export function getSemigroup<A = never>(): Semigroup<Ord<A>> {
   }
 }
 
+/**
+ * @since 2.3.1
```
gcanti: Technically this is a new feature and will be released in v2.4.0

src/Ord.ts
```diff
@@ -226,6 +227,16 @@ export function getSemigroup<A = never>(): Semigroup<Ord<A>> {
...
+ */
+export function getMonoid<A>(): Monoid<Ord<A>> {
+  return {
+    empty: fromCompare((a1, a2) => 0),
```
gcanti: We can define `empty` once and for all

```ts
const empty = fromCompare(() => 0)

/**
 * @since 2.4.0
 */
export function getMonoid<A>(): Monoid<Ord<A>> {
  return {
    empty,
    concat: getSemigroup<A>().concat
  }
}
```

gcanti:
Thanks for this PR @vicrac

> Js now guarantees to not to change order of elements in such a case - "no sorting applied" - sorting in js is now guaranteed to be stable

Actually it doesn't matter, `fromCompare(() => 0)` is the monoidal unit whether sorting is stable or not

PR #1067, opened 2019-12-27
add runState to State and StateReaderTaskEither

willheslam:
Add `runState` function to State monad.

This is something I was hoping to use after my experience with Crocks which has the same function:
https://crocks.dev/docs/crocks/State.html#runwith

Maybe this is overcomplicated and I can just use `M.map(ma(s), ...)` directly, given the implementation of `eval/exec`?
```
    evalState: (ma, s) => M.map(ma(s), ([a]) => a),
    execState: (ma, s) => M.map(ma(s), ([_, s]) => s)
```

This should match the Haskell state monad equivalent, looking at
https://en.wikibooks.org/wiki/Haskell/Understanding_monads/State#Getting_Values_and_State
http://hackage.haskell.org/package/mtl-2.2.2/docs/Control-Monad-State-Lazy.html#v:runState

where it looks like `evalState` and `execState` are implemented using `runState`:
```
evalState :: State s a -> s -> a
evalState p s = fst (runState p s)

execState :: State s a -> s -> s
execState p s = snd (runState p s)
```

### DIFF
--- a/src/StateT.ts
+++ b/src/StateT.ts
@@ -26,6 +26,7 @@ export interface StateM<M> {
   readonly gets: <S, A>(f: (s: S) => A) => StateT<M, S, A>
   readonly fromState: <S, A>(fa: State<S, A>) => StateT<M, S, A>
   readonly fromM: <S, A>(ma: HKT<M, A>) => StateT<M, S, A>
+  readonly runState: <S, A>(ma: StateT<M, S, A>, s: S) => HKT<M, [A, S]>
   readonly evalState: <S, A>(ma: StateT<M, S, A>, s: S) => HKT<M, A>
   readonly execState: <S, A>(ma: StateT<M, S, A>, s: S) => HKT<M, S>
 }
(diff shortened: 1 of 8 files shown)


src/StateT.ts
```diff
@@ -130,6 +134,7 @@ export function getStateM<M>(M: Monad<M>): StateM<M> {
     gets: f => s => M.of([f(s), s]),
     fromState: sa => s => M.of(sa(s)),
     fromM: ma => s => M.map(ma, a => [a, s]),
+    runState: (ma, s) => M.map(ma(s), ([a, s]) => [a, s]),
```
willheslam: This could implemented instead as
```
runState: (ma, s) => M.map(ma(s), x => x),
```
or
```
import { identity } from '../src/function'
...
runState: (ma, s) => M.map(ma(s), identity),
```

Think that'd be better?

willheslam:
Ah... after looking into this a bit more I've realised `runState` is already implemented as a function/thunk that the State type represents:
https://gcanti.github.io/fp-ts/modules/State.ts.html#state-interface

i.e.
```
S.runState(S.of(42), 'foo') // [42, 'foo']
```
can be already achieved with
```
S.of(42)('foo') // [42, 'foo']
```

So I'll close this preemptively! How embarrassing. :)

PR #1071, opened 2019-12-31
Add fetcher-ts to ecosystem

YBogomolov:
I've created a `fetcher-ts` module – a type-safe HTTP client with `io-ts` validation support and ability to transform into a `TaskEither`. I would like to see it added to the ecosystem.


gcanti:
Thanks @YBogomolov

PR #1086, opened 2020-01-13
add readonly modules

gcanti:
This PR will add the following modules:

- `ReadonlyArray` (done)
- `ReadonlyNonEmptyArray` (done)
- `ReadonlySet` (done)
- `ReadonlyMap` (done)
- `ReadonlyRecord` (done)
- `ReadonlyTuple` (done)
- other?

The goal is to obtain drop-in replacements for the mutable counterparts (with some exceptions, see the notes below).

Notes:

- `ReadonlyArray`
  - didn't port `copy`, seems useless
- `ReadonlyNonEmptyArray`
  - didn't port `copy`, seems useless
  - renamed `fromArray` to `fromNonEmptyArray`
- `Semigroup`
  - changed `fold`'s signature in order to support readonly arrays
- `Monoid`
  - changed `fold`'s signature in order to support readonly arrays
- `ReadonlyMap`
  - renamed `toArray` to `toReadonlyArray`
- `ReadonlyRecord`
  - renamed `toArray` to `toReadonlyArray`
- `ReadonlySet`
  - renamed `toArray` to `toReadonlyArray`  
  - renamed `subset` to `isSubset`

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "fp-ts",
-  "version": "2.4.4",
+  "version": "2.5.0",
   "description": "Functional programming in TypeScript",
   "files": [
     "lib",
@@ -55,6 +55,7 @@
     "ts-node": "^8.0.2",
     "tslint": "5.11.0",
     "tslint-config-standard": "8.0.1",
+    "tslint-immutable": "^6.0.1",
     "typescript": "^3.7.2"
   },
   "tags": [
--- a/src/ReadonlyArray.ts
+++ b/src/ReadonlyArray.ts
@@ -1,0 +1,1678 @@
+/**
+ * @since 2.5.0
+ */
+import { Alternative1 } from './Alternative'
+import { Applicative } from './Applicative'
+import { Compactable1, Separated } from './Compactable'
+import { Either } from './Either'
+import { Eq } from './Eq'
+import { Extend1 } from './Extend'
+import { FilterableWithIndex1 } from './FilterableWithIndex'
+import { Foldable1 } from './Foldable'
+import { FoldableWithIndex1 } from './FoldableWithIndex'
+import { Predicate, Refinement, identity } from './function'
+import { FunctorWithIndex1 } from './FunctorWithIndex'
+import { HKT } from './HKT'
+import { Monad1 } from './Monad'
+import { Monoid } from './Monoid'
+import { ReadonlyNonEmptyArray } from './ReadonlyNonEmptyArray'
+import { isSome, none, Option, some } from './Option'
+import { fromCompare, getMonoid as getOrdMonoid, Ord, ordNumber } from './Ord'
+import { pipeable } from './pipeable'
+import { Show } from './Show'
+import { TraversableWithIndex1 } from './TraversableWithIndex'
+import { Unfoldable1 } from './Unfoldable'
+import { Witherable1 } from './Witherable'
+
+declare module './HKT' {
+  interface URItoKind<A> {
+    readonly ReadonlyArray: ReadonlyArray<A>
+  }
+}
+
+/**
+ * @since 2.5.0
+ */
+export const URI = 'ReadonlyArray'
+
+/**
+ * @since 2.5.0
+ */
+export type URI = typeof URI
+
+/**
+ * @since 2.5.0
+ */
+// tslint:disable-next-line: readonly-array
+export function fromArray<A>(as: Array<A>): ReadonlyArray<A> {
+  const l = as.length
+  if (l === 0) {
+    return empty
+  }
+  const ras = Array(l)
+  for (let i = 0; i < l; i++) {
+    ras[i] = as[i]
+  }
+  return ras
+}
+
+/**
+ * @since 2.5.0
+ */
+// tslint:disable-next-line: readonly-array
+export function toArray<A>(ras: ReadonlyArray<A>): Array<A> {
+  const l = ras.length
+  const as = Array(l)
+  for (let i = 0; i < l; i++) {
+    as[i] = ras[i]
+  }
+  return as
+}
+
+/**
+ * @since 2.5.0
+ */
+export function getShow<A>(S: Show<A>): Show<ReadonlyArray<A>> {
+  return {
+    show: as => `[${as.map(S.show).join(', ')}]`
+  }
+}
+
+const concat = <A>(x: ReadonlyArray<A>, y: ReadonlyArray<A>): ReadonlyArray<A> => {
+  const lenx = x.length
+  if (lenx === 0) {
+    return y
+  }
+  const leny = y.length
+  if (leny === 0) {
+    return x
+  }
+  const r = Array(lenx + leny)
+  for (let i = 0; i < lenx; i++) {
+    r[i] = x[i]
+  }
+  for (let i = 0; i < leny; i++) {
+    r[i + lenx] = y[i]
+  }
+  return r
+}
+
+/**
+ * Returns a `Monoid` for `ReadonlyArray<A>`
+ *
+ * @example
+ * import { getMonoid } from 'fp-ts/lib/ReadonlyArray'
+ *
+ * const M = getMonoid<number>()
+ * assert.deepStrictEqual(M.concat([1, 2], [3, 4]), [1, 2, 3, 4])
+ *
+ * @since 2.5.0
+ */
+export function getMonoid<A = never>(): Monoid<ReadonlyArray<A>> {
+  return {
+    concat,
+    empty
+  }
+}
+
+/**
+ * Derives an `Eq` over the `ReadonlyArray` of a given element type from the `Eq` of that type. The derived `Eq` defines two
+ * arrays as equal if all elements of both arrays are compared equal pairwise with the given `E`. In case of arrays of
+ * different lengths, the result is non equality.
+ *
+ * @example
+ * import { eqString } from 'fp-ts/lib/Eq'
+ * import { getEq } from 'fp-ts/lib/ReadonlyArray'
+ *
+ * const E = getEq(eqString)
+ * assert.strictEqual(E.equals(['a', 'b'], ['a', 'b']), true)
+ * assert.strictEqual(E.equals(['a'], []), false)
+ *
+ * @since 2.5.0
+ */
+export function getEq<A>(E: Eq<A>): Eq<ReadonlyArray<A>> {
+  return {
+    equals: (xs, ys) => xs === ys || (xs.length === ys.length && xs.every((x, i) => E.equals(x, ys[i])))
+  }
+}
+
+/**
+ * Derives an `Ord` over the `ReadonlyArray` of a given element type from the `Ord` of that type. The ordering between two such
+ * arrays is equal to: the first non equal comparison of each arrays elements taken pairwise in increasing order, in
+ * case of equality over all the pairwise elements; the longest array is considered the greatest, if both arrays have
+ * the same length, the result is equality.
+ *
+ * @example
+ * import { getOrd } from 'fp-ts/lib/ReadonlyArray'
+ * import { ordString } from 'fp-ts/lib/Ord'
+ *
+ * const O = getOrd(ordString)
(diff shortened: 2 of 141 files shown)


gcanti:
@raveclassic @gkamperis @steida @mlegenhausen @sutarmin @Wenquer @giogonzo 

(tagging you all because of the discussion in https://github.com/gcanti/fp-ts/pull/1082, hope you don't mind..)

I have some questions:

1) should we make `Tuple` readonly or should we add a new `ReadonlyTuple` module?

Example

```ts
// before
export function swap<A, S>(sa: [A, S]): [S, A] {
  return [snd(sa), fst(sa)]
}

// after
export function swap<A, S>(sa: readonly [A, S]): readonly [S, A] {
  return [snd(sa), fst(sa)]
}
```

2) what about `State` and `Writer`?

```ts
export interface State<S, A> {
  (s: S): [A, S] // <= add the `readonly` modifier?
}

export interface Writer<W, A> {
  (): [A, W] // <= add the `readonly` modifier?
}
```

3) what about `Strong` and `Choice`?

```ts
// tuples...tuples everywhere...

export interface Strong<F> extends Profunctor<F> {
  readonly first: <A, B, C>(pab: HKT2<F, A, B>) => HKT2<F, [A, C], [B, C]>
  readonly second: <A, B, C>(pab: HKT2<F, B, C>) => HKT2<F, [A, B], [A, C]>
}

export interface Choice<F> extends Profunctor<F> {
  readonly left: <A, B, C>(pab: HKT2<F, A, B>) => HKT2<F, Either<A, C>, Either<B, C>>
  readonly right: <A, B, C>(pbc: HKT2<F, B, C>) => HKT2<F, Either<A, B>, Either<A, C>>
}
```

Wenqer:
@gcanti I personally prefer for using read-only in all three cases, but looks like it would be a breaking change.
Since initial plan was to add new read-only modules and collect feedback before making them mandatory, looks like we should add a ReadonlyTuple module and leave all three cases as is.

sutarmin:
Another question that comes to my mind is should we add helpers to transform data between readonly and mutable data types? E.g. for `ReadonlyArray` it could be `fromArray` and `toArray` (or maybe more descriptive names). It requires shallow copying of data structures, but in a lot of other libraries `Array` is more commonly used than `ReadonlyArray`, so developers will need to implement such transformations anyway.

gcanti:
@sutarmin added

- ReadonlyArray.fromArray
- ReadonlyMap.fromMap
- ReadonlyNonEmptyArray.fromArray
- ReadonlyRecord.fromRecord
- ReadonlySet.fromSet

Not sure about `to*` functions though, do we need them too?

gcanti:
Should be ready, anyone willing to review?

docs/recipes/purescript.md
```diff
@@ -53,8 +53,14 @@ data Foo = Bar String | Baz Boolean
 TypeScript
 
 ```ts
-interface Bar { type: 'Bar'; value: string }
-interface Baz { type: 'Baz'; value: boolean }
+interface Bar {
+  type: 'Bar'
```
raveclassic: `readonly` here?

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "fp-ts",
-  "version": "2.4.1",
+  "version": "2.5.0",
```
raveclassic: Why do you bump version in this PR?

src/ReadonlyArray.ts
```diff
@@ -0,0 +1,1678 @@
...
+ * @since 2.5.0
+ */
+export type URI = typeof URI
+
+/**
+ * @since 2.5.0
+ */
+// tslint:disable-next-line: readonly-array
+export function fromArray<A>(as: Array<A>): ReadonlyArray<A> {
+  const l = as.length
```
raveclassic: why not `identity`?

src/ReadonlyArray.ts
```diff
@@ -0,0 +1,1678 @@
...
+    ras[i] = as[i]
+  }
+  return ras
+}
+
+/**
+ * @since 2.5.0
+ */
+// tslint:disable-next-line: readonly-array
+export function toArray<A>(ras: ReadonlyArray<A>): Array<A> {
```
raveclassic: faster than `slice`?

src/ReadonlyArray.ts
```diff
@@ -0,0 +1,1678 @@
...
+/**
+ * @since 2.5.0
+ */
+export function getShow<A>(S: Show<A>): Show<ReadonlyArray<A>> {
+  return {
+    show: as => `[${as.map(S.show).join(', ')}]`
+  }
+}
+
+const concat = <A>(x: ReadonlyArray<A>, y: ReadonlyArray<A>): ReadonlyArray<A> => {
```
raveclassic: can we reuse `Array` module?

raveclassic: same for other `Readonly*` modules

package.json
gcanti: mhh.. why not? new features -> bump version

src/ReadonlyArray.ts
gcanti: based on https://github.com/gcanti/fp-ts/blob/5684ae1919044267d2b5a706800fd528bf6afc51/perf/Array/copy.ts seems a bit faster

src/ReadonlyArray.ts
gcanti: it would bloat the bundle, I'd prefer to keep them separated

src/ReadonlyArray.ts
raveclassic: No, I mean implementation:
```typescript
export const fromArray: <A>(as: Array<A>) => ReadonlyArray<A> = identity
```
Even more, we can just drop this function because we can safely pass `Array` as argument to functions accepting `ReadonlyArray`

src/ReadonlyArray.ts
raveclassic: Do you mean broken tree-shaking?

src/ReadonlyArray.ts
raveclassic: ok

package.json
raveclassic: I'm just unused to such approach, ok :)

src/ReadonlyArray.ts
raveclassic: Maybe at least reusing `ReadonlyArray` in `Array` module?

src/ReadonlyArray.ts
sutarmin: @raveclassic, with such implementation readonly array could be mutated after initialization (like in @gcanti's example above). Readonly and mutable structures should not share the same references.

raveclassic: ahhh indeed

src/ReadonlyArray.ts
gcanti: Do you mean?

```ts
// Array.ts

import * as RA from './ReadonlyArray'

export const getShow: <A>(S: Show<A>) => Show<Array<A>> = RA.getShow

export const getMonoid: <A = never>() => Monoid<Array<A>> = RA.getMonoid as any

// etc...
```

raveclassic: Yeah, exactly, however it would be nice to avoid casts, is it possible to reuse `Array` from `ReadonlyArray` since `ReadonlyArray`? That's what I meant originally.

gcanti: Some casts are unavoidable, for example `getMonoid`, `makeBy`, `range`, `replicate`, `flatten`, etc... (actually in the majority of cases)

src/ReadonlyArray.ts
DenisFrezzato: This actually made me frowning, so I played with that perf. Here the results on my machine (Linux) with Node.js 12.8:

Same perf from the repo, meaning array with 3 elements:
```
copy (native) x 42,886,253 ops/sec ±2.95% (80 runs sampled)
copy (Array) x 35,017,908 ops/sec ±2.70% (80 runs sampled)
Fastest is copy (native)
```

Array with 20 elements:
```
copy (native) x 22,314,806 ops/sec ±1.86% (90 runs sampled)
copy (Array) x 16,915,433 ops/sec ±1.20% (86 runs sampled)
Fastest is copy (native)
```

Array with 100 elements:
```
copy (native) x 8,201,057 ops/sec ±0.53% (92 runs sampled)
copy (Array) x 4,548,024 ops/sec ±0.93% (91 runs sampled)
Fastest is copy (native)
```

Array with 1k elements:
```
copy (native) x 843,804 ops/sec ±0.53% (94 runs sampled)
copy (Array) x 438,358 ops/sec ±1.06% (92 runs sampled)
Fastest is copy (native)
```

I actually had some weird results from the case with the array with 3 elements, like this one:
```
copy (native) x 29,569,974 ops/sec ±2.14% (82 runs sampled)
copy (Array) x 36,031,090 ops/sec ±3.06% (86 runs sampled)
Fastest is copy (Array) 
```

But generally it seems that `.slice` tends to be as twice as faster then `copy`.

gcanti: @DenisFrezzato here's mine (OSX 10.9.6, **Node.js 10.15.3**) and using `range` to build the arrays

`range(1, 3)`

```
copy (native) x 32,922,319 ops/sec ±0.52% (90 runs sampled)
copy (Array) x 79,747,647 ops/sec ±0.60% (85 runs sampled)
Fastest is copy (Array)
```

`range(1, 20)`

```
copy (native) x 17,306,844 ops/sec ±1.28% (89 runs sampled)
copy (Array) x 22,161,454 ops/sec ±1.71% (81 runs sampled)
Fastest is copy (Array)
```

`range(1, 100)`

```
copy (native) x 6,891,346 ops/sec ±1.01% (89 runs sampled)
copy (Array) x 5,313,439 ops/sec ±1.05% (87 runs sampled)
Fastest is copy (native)
```

`range(1, 1000)`

```
copy (native) x 893,195 ops/sec ±0.88% (87 runs sampled)
copy (Array) x 623,234 ops/sec ±1.01% (82 runs sampled)
Fastest is copy (native)
```

gkamperis: apologies for asking this... but how do you actually run the benchmark suites in /perf?

gcanti: @gkamperis `ts-node perf/Array/copy.ts`

gkamperis: @gcanti thanks! here is what I see: using the default suite I get native to be faster.
If I replace it with range(1, 3) then Array is faster...

gcanti: @gkamperis yep, in my setup, running the default suite, Array is still faster but the numbers are different (don't know why, I'm not an expert)

```
copy (native) x 65,308,620 ops/sec ±1.26% (88 runs sampled)
copy (Array) x 79,410,327 ops/sec ±0.62% (89 runs sampled)
Fastest is copy (Array)
```

For reference, `lodash` does something similar https://github.com/lodash/lodash/blob/04ebca6c86deba0ff733847f6c11fd5265e1ce03/.internal/copyArray.js

gkamperis: @gcanti not an expert either...

Trying a few more combos it looks like arrays defined at compile time are treated quite differently.

So [1, ..., 49] is treated differently to [].push(1..49) 

elements: 49
copy (native) x 9,184,822 ops/sec ±3.23% (83 runs sampled)
copy (native) Lit x 121,546,767 ops/sec ±5.32% (86 runs sampled)
copy (Array) x 13,041,231 ops/sec ±1.37% (90 runs sampled)
copy (Array) Lit x 11,902,280 ops/sec ±4.37% (80 runs sampled)
Fastest is copy (native) Lit

the difference in perf is huge for arrays fully defined at compile time.

gcanti:
@raveclassic the PR is still marked with "Changes requested", is there something to do?

raveclassic:
@gcanti Sure, it's ready. Just slipped my mind, sorry.

gcanti:
@raveclassic no problem, thanks for reviewing

PR #1098, opened 2020-01-22
Fix examples in recipes

xkollar:
Small issues in examples how to express Haskell/PureScript ideas in TypeScript.

### DIFF
--- a/docs/recipes/purescript.md
+++ b/docs/recipes/purescript.md
@@ -113,7 +113,7 @@ TypeScript
 
 ```ts
 function maybe<A, B>(whenNone: () => B, whenSome: (a: A) => B, fa: Option<A>): B => {
-  switch (fa.type) {
+  switch (fa._tag) {
     case 'None' :
       return whenNone()
     case 'Some' :
--- a/docs/recipes/purescript.md
+++ b/docs/recipes/purescript.md
@@ -94,9 +94,9 @@ export interface Some<A> {
   readonly value: A
 }
 
-export const none: Option<never> = { _tag: 'None' }
+export const none: Option<never> = ({ _tag: 'None' })
 
-export const some = <A>(a: A): Option<A> => new Some(a)
+export const some = <A>(a: A): Option<A> => ({ _tag: 'Some', value: a })
 ```
 
 ## Pattern matching
--- a/docs/recipes/purescript.md
+++ b/docs/recipes/purescript.md
@@ -122,6 +122,18 @@ const maybe = <A, B>(whenNone: () => B, whenSome: (a: A) => B, fa: Option<A>): B
 }
 ```
 
+Here TypeScript also provides exhaustiveness check as the function would otherwise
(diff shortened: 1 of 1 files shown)


docs/recipes/purescript.md
```diff
@@ -122,6 +122,9 @@ function maybe<A, B>(whenNone: () => B, whenSome: (a: A) => B, fa: Option<A>): B
 }
 ```
 
+Here TypeScript also provides exhaustiveness check because otherwise function
+ould lacks ending return statement and return type does not include 'undefined'.
```
DenisFrezzato: I believe a letter is missing here, maybe you wanted to write "would lack"?

xkollar: You are indeed right. Give me a sec, going to fix that commit…

xkollar: I believe it should be fixed as of now.

docs/recipes/purescript.md
```diff
@@ -122,6 +122,9 @@ function maybe<A, B>(whenNone: () => B, whenSome: (a: A) => B, fa: Option<A>): B
 }
 ```
 
+Here TypeScript also provides exhaustiveness check because otherwise function
+would lacks ending return statement and return type does not include 'undefined'.
```
DenisFrezzato:
```suggestion
would lack ending return statement and return type does not include 'undefined'.
```

xkollar: Oh my, sorry, will fix. 🤦

While are we doing this, wold it make sense to mention that in case of having switch otherwise, exhaustiveness check can be done with unification with never with something like

```
function assertUnreachable(x: never): never {
    throw new Error("Didn't expect to get here");
}
```
?

raveclassic: This is https://github.com/gcanti/fp-ts/blob/083e8608b175804c6d5783c89a5002143b654256/src/function.ts#L274-L276

xkollar: Thank you, change included

xkollar:
I have noticed that there is a branch removing all non-technical documentation… does this PR then make any sense at all?

gcanti:
@xkollar yes it does, `purescript.md` is technical documentation, so thank you for your fix

xkollar:
Hmmm… I think I resolved all the "change requests", and marked them as such, but GitHub is sill showing "1 review requesting changes"… am I missing something obvious?

docs/recipes/purescript.md
```diff
@@ -122,6 +122,18 @@ function maybe<A, B>(whenNone: () => B, whenSome: (a: A) => B, fa: Option<A>): B
 }
 ```
 
+Here TypeScript also provides exhaustiveness check as the function would otherwise
+lack ending return statement and its return type does not include 'undefined'.
+
+Alternativelly you can provide a default case using `absurd` function
```
DenisFrezzato:
```suggestion
Alternatively you can provide a default case using `absurd` function
```

PR #1117, opened 2020-02-03
MonadTask4 added

mlegenhausen:
`MonadTask4` and all upper dependencies added. `MonadTask4` was then added to `StateReaderTaskEither`. I also added the missing exports for `fromPredicate` and `filterOrElse`.

### DIFF
--- a/src/MonadTask.ts
+++ b/src/MonadTask.ts
@@ -3,8 +3,8 @@
  *
  * @since 2.0.0
  */
-import { HKT, Kind, Kind2, Kind3, URIS, URIS2, URIS3 } from './HKT'
-import { MonadIO, MonadIO1, MonadIO2, MonadIO2C, MonadIO3, MonadIO3C } from './MonadIO'
+import { HKT, Kind, Kind2, Kind3, URIS, URIS2, URIS3, URIS4, Kind4 } from './HKT'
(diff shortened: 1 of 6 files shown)


src/MonadTask.ts
```diff
@@ -48,3 +48,17 @@ export interface MonadTask3<M extends URIS3> extends MonadIO3<M> {
...
+ * @since 2.4.4
+ */
+export interface MonadTask4<M extends URIS4> extends MonadIO4<M> {
+  readonly fromTask: <S, R, E, A>(fa: Task<A>) => Kind4<M, S, R, E, A>
+}
+
+/**
+ * @since 2.4.4
+ */
+export interface MonadTask4C<M extends URIS4, E> extends MonadIO4C<M, E> {
```
gcanti: `MonadTask4` is required for `StateReaderTaskEither` but what's the use case for `MonadTask4C`?

mlegenhausen: Currently any use case. Can only image if someone wants to implement `getStateReaderTaskValidation` but for this we would need a lot more than this interface. I will remove it for now cause it was more of a copy paste move.

gcanti:
Thanks @mlegenhausen, released.

PR #1123, opened 2020-02-05
Porting StateEither from morphic-ts and adding tests

haredev:
Adding StateEither to FP-TS, see https://github.com/sledorze/morphic-ts/issues/4
StateEither is used in morphic to parse JSON schemas


gcanti:
@haredev thanks for this PR, let's add it to `fp-ts-contrib` though

haredev:
alright, moving it there :)

PR #1125, opened 2020-02-05
IORef this binding problems solved

mlegenhausen:
I made all attributes `readonly` and solved a problem that occurs when you try to `pipe` the functions of an `IORef` instance. The problem is that the `this` context gets lost when using the methods directly.

```ts
const ref = new IORef('hello')

pipe('world', ref.write)() // This would throw: Uncaught TypeError: Cannot set property 'value' of undefined
```


mlegenhausen:
What do you mean with remove? There are still there with the same signature just written in another way?

raveclassic:
@mlegenhausen They are moved from the prototype to the instance. ES6/TS classes execute parent constructors while extending so that these new class properties get available on child instance. But technically it's a breaking change because the methods are missing on the prototype: `IORef.prorotype.write` is not available anymore.

mlegenhausen:
Ok thats correct. I will change to the `bind` solution. But just for curiosity are you using `IORef` by extending it or modifying the `prototype` in any way?

gcanti:
@mlegenhausen Personally no. But who knows what's happening out there?

Maybe we could add a test with `pipe` to avoid regressions.

mlegenhausen:
Done

PR #1128, opened 2020-02-08
Add alga-ts library

YBogomolov:
I've ported Haskell library `alga` to TypeScript, and used `fp-ts` for HKT to provide Monad & Alternative instances. I would be happy to see it as a part of awesome `fp-ts` ecosystem.


gcanti:
Thanks @YBogomolov

PR #1146, opened 2020-02-25
Update learning-resources.md

morishjs:
`Getting started with fp-ts: Reader` post on `dev.to` is really helpful, hence this should be updated!


gcanti:
Thanks @morishjs

PR #1148, opened 2020-02-27
Fix TaskEither getApplyMonoid test

pheis:
TaskEither's getApplyMonoid test was testing twice concatting empty with right.
This PR fixes it so that it concats once with empty and right and once with empty and left.

### DIFF
--- a/test/TaskEither.ts
+++ b/test/TaskEither.ts
@@ -225,8 +225,8 @@ describe('TaskEither', () => {
     })
 
     it('empty (left)', async () => {
-      const x = await M.concat(M.empty, _.right('a'))()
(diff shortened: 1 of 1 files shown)


gcanti:
@pheis I think that "left" here means "left unit", i.e. `empty` must be both 

- a "right unit" `concat(a, empty) = a`
- a "left unit" `concat(empty, a) = a`

We can **add** a test with the `left` constructor though, keeping the existing one with the `right` constructor

pheis:
I got confused by the naming : p. It makes more sense now. I close this PR as it was due to my misunderstanding.

PR #1151, opened 2020-02-28
Remove redundant URI in spreads

sandersn:
Detected using typescript@next, which has an error for this case. (microsoft/TypeScript#37063)


gcanti:
Thanks @sandersn

PR #1154, opened 2020-03-02
Added mapNullable to Either

psilospore:
I was using Option where I was using mapNullable a few times. I switched to Either for error reporting and noticed there wasn't a mapNullable so thought it might be useful to add.

---
This change is [Reviewable](https://reviewable.io/reviews/gcanti/fp-ts/1154)

### DIFF
--- a/src/Either.ts
+++ b/src/Either.ts
@@ -223,7 +223,7 @@ export function mapNullable<E, A, B>(
   leftOrNull: E,
   f: (a: A) => B | null | undefined
 ): (ma: Either<E, A>) => Either<E, B> {
-  return (ma: Either<E, A>) => (isLeft(ma) ? left<E, B>(leftOrNull) : fromNullable<E>(leftOrNull)(f(ma.right)))
+  return (ma: Either<E, A>) => (isLeft(ma) ? ma : fromNullable<E>(leftOrNull)(f(ma.right)))
 }
 
 /**
--- a/src/Either.ts
+++ b/src/Either.ts
@@ -214,6 +214,19 @@ export function getShow<E, A>(SE: Show<E>, SA: Show<A>): Show<Either<E, A>> {
 }
 
 /**
+ * This is `chain` + `fromNullable`, useful when working with Either.
+ * Similar to Option.mapNullable except with Either, instead of None the value will be Left(leftOrNull)
+ * TODO add to docs
+ * @since 2.5.3
+ */
+export function mapNullable<E, A, B>(
+  leftOrNull: E,
+  f: (a: A) => B | null | undefined
+): (ma: Either<E, A>) => Either<E, B> {
+  return (ma: Either<E, A>) => (isLeft(ma) ? left<E, B>(leftOrNull) : fromNullable<E>(leftOrNull)(f(ma.right)))
+}
+
+/**
  * @since 2.0.0
  */
 export function getEq<E, A>(EL: Eq<E>, EA: Eq<A>): Eq<Either<E, A>> {
--- a/docs/modules/Either.ts.md
+++ b/docs/modules/Either.ts.md
@@ -81,6 +81,7 @@ Added in v2.0.0
 - [left](#left)
 - [map](#map)
 - [mapLeft](#mapleft)
+- [mapNullable](#mapnullable)
 - [orElse](#orelse)
 - [parseJSON](#parsejson)
 - [reduce](#reduce)
@@ -584,6 +585,23 @@ Added in v2.0.0
 ```
 
 Added in v2.0.0
+
+# mapNullable
+
+This is `chain` + `fromNullable`, useful when working with Either.
+Similar to Option.mapNullable except with Either, instead of None the value will be Left(leftOrNull)
+TODO add to docs
+
+**Signature**
+
+```ts
+export function mapNullable<E, A, B>(
+  leftOrNull: E,
+  f: (a: A) => B | null | undefined
+): (ma: Either<E, A>) => Either<E, B> { ... }
+```
+
+Added in v2.5.3
 
 # orElse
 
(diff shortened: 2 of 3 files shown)


src/Either.ts
```diff
@@ -213,6 +213,19 @@ export function getShow<E, A>(SE: Show<E>, SA: Show<A>): Show<Either<E, A>> {
...
+ * This is `chain` + `fromNullable`, useful when working with Either.
+ * Similar to Option.mapNullable except with Either, instead of None the value will be Left(leftOrNull)
+ * TODO add to docs
+ * @since 2.5.3
+ */
+export function mapNullable<E, A, B>(
+  leftOrNull: E,
+  f: (a: A) => B | null | undefined
+): (ma: Either<E, A>) => Either<E, B> {
+  return (ma: Either<E, A>) => (isLeft(ma) ? left<E, B>(leftOrNull) : fromNullable<E>(leftOrNull)(f(ma.right)))
```
gcanti: Is there a reason why you are returning `left(leftOrNull)` when `ma` is a `Left`?

psilospore:
> Is there a reason why you are returning `left(leftOrNull)` when `ma` is a `Left`?

Nope! that was a mistake

gcanti:
`Option`'s `mapNullable` is pretty old... now that we have optional chaining, a `mapNullable` utility seems awkward:

this

```ts
assert.deepStrictEqual(
  pipe(
    _.fromNullable(err)(x1.a),
    _.mapNullable(err, x => x.b),
    _.mapNullable(err, x => x.c),
    _.mapNullable(err, x => x.d)
  ),
  _.left(err)
)
```

is equivalent to this

```ts
assert.deepStrictEqual(_.fromNullable(err)(x1.a?.b?.c?.d), _.left(err))
```

gkamperis:
Isn't mapNullable also a transformer to something completely different as well though? For example a string to a number...

psilospore:
In my use case I was using a function that might return null.

```
pipe(
    fromNullable(ERRORS.ERROR_REASON_1)(maybeNull('hi')),
    chain((foo: string) =>
      fromNullable(ERRORS.ERROR_REASON_2)(maybeNull2(foo)),
    ),
    fold(...)
```

Which could be

```
pipe(
    fromNullable(ERRORS.ERROR_REASON_1)(maybeNull('hi')),
    mapNullable((ERRORS.ERROR_REASON_2, (foo) => maybeNull2(foo)),
    fold(...)
```

gcanti:
@gkamperis you are right, `mapNullable` is basically `chainNullableK`. Following a recent convention we often have this triad:

- fromSomething
- fromSomethingK
- chainSomethingK

for example in the `TaskEither` module we have

- `fromEither`
- `fromEitherK`
- `chainEitherK`

So, by setting `Something = Nullable`, we can have

- `fromNullable`
- `fromNullableK`
- `chainNullableK` (<= this is `mapNullable` which we could deprecate)

for both `Option` and `Either`.

**Option**

```ts
import * as O from 'fp-ts/lib/Option'

// ---------------------------------------
// Option
// ---------------------------------------

// existing
const fromNullable = O.fromNullable

// missing
function fromNullableK<A extends ReadonlyArray<unknown>, B>(
  f: (...a: A) => B | null | undefined
): (...a: A) => O.Option<NonNullable<B>> {
  return (...a) => fromNullable(f(...a))
}

// existing but with a different name
const chainNullableK = O.mapNullable

// ---------------------------------------
// examples
// ---------------------------------------

import { pipe } from 'fp-ts/lib/pipeable'
import { flow } from 'fp-ts/lib/function'

export interface X {
  a?: {
    b?: {
      c: string
    }
  }
}

declare function maybeNull1(s: string): string | null
declare function maybeNull2(s: string): number | null
declare const x: X

//
// use case: optional fields
//

// before
pipe(
  fromNullable(x.a),
  O.mapNullable(a => a.b),
  O.mapNullable(b => b.c),
  O.mapNullable(maybeNull1)
)

// after
pipe(fromNullable(x.a?.b?.c), chainNullableK(maybeNull1))

//
// use case: lifting nullable functions
//

// before
pipe(fromNullable(maybeNull1('hi')), O.mapNullable(maybeNull2))

// after (point free)
flow(fromNullableK(maybeNull1), chainNullableK(maybeNull2))
```

**Either**

```ts
import * as E from 'fp-ts/lib/Either'

// ---------------------------------------
// Either
// ---------------------------------------

// existing
const fromNullable = E.fromNullable

// missing
function fromNullableK<E>(
  e: E
): <A extends ReadonlyArray<unknown>, B>(
  f: (...a: A) => B | null | undefined
) => (...a: A) => E.Either<E, NonNullable<B>> {
  const from = fromNullable(e)
  return f => (...a) => from(f(...a))
}

// missing
export function chainNullableK<E>(
  e: E
): <A, B>(f: (a: A) => B | null | undefined) => (ma: E.Either<E, A>) => E.Either<E, NonNullable<B>> {
  const from = fromNullableK(e)
  return f => E.chain(from(f))
}

// ---------------------------------------
// examples
// ---------------------------------------

import { pipe } from 'fp-ts/lib/pipeable'
import { flow } from 'fp-ts/lib/function'

export interface X {
  a?: {
    b?: {
      c: string
    }
  }
}

declare function maybeNull1(s: string): string | null
declare function maybeNull2(s: string): number | null
declare const x: X

//
// use case: optional fields
//

pipe(fromNullable('err1')(x.a?.b?.c), chainNullableK('err2')(maybeNull1))

//
// use case: lifting nullable functions
//

flow(fromNullableK('err1')(maybeNull1), chainNullableK('err2')(maybeNull2))
```

@gkamperis @psilospore what do you think?

waynevanson:
@gcanti Looks good.

Issue #1168, opened 2020-03-20
Supplying a non-Either value with a type assertion should cause isLeft and isRight to return false


leilapearson:
### Current Behavior

`isLeft` will return `undefined` when called like this:

`isLeft(('Left' as unknown) as Either<unknown, unknown>);`

`isRight` will return `true` when called like this:

`isRight(('Right' as unknown) as Either<unknown, unknown>);`

### Expected behavior

When the supplied object is not an `Either` (neither a `Left` nor a `Right`), both `isRight` and `isLeft` should return `false`.  

### Reproducible example

See above, and see tests in pull request to be submitted shortly.

### Suggested solution(s)

Will submit a pull request shortly.

### Additional context

Ran into this issue when trying to use `isLeft` and `isRight` as type guards for some custom Jest matchers I created for matching against `Either` values:
```
expect.toBeEither(received: unknown)
expect.toBeLeft(received: unknown)
expect.toBeRight(received: unknown)
```
etc...

### Your environment

Which versions of fp-ts are affected by this issue? 

| Software   | Version(s) |
| ---------- | ---------- |
| fp-ts      |     2.0.0+ (tested with master and 2.5.3)     |
| TypeScript  |    3.5+ (tested with 3.7.4)   |

The bug exists from the time that  `isLeft` and `isRight` were introduced in fp-ts 2.0.0.

leilapearson:
Should I be adding checks for `null` and `undefined` too? 

The current code will error at runtime if you pass a null or undefined value:

`TypeError: Cannot read property '_tag' of null`

`TypeError: Cannot read property '_tag' of undefined`

If it makes sense to reviewers to add these checks, then I can take a look at other `is*` functions too (such as `Option.isSome()` and `Option.isNone()` for example) and make that change for all of them in a separate pull request.

giogonzo:
Hi @leilapearson, `fp-ts` is meant to be used in TypeScript (and with a [`strict: true` tsconfig](https://github.com/gcanti/fp-ts#typescript-compatibility)). The code samples above are indeed not valid TS (well.. by casting, TS is basically turned off).

If for any reason you want to allow `unknown`s as input in a jest matcher, I think you should first make sure that that value is an `Either`, then run the type guard to check if it's `Right`.

Using `io-ts` (and if I got the intent of e.g. `toBeRight` correctly), it could look something like:

```ts
import * as t from 'io-ts'
import * as E from 'fp-ts/lib/Either'
[...]

PR #1169, opened 2020-03-20
Handle bad type assertion in isLeft and isRight

leilapearson:
Supplying a non-Either value with an "as Either<unknown, unknown>"
type assertion should cause both isLeft and isRight to return false.

See issue [#1168](https://github.com/gcanti/fp-ts/issues/1168)


leilapearson:
Closing per discussion in issue #1168

Issue #1103, opened 2020-01-24
Making the documentation more welcoming to beginners


tstelzer:
## 📖 Documentation

I'm always excited when talking about fp-ts and its extensive ecosystem, but I always cringe visibly when asked about its documentation. I can see (and applaud) attempts to make it easier to get on board!

The "core concepts" page puts this nicely:

>>> Don’t panic. If “higher order abstractions” and “type classes” sound like insanely complicated words to you, the first thing you should know is that you don’t have to understand all the intricate details of fp-ts in order to make good use of it. There is a learning curve, certainly, but don’t fear that you can’t learn to use fp-ts effectively, even as a beginner in functional programming. The trick is to start using the parts that are easier to understand and then gradually expand your knowledge.

Yes, that's the spirit! 

But.

For a minute, assume the position of a bright-eyed programmer, who's _just_ heard about functional programming and is excited to have found this extensive toolkit in typescript; fp-ts. Let's say they are self-taught. They started to get into this field, learning object-oriented programming. Or maybe they started building games, when they were young and are now getting into web development via JS. Or maybe they studied CS a couple of years ago, just getting by, without ever really grokking the fundamental concepts. Point being, they are a workman, not an academic. I would call myself a workman programmer.

Now they jump into the documentation, eager to learn about this thing. So they click on "getting started", and are blasted away by theory. A couple of examples from the very first pages in "getting started" and "functional design":

>>> In this article the term “combinator” refers to [...] A style of organizing libraries centered around the idea of combining things. Usual So I know that, ly there is some type T, some “primitive” values of type T, and some “combinators” which can combine values of type T in various ways to build up more complex values of type T

and

>>>The declaration may be read as a type A belongs to type class Eq if there is a function named equal of the appropriate type, defined on it
What about the instances?
A programmer can make any type A a member of a given type class C by using an instance declaration that defines implementations of all of C’s members for the particular type A.

or this last sentence on "getting started":

>>> Finally another useful way to build an Eq instance is the contramap combinator: given an instance of Eq for A and a function from B to A, we can derive an instance of Eq for B

[...]

PR #1171, opened 2020-03-21
Add a description and example for Option.alt

ruizb:
Hello, based on the comments from the https://github.com/gcanti/fp-ts/issues/1103 issue, I'm assuming we can still improve the documentation from the `src` files. This is my first attempt at this task, please let me know if the changes proposed on this branch are OK. I'd like to contribute more on this matter, but first I want to make sure these kinds of changes are acceptable for such an amazing code base!

- I'm using the [optional chaining](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html#optional-chaining) from TS 3.7 in the example, I hope that's fine.
- I had to use [markserv](https://github.com/markserv/markserv) to serve the Markdown files from the `docs` directory as HTML pages. Is there a "better" way to check the changes regarding the documentation on local?
- Do we have some guidelines somewhere regarding the size/complexity of descriptions and examples?

### DIFF
--- a/docs/modules/Option.ts.md
+++ b/docs/modules/Option.ts.md
@@ -150,10 +150,46 @@ Added in v2.0.0
 
 # alt
 
+Similar to [Option.orElse](<https://www.scala-lang.org/api/current/scala/Option.html#orElse[B%3E:A](alternative:=%3EOption[B]):Option[B]>) from Scala,
+`alt` provides an alternative value of type `Option<A>` as a fallback when `fa` is `None`.
+
 **Signature**
 
 ```ts
 <A>(that: () => Option<A>) => (fa: Option<A>) => Option<A>
+```
+
+**Example**
+
+```ts
+import { Option, fromNullable, filter, alt } from 'fp-ts/lib/Option'
+import { pipe } from 'fp-ts/lib/pipeable'
+
+interface Store {
+  a?: {
+    b?: unknown
+  }
+}
+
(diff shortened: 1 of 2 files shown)


gcanti:
@ruizb thanks for this PR. Honestly I don't know what to do with functions like `alt` (or `map`, `chain`, etc...)

Theoretically if you know what the type class `Alt` is, you don't need any comment / example.

ruizb:
Hi Giulio! I understand, at first I thought the same way as you did regarding the documentation for all these functions. However, I think it's useful when exploring and looking for "equivalent" code from other languages such as Scala or frameworks/libraries such as RxJS.

I know there's a tradeoff here:
- Adding this documentation means more work regarding code maintainability when breaking changes occur.
- It also means we might get a feeling of "duplication" if we start adding code examples for _every_ `alt`, `map`, etc. of every type class instances in this project.

I still think it's useful to have examples like this one so people can discover FP using this library, without the need of having prior knowledge about Fantasy-Land and all the FP vocabulary.

Is there such a thing as "too much documentation/examples" though? 😁

gentoid:
Hi guys. Just came to a thought: what if provide a link to the corresponding type class? So `alt` will have a link to `Alt`, etc. What do you think guys?

mohaalak:
I was checking this PR and I think the documentation was on point. I know that if you know Alt you don't need any document for it but it will help a lot to beginners to know what's the use case for some function. if you accept these kinds of PRs I think I can write some too.

PR #1181, opened 2020-04-06
Safe split and join

cyberixae:
This PR adds string `split` and `join` functions that work with NonEmptyArrays.

### DIFF
--- a/docs/modules/Strong.ts.md
+++ b/docs/modules/Strong.ts.md
@@ -1,6 +1,6 @@
 ---
 title: Strong.ts
-nav_order: 86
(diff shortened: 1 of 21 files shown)


gcanti:
Are those signatures correct?

```ts
import * as assert from 'assert'

assert.deepStrictEqual([].join(''), '')
assert.deepStrictEqual(''.split(''), [])
```

cyberixae:
Never mind. It doesn't work. The `''.split('')`  case is particularly  nasty. I can't figure out a reliable type signature that would guarantee one of the strings to be non-empty.

PR #1190, opened 2020-04-25
Add missing StateM2C and StateM3C

qlonik:
Add missing `StateM2C` and `StateM3C` definitions, as well as their usage in `getStateM()`.

I added `Since @2.5.4` in the jsdoc, but i'm not sure what is the value that actually need to go there. Maybe it should be `@2.6.0`? 

As a side note, it seems that it is possible to type `state` in `src/State.ts` on line 88, as `StateM1<IdentityURI>`. This would allow state to be passed in functions which require `StateM` instances. Should that be added?


gcanti:
Thanks @qlonik, 2.5.4 is fine. Released.

PR #1200, opened 2020-05-06
add Eq.getMonoid

gcanti:


gkamperis:
@gcanti should this be in `Semigroup` instead?

...and maybe as `semigroupEq` similar to `semigroupString`, `semigroupSum` etc?
or as `getEqSemigroup` similar to `getObjectSemigroup`

gcanti:
@gkamperis usually `Semigroup` / `Monoid` instances are contained in the corresponding data type module, examples

- `Array.ts` contains `getMonoid`
- `Option.ts` contains `getMonoid`
- `Either.ts` contains `getSemigroup`

etc...

`semigroupString`, `semigroupSum`, `getObjectSemigroup` are in `Semigroup.ts` because there's no `string.ts`, `number.ts`, `Object.ts` modules.

Though maybe I should add `getMonoid` instead of `getSemigroup`

```ts
const empty: Eq<unknown> = {
  equals: () => true
}

/**
 * @since 2.6.0
 */
export function getMonoid<A>(): Monoid<Eq<A>> {
  return {
    concat: (x, y) => fromEquals((a, b) => x.equals(a, b) && y.equals(a, b)),
    empty
  }
}
```

gkamperis:
@gcanti thanks for the clarification.

the other reason I made the comment is that in my mind, and I am a total novice here, the `Eq `abstraction is a building block for `Semigroup` so the `Semigroup` module should "reference" `Eq `and not the other way around.

gcanti:
> the Eq abstraction is a building block for Semigroup

what do you mean?

gkamperis:
nothing... :) pls ignore me! I understand that the emphasis is on `Eq`... thanks again for the time to explain it

test/Eq.ts
```diff
@@ -63,11 +63,11 @@ describe('Eq', () => {
 
   it('getSemigroup', () => {
```
gkamperis, line 64: change desc to getMonoid?

Issue #1087, opened 2020-01-14
Tree shaking: function calls on module instantiation are not removed


OliverJAsh:
### Current Behavior

[Reduced test case](https://github.com/OliverJAsh/tree-shaking-test/tree/fp-ts-ord-rollup)

`src/index.js`:

```ts
import { max } from "fp-ts/es6/Ord";
console.log(max);
```

`rollup.config.js`:

```js
import resolve from "@rollup/plugin-node-resolve";

export default {
  input: "src/index.js",
  output: {
    file: "target/rollup.js",
    format: "cjs"
  },
  plugins: [resolve()]
};
```

`package.json`:

```json
{
  "dependencies": {
    "fp-ts": "^2.0.0",
    "@rollup/plugin-node-resolve": "^7.0.0",
    "rollup": "^1.29.0"
  }
}
```

Then run `rollup -c rollup.config.js`.

### Expected behavior

The output of rollup should not include lots of unused code:

```js
'use strict';

/**
 * The `Ord` type class represents types which support comparisons with a _total order_.
 *
 * Instances should satisfy the laws of total orderings:
 *
 * 1. Reflexivity: `S.compare(a, a) <= 0`
 * 2. Antisymmetry: if `S.compare(a, b) <= 0` and `S.compare(b, a) <= 0` then `a <-> b`
 * 3. Transitivity: if `S.compare(a, b) <= 0` and `S.compare(b, c) <= 0` then `S.compare(a, c) <= 0`
 *
 * See [Getting started with fp-ts: Ord](https://dev.to/gcanti/getting-started-with-fp-ts-ord-5f1e)
 *
 * @since 2.0.0
 */
/**
 * Take the maximum of two values. If they are considered equal, the first argument is chosen
 *
 * @since 2.0.0
 */
function max(O) {
    return function (x, y) { return (O.compare(x, y) === -1 ? y : x); };
}

console.log(max);
```

### Actual behaviour

The output of rollup includes lots of unused code:

<details>
<summary>Show code snippet</summary>

```js
'use strict';

/**
 * @since 2.0.0
 */
/**
 * @since 2.0.0
 */
function identity(a) {
    return a;
}

var isFunctor = function (I) { return typeof I.map === 'function'; };
var isContravariant = function (I) { return typeof I.contramap === 'function'; };
var isFunctorWithIndex = function (I) { return typeof I.mapWithIndex === 'function'; };
var isApply = function (I) { return typeof I.ap === 'function'; };
[...]

PR #1212, opened 2020-05-16
Tree Shakable Either

mikearnaldi:
This PR is a prototype for Tree Shakability to work with fp-ts, it removes pipeable and exposes additional data-last functions (with _ suffix) used to avoid always using the full instance for composition in other modules and external code.

This PR updates the Either module in a non-breaking way and it is fully shakable by webpack 4+ and rollup.

If the changes are accepted I will continue with the remaining modules giving priority to modules used in the libraries of the fp-ts core ecosystem like newtype-ts, monocle-ts, io-ts, fp-ts-contrib leaving last unused modules.

Any help on this is hugely welcome!

### DIFF
--- a/docs/modules/Either.ts.md
+++ b/docs/modules/Either.ts.md
@@ -48,22 +48,29 @@ Added in v2.0.0
 - [URI (type alias)](#uri-type-alias)
 - [URI](#uri)
 - [alt](#alt)
+- [alt\_](#alt_)
 - [ap](#ap)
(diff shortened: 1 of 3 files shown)


mikearnaldi:
closing as data last functions exposed are not welcome. might be reopen if we decide to kill pipeable #1087

raveclassic:
related https://github.com/gcanti/fp-ts/pull/942

PR #1214, opened 2020-05-18
Add raise function to throw an error as an expression (to 1.x)

Nymphium:
The function provided is just a wrapper of `throw` *statement*.


gcanti:
See https://github.com/gcanti/fp-ts/pull/1215#issuecomment-629971740

PR #1215, opened 2020-05-18
Add raise function to throw an error as an expression (to 2.x)

Nymphium:
The function provided is just a wrapper of `throw` *statement*.
I also add it to 1.x (see #1214).


gcanti:
@Nymphium thanks for this PR but throwing errors is not something I want to add suppor for

PR #1219, opened 2020-05-25
Fix IO example

sgioia9:
In the example:
```ts
pipe(
    fromNullable(input),
    fold(log('Received null'), value => log(`Received ${value}`)),
  );
```
the left side of the fold needs to return a thunk, otherwise the resulf of the pipe isn't callable.


gcanti:
Thanks @sgioia9

Issue #1227, opened 2020-06-03
Import Ergonomics


SRachamim:
## 🚀 Feature request
I think the entire import design is a mess. I find it hard to auto-import functions, types and utilities.

### Current Behavior
This is how a typical file of mine looks:

```typescript
import { pipe } from 'fp-ts/lib/pipeable';
import { flow, identity } from 'fp-ts/lib/function';
import { stateReaderTaskEither, either } from 'fp-ts';
import { StateReaderTaskEither } from `fp-ts/lib/StateReaderTaskEither';

// weird use of proofs
either.either
```

### Desired Behavior

I'de like to find everything on the index:

```typescripr
import {
  pipe,
  flow,
  identity,
  Either,
  either,
} from 'fp-ts';

// general name for proofs (I'm a noob, so maybe there's a better name
either.proof
```

### Suggested Solution

In addition to the current `index.ts`, it could be better to export all types, all `function` utilities and `pipe`.

Consider also to export very handy utilities like `sequnce` and `traverse`.

### Who does this impact? Who is this for?

Tree shaking is not always relevant. In those cases I prefer ergonomics: auto-import should work like a charm and import should be grouped so it would be easy to manual import additional utilities.

steida:
It's already answered somewhere in issues. Long story short: It's already perfect except Webpack 4 will not tree-shake it properly. Webpack5 will.

steida:
It's already solved. Search in issues. Webpack 5 will tree-shake it properly.

SRachamim:
What already solved? tree-shaking is not the issue here. Try to read the feature-request again and let me know if I wasn't clear enough.

steida:
@SRachamim Jeez, it's so hard to do what I recommend to you? Here is it: https://github.com/gcanti/fp-ts/issues/1185#issuecomment-613334914 The only problem is such importing is not properly tree-shaked with Webpack < 5 Now please say "Thank you" and close the issue. Thank you.

aqrln:
Hey, let's keep the conversation civil and on track, because it becomes kinda heated.

@gcanti what do you think about adopting a Code of Conduct, such as https://www.contributor-covenant.org?

aqrln:
[...]

Issue #1053, opened 2019-12-17
Split between lib/ and es6/ does not work well


wereHamster:
### Current Behavior

The following file, when processed by a bundler (rollup, webpack etc), will include the fp-ts Either module twice, once the es6 version and once the lib version.

```
import { right } from "fp-ts/es6/Either";
import { _right } from "monocle-ts/es6/Either";

console.log(right(42), _right(right(42)));
```

Why? Because the Either module from monocle imports `fp-ts/lib/Either`, even if itself was imported through `monocle-ts/es6/Either`.

### Expected behavior

The fp-ts Either module should appear only once in the output.

### Reproducible example

Save the above example as `input.js`, and bundle using the following rollup config:

```
import commonjs from "rollup-plugin-commonjs";
import nodeResolve from "rollup-plugin-node-resolve";

export default {
  input: "input.js",
  output: { file: "output.js", format: "cjs" },
  plugins: [nodeResolve({}), commonjs({})]
};
```

The resulting file will be 144KB, and contain a lot of unnecessary code because rollup can't threeshake commonjs modules.

If I edit the monocle-ts package contents to always import from fp-ts/es6/ instead of fp-ts/lib/, then the size drops to 50K

### Suggested solution(s)

The only solution I'm aware of is this: Every module that the user is expected to import in their code has to be a folder, and that folder must contain a package.json which provides pointers to the entry points (`main` being the commonjs entry point, and `module` being the ES6 module entry point).

In other words, in any import statement `import … from "SOMEMODULE"`, the "SOMEMODULE" must refer to a folder with its own package.json, ie. `node_modules/SOMEMODULE/package.json`. If SOMEMODULE contains path separators, same applies, eg. `import … from 'fp-ts/Either'` -> `node_modules/fp-ts/Either/package.json`.

### Additional context

The `@material-ui/core` package solves this in the way I described. You can have a look at the contents of their package here: https://unpkg.com/browse/@material-ui/core@4.8.0/.

Every direct subfolder of this package is allowed to be imported. As you drill down to the subfolders, you'll see a package.json en every one (eg. https://unpkg.com/browse/@material-ui/core@4.8.0/Avatar/package.json).

wmaurer:
This suggested solution has also been mentioned here:
https://github.com/gcanti/io-ts/issues/349#issuecomment-511234863

baetheus:
I have an alternative suggestion - stop shipping es5 with commonjs modules and only ship es6 with esmodules.

[...]

PR #1241, opened 2020-06-15
Expose fp-ts modules without lib/es6 prefix 

StefanoMagrassi:
This PR allows importing `fp-ts` modules without specifing the `/lib` or `/es6` prefix:

```ts
// from
import * as O from 'fp-ts/lib/Option'

// to
import * as O from 'fp-ts/Option'
```

Build tools will automatically resolve the right path based on the current environment, pointing to `/lib` version for commonjs and `/es6` version for ES6.

For now the `/lib` and `es6` paths are still available, so this feature is **completely backward compatible**.

This should enhance the library ergonomics and tree-shakeability.

**Note:** in order to achieve this goal, the standard `npm publish` command was disabled in favour of an `npm run release` command which builds the library and prepares the package for publication.

ref #1227 #1053


waynevanson:
I'm extremely eager for this! Great work.

anilanar:
@gcanti Is there something we can help to move this PR forward?

gcanti:
@anilanar planned for v2.8

Issue #1251, opened 2020-06-24
`Either.parseJSON` should return JSON type instead of `unknown`


OliverJAsh:
### Current Behavior

`Either.parseJSON` returns `Either<E, unknown>`, but we can refine the `unknown` type to something like `AnyJson`:

```ts
// Copied from https://github.com/Microsoft/TypeScript/issues/1897#issuecomment-338650717
export type AnyJson = boolean | number | string | null | JsonArray | JsonMap;
export type JsonMap = { [key: string]: AnyJson };
export type JsonArray = Array<AnyJson>;
```

Would you accept a PR for this?

### Your environment

Which versions of fp-ts are affected by this issue? Did this work in previous versions of fp-ts?

| Software   | Version(s) |
| ---------- | ---------- |
| fp-ts      |            |
| TypeScript |            |

gcanti:
> Would you accept a PR for this?

👍

Issue #685, opened 2019-01-03
Rename `Record` to `Dict`?


OliverJAsh:
**Do you want to request a _feature_ or report a _bug_?**
Naming suggestion

**What is the current behavior?**
The `Record` module seems to model a "dictionary" type, where the keys are unknown (`Record<string, A>`).

If I understand correctly, the name "record" usually refers to an object with known keys, and "dictionary"/"map" refers to an object with unknown keys.

If that's correct, perhaps we should rename the `Record` module to `Dict{,ionary}`?

It would be fine to continue using TypeScript's built-in `Record` helper to model the dictionary (`Record<string, A>`), as this is just an alias for `{ [K in string]: A }`. Alternatively, you could define the dictionary type using an index signature instead of mapped types: `{ [key: string]: A }`. I'm not sure if there's much difference.

gcanti:
I don't like it either, but that's how the built-in type is called

OliverJAsh:
Hmm, true! I resolved to only using the built-in type `Record` for actual records (where the key is not just `string`), and then for dictionaries using index signatures (instead of mapped types which `Record` is an alias for);

``` ts
type Dict<T> = { [index: string]: T }
```

gcanti:
> where the key is not just `string`

some functions could take into account that, e.g. `size`, `isEmpty`, `map`...

Example

```ts
export function map<K extends PropertyKey, A, B>(fa: Record<K, A>, f: (a: A) => B): Record<K, B>
export function map<A, B>(fa: Record<string, A>, f: (a: A) => B): Record<string, B>
export function map<A, B>(fa: Record<string, A>, f: (a: A) => B): Record<string, B> {
  return mapWithKey(fa, (_, a) => f(a))
}
```

```ts
declare const x: Record<'a' | 'b', number>

// y: Record<"a" | "b", boolean>
const y = R.map(x, n => n > 2)
```

giogonzo:
@gcanti this would help, I had to do some similar casts when `map`ping over a record with known keys . How could `size`, `isEmpty` have benefits instead?

giogonzo:
Real world example:
```ts
const inlineErrors : InlineErrors<K> = mapWithKey( // type InlineErrors<K extends string> = Record<K, Option<string>>;
  this.props.inlineErrors,
[...]

PR #1252, opened 2020-06-24
Refine `Either.parseJSON` return type

OliverJAsh:
Fixes https://github.com/gcanti/fp-ts/issues/1251

### DIFF
--- a/docs/modules/Either.ts.md
+++ b/docs/modules/Either.ts.md
@@ -417,7 +417,7 @@ Converts a JavaScript Object Notation (JSON) string into an object.
 **Signature**
 
 ```ts
-export declare function parseJSON<E>(s: string, onError: (reason: unknown) => E): Either<E, unknown>
+export declare function parseJSON<E>(s: string, onError: (reason: unknown) => E): Either<E, AnyJson>
 ```
 
 **Example**
--- a/docs/modules/Either.ts.md
+++ b/docs/modules/Either.ts.md
@@ -88,6 +88,9 @@ Added in v2.0.0
   - [Left (interface)](#left-interface)
   - [Right (interface)](#right-interface)
 - [utils](#utils)
+  - [Json (type alias)](#json-type-alias)
+  - [JsonArray (interface)](#jsonarray-interface)
+  - [JsonRecord (interface)](#jsonrecord-interface)
   - [elem](#elem)
   - [exists](#exists)
   - [toError](#toerror)
@@ -417,7 +420,7 @@ Converts a JavaScript Object Notation (JSON) string into an object.
 **Signature**
 
 ```ts
-export declare function parseJSON<E>(s: string, onError: (reason: unknown) => E): Either<E, AnyJson>
+export declare function parseJSON<E>(s: string, onError: (reason: unknown) => E): Either<E, Json>
 ```
 
 **Example**
@@ -802,6 +805,38 @@ export interface Right<A> {
 Added in v2.0.0
 
 # utils
+
+## Json (type alias)
+
+**Signature**
+
+```ts
+export type Json = boolean | number | string | null | JsonArray | JsonRecord
+```
+
+Added in v2.6.7
+
+## JsonArray (interface)
+
+**Signature**
+
+```ts
+export interface JsonArray extends ReadonlyArray<Json> {}
+```
+
+Added in v2.6.7
+
+## JsonRecord (interface)
+
+**Signature**
+
+```ts
+export interface JsonRecord {
+  readonly [key: string]: Json
+}
+```
+
+Added in v2.6.7
 
 ## elem
 
--- a/src/Either.ts
+++ b/src/Either.ts
@@ -169,6 +169,13 @@ export function tryCatch<E, A>(f: Lazy<A>, onError: (e: unknown) => E): Either<E
   }
 }
 
+// Copied from https://github.com/Microsoft/TypeScript/issues/1897#issuecomment-338650717
+type AnyJson = boolean | number | string | null | JsonArray | JsonMap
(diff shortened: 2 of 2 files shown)


src/Either.ts
```diff
@@ -169,6 +169,13 @@ export function tryCatch<E, A>(f: Lazy<A>, onError: (e: unknown) => E): Either<E
   }
 }
 
+// Copied from https://github.com/Microsoft/TypeScript/issues/1897#issuecomment-338650717
+type AnyJson = boolean | number | string | null | JsonArray | JsonMap
+interface JsonMap {
+  readonly [key: string]: AnyJson
```
OliverJAsh, line 175: I made this `readonly` because the lint rule told me to. 🤷‍♂️

src/Either.ts
```diff
@@ -169,6 +169,13 @@ export function tryCatch<E, A>(f: Lazy<A>, onError: (e: unknown) => E): Either<E
   }
 }
 
+// Copied from https://github.com/Microsoft/TypeScript/issues/1897#issuecomment-338650717
+type AnyJson = boolean | number | string | null | JsonArray | JsonMap
+interface JsonMap {
```
OliverJAsh, line 174: I initially use `type` but recursive type aliases are only supported in newer TS versions, so had to revert to `interface`.

OliverJAsh:
Sure, let me do that.

> * rename `JsonMap` to `JsonRecord` for consistency with the built-in `Record` type

Note: I still feel uncomfortable using "record" here (https://github.com/gcanti/fp-ts/issues/685).

OliverJAsh:
@gcanti Maybe we should export them too, WDYT?

gcanti:
@OliverJAsh yes, that's a useful type which [I would like to import somewhere else](https://github.com/gcanti/io-ts/blob/c85564afc304b852033e0dc1b99ca22cea920afa/src/JsonEncoder.ts#L13-L31)

docs/modules/Either.ts.md
```diff
@@ -417,7 +417,7 @@ Converts a JavaScript Object Notation (JSON) string into an object.
 **Signature**
 
 ```ts
-export declare function parseJSON<E>(s: string, onError: (reason: unknown) => E): Either<E, unknown>
+export declare function parseJSON<E>(s: string, onError: (reason: unknown) => E): Either<E, AnyJson>
```
gcanti, line 420: looks like the generated docs are stale

awerlogus:
I think, you should add also `| readonly []` in `AnyJson` because https://github.com/microsoft/TypeScript/issues/38727#issuecomment-643495070

gcanti:
@awerlogus why? all tuples, `readonly []` included, are already modeled by `JsonArray`

awerlogus:
@gcanti compare the types of variables `a` and `b`.

```ts
type Json1 = boolean | number | string | null | JsonArray1 | JsonRecord1

interface JsonArray1 extends ReadonlyArray<Json1> { }

export interface JsonRecord1 {
  readonly [key: string]: Json1
}

function handleJson1<T extends Json1>(arg: T): T {
  return arg
}


type Json2 = boolean | number | string | null | JsonArray2 | JsonRecord2 | readonly []

interface JsonArray2 extends ReadonlyArray<Json2> { }

export interface JsonRecord2 {
  readonly [key: string]: Json2
}

function handleJson2<T extends Json2>(arg: T): T {
  return arg
}

// number[]
const a = handleJson1([1, 2, 3])

// [number, number, number]
const b = handleJson2([1, 2, 3])
```

Type parameters will be not inferred as tuples if the type that extends them doesn't contain `readonly []` (or another tuple) explicitly. That's why I always combine `ReadonlyArray` with `readonly []`.

gcanti:
@awerlogus ok but `handleJson` is just the identity function, could you show a compelling use case where `Json2` will be actually useful?

awerlogus:
@gcanti 

> could you show a compelling use case where `Json2` will be actually useful?

You're requesting for a json processing example? It obviously will be absolutely synthetic. Somebody may be needed to handle tuples and simple arrays differently on type level. And it will be better if types will be inferred as narrow as possible. That's all. Adding `readonly []` is not such a serious suggestion, which will determine the development of humanity for years to come, to be taken with such hostility.

gcanti:
@awerlogus there's no hostility in my request, I'm genuinely interested in a use case since I never thought of adding `readonly []` to the untagged union (as linked in my comment [here](https://github.com/gcanti/fp-ts/pull/1252#issuecomment-649204318)) and I may be missing something important.

awerlogus:
@gcanti as I said, the example turned out to be synthetic. Here you can see that type inference in case of `getJsonLength2` provides us more info to operate with on type level.

```ts
type Json1 = boolean | number | string | null | JsonArray1 | JsonRecord1

interface JsonArray1 extends ReadonlyArray<Json1> { }

export interface JsonRecord1 {
  readonly [key: string]: Json1
}

declare function getJsonLength1<T extends Json1>(arg: T): T extends ReadonlyArray<any> ? T['length'] : 1


type Json2 = boolean | number | string | null | JsonArray2 | JsonRecord2 | readonly []

interface JsonArray2 extends ReadonlyArray<Json2> { }

export interface JsonRecord2 {
  readonly [key: string]: Json2
}

declare function getJsonLength2<T extends Json2>(arg: T): T extends ReadonlyArray<any> ? T['length'] : 1


// number
const a = getJsonLength1([1, 2, 3])

// 3
const b = getJsonLength2([1, 2, 3])


// 1
const c = getJsonLength1({ a: 'foo', b: 'bar' })

// 1
const d = getJsonLength2({ a: 'foo', b: 'bar' })
```

PR #1259, opened 2020-07-01
Record / ReadonlyRecord: handle this in `hasOwnProperty`, fix #1249

### LINKED ISSUE #1249: Re-exporting the Record module creates a runtime exception
### Current Behavior

We usually "extend" some modules with our own operators and we do this by re-exporting the module from another file such as:

```ts
// modules/Record/index.ts
export * from 'fp-ts/lib/Record';
export * from './operators';

// modules/Record/operators.ts
import * as R from 'fp-ts/lib/Record';

export const fn = () => {}
```

So in the codebase, we can do:

```ts
import * as R from 'modules/Record';

R.fn();
R.keys();
...
```

We've been doing that for different modules such as `Option`, `Array`, `Either` etc and all is good but for some reason re-exporting `Record` creates the following runtime exception

image

### Expected behavior

No runtime exception. Allowing us to re-export the module to extend it.

### Your environment

| Software   | Version(s) |
| ---------- | ---------- |
| fp-ts      |           2.6.5 |
| TypeScript |            3.9.2|

gcanti:


test/ReadonlyRecord.ts
```diff
@@ -407,6 +407,11 @@ describe('ReadonlyRecord', () => {
     const x: _.ReadonlyRecord<string, number> = { a: 1 }
     assert.deepStrictEqual(_.hasOwnProperty('a', x), true)
     assert.deepStrictEqual(_.hasOwnProperty('b', x), false)
+    // TODO: remove in v3
```
OliverJAsh, line 410: What will we do instead of this in v3?

PR #1264, opened 2020-07-09
Task: implement neglect function

cyberixae:
This PR adds `neglect` function that turns a Task into a detached background job.

### DIFF
--- a/docs/modules/Task.ts.md
+++ b/docs/modules/Task.ts.md
@@ -55,6 +55,7 @@ Added in v2.0.0
 - [model](#model)
   - [Task (interface)](#task-interface)
 - [utils](#utils)
+  - [neglect](#neglect)
   - [never](#never)
 
 ---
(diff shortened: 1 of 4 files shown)


gcanti:
@cyberixae not sure I'm following, `neglect` is just the `identity` function

```ts
export const neglect: (t: Task<unknown>) => IO<void> = identity
```

cyberixae:
> @cyberixae not sure I'm following, `neglect` is just the `identity` function
> 
> ```ts
> export const neglect: (t: Task<unknown>) => IO<void> = identity
> ```

On type level they are the same but if you `console.log` the `void` returned by my `neglect` implementation it actually returns `undefined` since it discards the Promise instead of returning it. If I replace the implementation with `identity` the void value will be `Promise<something>` at run time, which I find to be a bit confusing.

gcanti:
What I meant is that I don't see how `neglect` would be useful in practice since is just the identity function.

Also, keep in mind that in TypeScript `void` doesn't mean that a function returns `undefined` at runtime

```ts
const f: (s: string) => void = (s) => s.length // no error
```

Thanks for this PR but I'm afraid I can't accept it

cyberixae:
Wait, what?! I encountered the problem because I was using the `function` keyword instead of double arrow. See the attached screenshot.

vscerr

cyberixae:
Anyway, thanks for reviewing the PR. I never noticed before that void behaved that way with double arrow.

PR #1155, opened 2020-03-03
Added Option.toArray

psilospore:
In Scala I can do `Option(1).toList` this would do something similar

Issue #1226, opened 2020-06-03
Consider dropping support for mutable modules


gcanti:
Namely:

- Array
- Map
- NonEmptyArray
- Record
- Set
- Tuple

SRachamim:
As a functional library it's OK to enforce immutability.

I would get rid of all `Readonly*` modules and make the rest immutable.

mlegenhausen:
Would love to see that readonly is the default in the whole `fp-ts` ecosystem but I already had my struggle to switch to the readonly instances with `fp-ts@2` because `fp-ts-contrib`, `io-ts-types` and `monocle-ts` has not full readonly support. If we could add the readonly instances e. g. `ReadonlyZipper` for `fp-ts-contrib`, `readonlyNonEmptyArray` for `io-ts-types` and `indexReadonlyNonEmptyArray` for `monocle-ts` I would already give it a try.

gcanti:
> and indexReadonlyNonEmptyArray for monocle-ts

@mlegenhausen https://github.com/gcanti/monocle-ts/pull/124

gcanti:
> I would get rid of all Readonly* modules and make the rest immutable

There are two options I guess:

**Option 1**

Keep the `Readonly*` modules and simply remove the mutable counterparts (:+1: if you prefer this option)

**Option 2**

Remove all the mutable modules and **rename** `Readonly<X>` module to `<X>` (:-1: if you prefer this option)

gkamperis:
@gcanti based on a similar [conversation...](https://github.com/gcanti/fp-ts/pull/1082)

this [explains](https://github.com/gcanti/fp-ts/pull/1082#issuecomment-573446771) my vote :)

SRachamim:
We can see this issue in two lights:

1. We're dealing with a strongly-typed pure functional library. We're expecting only readonly data-structures and we're promising to give back only readonly data structures. In this case, the immutability is an axiom, and we don't need to explicitly prefix all our data-structures with a `Readonly` prefix. We'll have one `Array` module which is readonly.

...or:

2. We're a _TypeScript_ library, in which the default is mutable structures. There is already a convention of prefixing with `Readonly` all immutable data-structures. This is aligned with the current situation, in which we have both mutable `Array` and immutable `ReadonlyArray`.

I tend to like the first philosophy.

anilanar:
[...]

PR #1272, opened 2020-07-21
Foldable: add `toArray`

newswim:
👋 

I saw an open PR, "Added Option.toArray" (https://github.com/gcanti/fp-ts/pull/1155), and noticed @gcanti recommended instead adding `toArray` to Foldable.

I went ahead and added it. However, I'm not 100% sure about the implementation.

In 2.x+, `Array` is deprecated in favor of `ReadOnlyArray`, but the original implementation of `toArray` uses a combination of `reduce` and `arr.push`.

**Possible problem**: `.push` does not exist on `ReadOnlyArray` since it mutates the array in place. This method is probably faster, but it might not desirable in the 2.x+ world.

Would it make sense to add another instance, `toReadOnlyArray`?

**Update**: I went ahead and also added a `toReadOnlyArray` instance. There appeared to be some prior art in the Map, Set, and Record modules.

I'm open to feedback and/or edits.

---
- [x] Fork [the repository](https://github.com/gcanti/fp-ts) and create your branch from `master`.
- [x] Run `npm install` in the repository root.
- [x] If you've fixed a bug or added code that should be tested, add tests!
- [x] Ensure the test suite passes (`npm test`).

### DIFF
--- a/docs/modules/Foldable.ts.md
+++ b/docs/modules/Foldable.ts.md
@@ -32,6 +32,7 @@ Added in v2.0.0
   - [foldM](#foldm)
   - [getFoldableComposition](#getfoldablecomposition)
   - [intercalate](#intercalate)
+  - [toArray](#toArray)
   - [traverse\_](#traverse_)
 
 ---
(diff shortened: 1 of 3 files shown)


gcanti:
@newswim thanks for this PR, I would only keep the readonly array version (which is future proof, `fp-ts@3.x` will drop all mutable modules https://github.com/gcanti/fp-ts/issues/1226)

```ts
/**
 * Transforms a `Foldable` into a read-only array.
 *
 * @example
 * import { toArray } from 'fp-ts/lib/Foldable'
 * import { tree, make } from 'fp-ts/lib/Tree'
 *
 * const t = make(1, [make(2, []), make(3, []), make(4, [])])
 * assert.deepStrictEqual(toArray(tree)(t), [1, 2, 3, 4])
 *
 * @since 2.8.0
 */
export function toArray<F extends URIS4>(F: Foldable4<F>): <S, R, E, A>(fa: Kind4<F, S, R, E, A>) => ReadonlyArray<A>
export function toArray<F extends URIS3>(F: Foldable3<F>): <R, E, A>(fa: Kind3<F, R, E, A>) => ReadonlyArray<A>
export function toArray<F extends URIS3, E>(F: Foldable3C<F, E>): <R, A>(fa: Kind3<F, R, E, A>) => ReadonlyArray<A>
export function toArray<F extends URIS2>(F: Foldable2<F>): <E, A>(fa: Kind2<F, E, A>) => ReadonlyArray<A>
export function toArray<F extends URIS2, E>(F: Foldable2C<F, E>): <A>(fa: Kind2<F, E, A>) => ReadonlyArray<A>
export function toArray<F extends URIS>(F: Foldable1<F>): <A>(fa: Kind<F, A>) => ReadonlyArray<A>
export function toArray<F>(F: Foldable<F>): <A>(fa: HKT<F, A>) => ReadonlyArray<A>
export function toArray<F>(F: Foldable<F>): <A>(fa: HKT<F, A>) => ReadonlyArray<A> {
  return <A>(fa: HKT<F, A>) =>
    // tslint:disable-next-line: readonly-array
    F.reduce(fa, [], (acc: Array<A>, a) => {
      acc.push(a)
      return acc
    })
}
```

p.s.
docs are auto-generated by running `npm run docs`

newswim:
Ah, thank you!

I believe I made the appropriate updates, let me know!

PR #1293, opened 2020-08-07
ordBoolean test should test boolean compare

kwasniew:
Fixing test body to reflect what the test is describing.


gcanti:
Thanks @kwasniew

PR #1311, opened 2020-09-09
Replace ts-jest by @swc-node/jest to speedup jest tests

Brooooooklyn:
## `@swc-node/jest`

```
Test Suites: 64 passed, 64 total
Tests:       1193 passed, 1193 total
Snapshots:   0 total
Time:        4.434 s
Ran all test suites.
```

## `ts-jest`
```
Test Suites: 64 passed, 64 total
Tests:       1193 passed, 1193 total
Snapshots:   0 total
Time:        41.142 s
Ran all test suites.
```

### DIFF
--- a/jest.config.js
+++ b/jest.config.js
@@ -1,5 +1,5 @@
 module.exports = {
-  preset: 'ts-jest',
+  preset: '@swc-node/jest',
   testEnvironment: 'node',
   collectCoverage: true,
   collectCoverageFrom: [
@@ -12,9 +12,6 @@ module.exports = {
     '!src/WriterT.ts',
     '!src/ValidationT.ts'
   ],
-  transform: {
-    '^.+\\.tsx?$': 'ts-jest'
-  },
   testRegex: 'test',
   moduleFileExtensions: ['ts', 'js'],
   coverageThreshold: {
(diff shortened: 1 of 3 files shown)


DenisFrezzato:
It's worth mentioning that `@swc-node/jest` is faster because the type checking is skipped. I do think that type checking tests is a desired behaviour.

Brooooooklyn:
> I do think that type checking tests is a desired behaviour.

I think `TypeCheck` was already done by `dtslint` and `build` command. `TypeCheck` in `jest` is redundant.

> It's worth mentioning that @swc-node/jest is faster because the type checking is skipped

`@swc-node/jest` is also far more faster than `ts-jest` without `typecheck`.

Brooooooklyn:
Feel free to close if you think this PR is unnecessary

DenisFrezzato:
> I think TypeCheck was already done by dtslint and build command. TypeCheck in jest is redundant.

It's not redundant. Neither `dtslint` nor the build check the test files.

If we do feel the need to speed up tests, we can split compilation and execution in two steps: if the compilation is made with `tsc`, we can make leverage on `incremental` option, and then just run the test suite on the generated JS files.

PR #1328, opened 2020-10-13
Feature/docs

SRachamim:
Add descriptions and examples to each of the following functions in `Either`:

- `of`
- `foldMap`
- `reduce`
- `reduceRight`
- `flatten`
- `sequence`
- `traverse`

### DIFF
--- a/docs/modules/Either.ts.md
+++ b/docs/modules/Either.ts.md
@@ -286,10 +286,26 @@ Added in v2.0.0
 
 ## reduce
 
+Left-associative fold of a structure.
+
 **Signature**
 
 ```ts
 export declare const reduce: <A, B>(b: B, f: (b: B, a: A) => B) => <E>(fa: Either<E, A>) => B
+```
+
+**Example**
+
+```ts
+import { pipe } from 'fp-ts/function'
+import * as E from 'fp-ts/Either'
+
+const empty = 'prefix'
+const concat = (a: string, b: string) => `${a}:${b}`
+
+assert.deepStrictEqual(pipe(E.right('a'), E.reduce(empty, concat)), 'prefix:a')
+
+assert.deepStrictEqual(pipe(E.left('e'), E.reduce(empty, concat)), 'prefix')
 ```
 
 Added in v2.0.0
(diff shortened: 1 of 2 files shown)


raveclassic:
@gcanti This is great, let's merge

gcanti:
Thanks @SRachamim

docs/modules/Either.ts.md
```diff
@@ -250,32 +260,80 @@ Added in v2.0.0
...
 export declare const reduce: <A, B>(b: B, f: (b: B, a: A) => B) => <E>(fa: Either<E, A>) => B
 ```
 
+**Example**
+
+```ts
+import { pipe } from 'fp-ts/function'
+import * as E from 'fp-ts/Either'
+
+const empty = 'prefix'
```
gkamperis, line 303: as a noob I find this confusing vs the other "empty" like monoidString.empty above...
maybe a different var name?

docs/modules/Either.ts.md
```diff
@@ -346,12 +404,24 @@ Added in v2.6.0
...
 ```ts
 export declare const flatten: <E, A>(mma: Either<E, Either<E, A>>) => Either<E, A>
 ```
 
+**Example**
+
+```ts
+import * as E from 'fp-ts/Either'
+
+assert.deepStrictEqual(E.flatten(E.right(E.right('a'))), E.right('a'))
```
gkamperis, line 420: as it is removing one level only do you think this is a good example?

assert.deepStrictEqual(E.flatten(E.right(E.right(E.right('a')))), E.right(E.right('a')))

docs/modules/Either.ts.md
```diff
@@ -370,22 +440,51 @@ Added in v2.6.3
...
+**Example**
+
+```ts
+import { pipe } from 'fp-ts/function'
+import * as E from 'fp-ts/Either'
+import * as O from 'fp-ts/Option'
+
+assert.deepStrictEqual(pipe(E.right(O.some('a')), E.sequence(O.option)), O.some(E.right('a')))
+
+assert.deepStrictEqual(pipe(E.right(O.none), E.sequence(O.option)), O.none)
```
gkamperis, line 460: this baffles me as noob... if I can have a right of none then why can I not have a some of right of none?

also would be great if there was a left example....

PR #1335, opened 2020-10-20
Add fp-ts-std to ecosystem docs

samhh:
:slightly_smiling_face:


anthonyjoeseph:
`omit` from `fp-ts-std/lib/Record` just saved my day!

PR #1339, opened 2020-10-25
refactor: use JSON.stringify in `toError` instead of String casting

vecerek:
A tiny improvement to `toError` in case an object is thrown. A comparison of string casting to JSON stringify:

```sh
> String({ a: "123" })
'[object Object]'
> JSON.stringify({ a: "123" })
'{"a":"123"}'
```

Motivation: sometimes there's not much to do about errors other than logging them; `[object Object]` does not provide much value in logs for debugging purposes.

### DIFF
--- a/src/Either.ts
+++ b/src/Either.ts
@@ -1021,7 +1021,7 @@ export const either: Monad2<URI> &
  * @since 2.0.0
  */
 export function toError(e: unknown): Error {
-  return e instanceof Error ? e : new Error(String(e))
+  return e instanceof Error ? e : new Error(JSON.stringify(e))
 }
 
 /**
--- a/test/Either.ts
+++ b/test/Either.ts
@@ -352,7 +352,14 @@ describe('Either', () => {
           // tslint:disable-next-line: no-string-throw
           throw 'string error'
         }, _.toError),
-        _.left(new Error('string error'))
+        _.left(new Error('"string error"'))
+      )
+
+      assert.deepStrictEqual(
+        _.tryCatch(() => {
+          throw { ECODE: 1 }
+        }, _.toError),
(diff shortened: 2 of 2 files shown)


raveclassic:
Note that `JSON.stringify` is unsafe and may throw

vecerek:
@raveclassic I guess the best that can be done is to fallback to casting the unknown to a string in case `JSON.stringify` throws. I'd add a test case for it but I don't know what makes it throw an error 😅

DenisFrezzato:
> I don't know what makes it throw an error

See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify#Exceptions.

SRachamim:
Use `fc.anything()` in order to make sure that it works on any input (including a record).

vecerek:
> See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify#Exceptions.

Thanks, @DenisFrezzato. I've added a test case featuring `BigInt`. I also discovered the `stringifyJson` function inside the `Either` module, so I decided to reimplement it using that one. I've amended my previous commit in https://github.com/gcanti/fp-ts/pull/1339/commits/c5b3a963066fcbe632541c06ceee4c967bf02dbf.

> Consider using `fc.anything()` in order to make sure that it works on any input (including a record).

@SRachamim thanks for the suggestion. Unfortunately, I don't have any experience using `fast-check` and I found it difficult to come up with a law for `toError` since it's a mapping between an `unknown` and a JSON string:
```
JSON.parse(toError(BigInt(Number.MAX_SAFE_INTEGER)).message) !== BigInt(Number.MAX_SAFE_INTEGER)
```
Do you have any suggestions?

Issue #1344, opened 2020-10-29
Solution for Stack Safety on sequencing arrays (no trampoline)


mohaalak:
# 🚀 Feature request

# Current Behavior

currently for using `sequence` or `traverse` on arrays we do this

```typescript
import * as A from "fp-ts/lib/Array";
import * as T from "fp-ts/lib/Task";

A.sequence(T.task)([T.of(1), T.of(2)])

```

but after reading this [issue](https://github.com/gcanti/fp-ts/issues/1315) I was thinking if we want to colocate the sequence of Array with our  data types we can optimize and get better performance also stack safety.
if  we look at issues concerning stack safety most(all) of them is there are some array that wants to traverse or sequence on some data types like [this issue](https://github.com/gcanti/fp-ts/issues/1177) or [this issue](https://github.com/gcanti/fp-ts/issues/983) or [this issue](https://github.com/gcanti/fp-ts/issues/907)

I write sequence arrays for major data types if the community and @gcanti think this is a good idea I can make a PR.

# Suggested Solution
I named this function `sequenceA`, A is an acronym for Array

## Option
as far as know option is stack safe but the `A.sequence(O.option)` is painfully slow for large arrays.

my solution is this

```typescript
const sequenceA = <A>(arr: Array<Option<A>>) => {
  const result = [];
  for (const a of arr) {
    if (a._tag === "None") {
      return none;
    }
    result.push(a.value);
  }
  return some(result);
};
```

I write a small benchmark to see how much difference it makes

```typescript
import * as O from "fp-ts/lib/Option";
import * as A from "fp-ts/lib/Array";
import { pipe } from "fp-ts/lib/pipeable";

const sequenceA = <A>(arr: Array<O.Option<A>>): Option<Array<A>> => {
  const result = [];
  for (const a of arr) {
    if (a._tag === "None") {
      return O.none;
    }
    result.push(a.value);
  }
  return O.some(result);
};

console.time("current");
pipe(A.range(0, 100000), A.map(O.some), A.sequence(O.option));
console.timeEnd("current");

console.time("suggestion");
pipe(A.range(0, 100000), A.map(O.some), sequenceA);
console.timeEnd("suggestion");
```

the result is 
```
current: 40.118s
suggestion: 17.546ms
```
> from now on I write just the result for other data types the array size is *100,000* for all of them

## Either

```typescript
export const sequenceA: <E, A>(
  arr: Array<Either<E, A>>
[...]

Issue #1315, opened 2020-09-17
Expose `Apply` interface in the different modules


SRachamim:
## 🚀 Feature request

I expect to be able to find `sequence`, `sequenceT`, `sequenceS` and `traverse` on the various modules which admits `Apply`.

### Current Behavior

I should import `Apply` and/or `Array` and use their `sequence`/`traverse` methods by powering them with the right typeclass, which is akward:

```typescript
import { pipe } from 'fp-ts/function';
import { apply, task } from 'fp-ts';

const ma = pipe(
  [1, 2, 3],
  apply.traverse(task.task)(task.right),
);
```

### Desired Behavior

```typescript
import { pipe } from 'fp-ts/function';
import { taskEither } from 'fp-ts';

const ma = pipe(
  [taskEither.right(1), taskEither.right(2), taskEither.right(3)],
  taskEither.sequence,
);
```

```typescript
import { pipe } from 'fp-ts/function';
import { task } from 'fp-ts';

const ma = pipe(
  [task.right(1), task.right('berlin')],
  task.sequenceT,
);
```

```typescript
import { pipe } from 'fp-ts/function';
import { readerTaskEither } from 'fp-ts';

const ma = pipe(
  { rank: readerTaskEither.right(1), readerTaskEither.right('berlin')],
  readerTaskEither.sequenceS,
);
```

```typescript
import { pipe } from 'fp-ts/function';
import { task } from 'fp-ts';

const ma = pipe(
  [1, 2, 3],
  task.traverse(task.right),
);
```

### Suggested Solution

Every module which admits `Apply` should export a `sequence`/`traverse` with batteries included:

On `Task.ts`:
```typescript
import { sequenceT, sequenceS } from './Apply';
import { sequence, traverse } from './Array';

export const sequenceT = sequenceT(task);
export const sequenceS = sequenceS(task);
export const sequence = sequence(task);
export const traverse = traverse(task); 
```

I can pull-request if you find it helpful and want to incorporate it into the library.

raveclassic:
Actually this would be helpful, I always find myself building them in eve single project where I use fp-ts.

gcanti:
Closed by https://github.com/gcanti/fp-ts/pull/1345

PR #1345, opened 2020-10-30
solution for stack safe & better performance for array sequence

mohaalak:
I think I explain most of things that I want to do [here](https://github.com/gcanti/fp-ts/issues/1344)
I've got the idea from [this issue](https://github.com/gcanti/fp-ts/issues/1315)
the idea is that I add a function that `sequence`  array to data type, it does not use `ap` so it's stack safe and perform better.
the question that I have is that I don't know if it's name `sequenceA` make sense or not. 🤔
this pull request is draft until I commit all these data types.

- [x] Either

- [x] Option

- [x] Task

- [x] TaskSeq

- [x] IO

- [x] Reader

- [x] State

- [x] TaskEither

- [x] TaskEitherSeq 

- [x] IOEither

- [x] IOEitherSeq

- [x] ReaderTask

- [x] ReaderTaskSeq

- [x] ReaderEither

- [x] ReaderTaskEither

- [x] ReaderTaskEitherSeq

- [x] StateReaderTaskEither

### DIFF
--- a/src/StateReaderTaskEither.ts
+++ b/src/StateReaderTaskEither.ts
@@ -850,3 +850,47 @@ export const apS: <A, N extends string, S, R, E, B>(
 ) => (
   fa: StateReaderTaskEither<S, R, E, A>
 ) => StateReaderTaskEither<S, R, E, { [K in keyof A | N]: K extends keyof A ? A[K] : B }> = apSW
+
+// -------------------------------------------------------------------------------------
+// array utils
+// -------------------------------------------------------------------------------------
+
+/**
+ * @since 2.9.0
+ */
+export const traverseArrayWithIndex: <S, R, E, A, B>(
+  f: (index: number, a: A) => StateReaderTaskEither<S, R, E, B>
+) => (arr: ReadonlyArray<A>) => StateReaderTaskEither<S, R, E, ReadonlyArray<B>> = (f) => (arr) => (s) => (
+  r
+) => async () => {
+  let lastState = s
+  // tslint:disable-next-line: readonly-array
+  const result = []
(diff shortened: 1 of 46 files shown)


mohaalak:
I have another problem for naming I wrote the `sequenceA` for `Task` it's the parallel version, what name should I give to sequential `sequenceSeqA`?

src/StateReaderTaskEither.ts
```diff
@@ -850,3 +850,47 @@ export const apS: <A, N extends string, S, R, E, B>(
...
 ) => StateReaderTaskEither<S, R, E, { [K in keyof A | N]: K extends keyof A ? A[K] : B }> = apSW
+
+// -------------------------------------------------------------------------------------
+// array utils
+// -------------------------------------------------------------------------------------
+
+/**
+ * @since 2.9.0
+ */
+export const traverseArrayWithIndex: <S, R, E, A, B>(
```
DenisFrezzato, line 861: This looks more like `traverseSeqArrayWithIndex`, right? I guess we also want the "parallel" version.

DenisFrezzato (changes requested):

src/StateReaderTaskEither.ts
mohaalak: I don't think it can be done, in State Monad you need the last state from the previous monad
and If you look at this file you can see that `stateReaderTaskEither` and `stateReaderTaskEitherSeq` are the same.

DenisFrezzato (approved):

gcanti:
@mohaalak @iamomiid thank you

PR #1352, opened 2020-11-13
Array chain rec

qlonik:
Addition of two implementations of chainRec (depth-first and breadth-first) for ReadonlyArray and Array.

In `ChainRec` instance by default the depth-first method is selected, but there is access to breadth-first via `BreadthFirstChainRec` instances.

As for the optimization of each of the two algorithms, breadth-first avoids initial array copy by having two loops that do the same thing. In depth-first I couldn't avoid copy of the initial array.

### DIFF
--- a/src/Array.ts
+++ b/src/Array.ts
@@ -6,6 +6,7 @@ import { Alternative1 } from './Alternative'
 import { Applicative as ApplicativeHKT, Applicative1 } from './Applicative'
 import { apFirst as apFirst_, Apply1, apS as apS_, apSecond as apSecond_ } from './Apply'
 import { bind as bind_, Chain1, chainFirst as chainFirst_ } from './Chain'
+import { ChainRec1 } from './ChainRec'
 import { Compactable1 } from './Compactable'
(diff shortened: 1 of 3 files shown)


qlonik:
Friendly bump: is this useful?

gcanti:
@qlonik I'm sorry, somehow I missed this PR, I apologize.

Following the current naming convention (for example `SemigroupSum`, `SemigroupProduct`), I would rename: 

- `ChainRec` to `ChainRecDepthFirst`
- `BreadthFirstChainRec` to `ChainRecBreadthFirst`

(I'll retarget this PR to the `2.11` branch which is dedicated to the next release)

qlonik:
Done. I also rebased it for clean merge. 

In the `Array.ts` module, I'm reusing implementations from `ReadonlyArray.ts`. It used to be like that for other non-pipeable functions, which is why I implemented like this. However, it seems that now they are implemented independently between two modules. So, should I copy implementation from readonly array to regular array?

gcanti:
Thanks @qlonik

PR #1367, opened 2020-12-23
Add flatZip and transpose

cyberixae:
This PR adds two new functions `flatZip`  and `transpose`, that operate on arrays. The `flatZip` function is used for merging several arrays into one by alternating elements from each array. The `transpose` function flips the x and y axis of a two dimensional array. See the [flat-zip](https://github.com/fregante/flat-zip) npm package and Haskell [transpose](http://zvon.org/other/haskell/Outputlist/transpose_f.html) function for a more detailed explanation.


kasamachenkaow:
Hello, transpose seems to be a very useful function, is it possible to continue on this PR?

PR #1368, opened 2020-12-28
Add more `/*#__PURE__*/` comments to improve tree shaking

OliverJAsh:
I inspected my bundle and discovered a few things being bundled which shouldn't have been. These comments fixed that.

There might be more to fix—I just fixed the ones that were causing me problems.

### DIFF
--- a/src/Array.ts
+++ b/src/Array.ts
@@ -1634,7 +1634,7 @@ export const some: <A>(predicate: Predicate<A>) => (as: Array<A>) => boolean = R
 /**
  * @since 2.9.0
  */
-export const Do: Array<{}> = of({})
+export const Do: Array<{}> = /*#__PURE__*/ of({})
 
 /**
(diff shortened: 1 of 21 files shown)


gcanti:
Thanks @OliverJAsh, looks like there's a bug in the typescript emitter, you must put the `/*#__PURE__*/ ` comment in its own line otherwise it doesn't show up in the output file, so

```ts
export const Do: Array<{}> = 
  /*#__PURE__*/ 
  of({})
```

instead of

```ts
export const Do: Array<{}> = /*#__PURE__*/ of({})
```

OliverJAsh:
@gcanti Thanks, fixed!

OliverJAsh:
Related TS bug: https://github.com/microsoft/TypeScript/issues/40943

gcanti (approved):

PR #1370, opened 2020-12-29
Add more `/*#__PURE__*/` comments to improve tree shaking

OliverJAsh:
More of the same (#1368)

@gcanti WDYT about using a tool which adds these automatically? Something like https://github.com/Andarist/babel-plugin-annotate-pure-calls.


gcanti:
@OliverJAsh a linting rule would be nice (but I can't find it)

PR #1373, opened 2021-01-07
Fix documentation typos

mlrv:
Fix two small typos

`this function have` -> `this function has`
`and perform better` -> `and performs better`


samhh:
Someone with a better formal understanding of English than me should verify this, but I also think that "behavior of" in all those examples should actually be "behavior as".

mlrv:
> Someone with a better formal understanding of English than me should verify this, but I also think that "behavior of" in all those examples should actually be "behavior as".

Right, it does sound better to me as well

gcanti:
Thanks @mlrv

PR #1375, opened 2021-01-15
Add eqBigInt and ordBigInt

### LINKED ISSUE #1371: Add ordBigInt
## 🚀 Feature request

Is there currently an ordering for the `bigint` type? I might be missing something

### Current Behavior

There is no provided ordering for `bigint` 

### Desired Behavior

``` typescript
import { clamp, ordBigInt } from 'fp-ts/Ord'

const ten = pipe(
    BigInt(100),
    clamp (ordBigInt) (-10, 10)
) // return 10n
```

### Suggested Solution

In `Ord.ts`:

```typescript
function looseEqual<A>(a: A, b: A): boolean {
    return a == b
}

export const ordBigInt: Ord<BigInt> = {
    equals: looseEqual,
    compare
}
```

This is for fp-ts users who deal with the `bigint` type and any orderings

``` typescript
const ordBigInt = contramap ((a: bigint) => a.toString()) // incorrect
```

``` typescript
const ordBigInt = contramap ((a: bigint) => +a.toString()) // overflow risk
```

### Your environment

| Software   | Version(s) |
| ---------- | ---------- |
| fp-ts      |   2.8.5         |
| TypeScript |   3.9.7      |

EricCrosson:
Closes #1371

### DIFF
--- a/test/Eq.ts
+++ b/test/Eq.ts
@@ -53,6 +53,12 @@ describe('Eq', () => {
     assert.deepStrictEqual(_.eqDate.equals(new Date(1), new Date(0)), false)
   })
 
+  it('eqBigInt', () => {
+    assert.deepStrictEqual(_.eqBigInt.equals(BigInt(0), BigInt(0)), true)
+    assert.deepStrictEqual(_.eqBigInt.equals(BigInt(0), BigInt(1)), false)
+    assert.deepStrictEqual(_.eqBigInt.equals(BigInt(1), BigInt(0)), false)
+  })
+
   it('getStructEq', () => {
     const S = _.getStructEq<Person>({
       name: _.eqString,
--- a/docs/modules/Eq.ts.md
+++ b/docs/modules/Eq.ts.md
@@ -29,6 +29,7 @@ Added in v2.0.0
   - [URI](#uri)
   - [URI (type alias)](#uri-type-alias)
   - [eq](#eq)
+  - [eqBigInt](#eqbigint)
   - [eqBoolean](#eqboolean)
   - [eqDate](#eqdate)
   - [eqNumber](#eqnumber)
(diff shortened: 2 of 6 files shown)


EricCrosson:
Open to feedback :+1:

pashutk:
Hi!

Do you have a reason to use loose equality instead of `strictEqual`?
Seems like `===` works well for bigint equality check.

Screenshot 2021-01-15 at 19 35 32

EricCrosson:
I did witness that too in a test after I put up this PR, but came to this implementation after reading the documentation on [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/BigInt#comparisons). This doc seems to be a superset of the information in the [tc39 bigint proposal](https://github.com/tc39/proposal-bigint#comparisons).

Since `bigint`s are wrapped in objects, this loose equality makes sense from a theoretical level. I would expect the strict equality to return true only when the left and right operand are the same object in memory. I'm not sure how to explain yours and my local test just yet

pashutk:
@EricCrosson 
> I would like some confidence about why the strict equality should be a correct alternative before diverging from the specification.

Maybe this can help: [3.2.6 Strict Equality Comparison](https://tc39.es/proposal-bigint/#sec-strict-equality-comparison)

pashutk:
Also, looks like the link you mentioned as a bigint proposal telling us about comparison between bigint and number, not between two instances of bigint.

test/Eq.ts
```diff
@@ -53,6 +53,12 @@ describe('Eq', () => {
     assert.deepStrictEqual(_.eqDate.equals(new Date(1), new Date(0)), false)
   })
 
+  it('eqDate', () => {
```
DenisFrezzato, line 56:
```suggestion
  it('eqBigInt', () => {
```

EricCrosson:
> Also, looks like the link you mentioned as a bigint proposal telling us about comparison between bigint and number, not between two instances of bigint.

Aha, yes that does clear it up! Thanks for the helpful links @pashutk, I'm much happier to use strict equality

gcanti:
Thanks @EricCrosson @pashutk @DenisFrezzato

PR #1376, opened 2021-01-16
tryCatch impl (rxjs parity)

anthonyjoeseph:
Logically equivalent semantic change in the implementation of `tryCatch` for parity with the rxjs `tryCatch` https://github.com/gcanti/fp-ts-rxjs/pull/46, as per @OliverJAsh suggestion https://github.com/gcanti/fp-ts-rxjs/issues/31#issuecomment-760982846

Uses `promise.then(a).catch(e)` instead of `promise.then(a, e)`

### DIFF
--- a/docs/modules/TaskEither.ts.md
+++ b/docs/modules/TaskEither.ts.md
@@ -533,10 +533,9 @@ Converts a function returning a `Promise` to one returning a `TaskEither`.
 **Signature**
 
 ```ts
-export declare function tryCatchK<E, A extends ReadonlyArray<unknown>, B>(
-  f: (...a: A) => Promise<B>,
-  onRejected: (reason: unknown) => E
-): (...a: A) => TaskEither<E, B>
+export declare function tryCatchK<A extends ReadonlyArray<unknown>, B>(
+  f: (...a: A) => Promise<B>
+): (...a: A) => TaskEither<unknown, B>
 ```
 
 Added in v2.5.0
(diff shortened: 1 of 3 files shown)


gcanti:
> Uses promise.then(a).catch(e) instead of promise.then(a, e)

@anthonyjoeseph What's the motivation? Could you please add a test showing what this change is intended to solve?

anthonyjoeseph:
The idea is parity with the proposed `tryCatch` implementation in for rxjs https://github.com/gcanti/fp-ts-rxjs/pull/48

This is no longer just a semantic change! Now, `tryCatch` and `tryCatchK` no longer accept an `onRejected` function. Tests have been updated to reflect this.

Semantically, `promise.then(E.right).catch(E.left)` is closer to `observable.pipe(catchError(E.left))` than `promise.then(E.right, E.left)`. That was the original purpose, which should have no bearing on tests, since E.right will never throw an error.

mohaalak:
I think it's not a good idea, it's not backward compatible, and you are introducing a breaking change.

anthonyjoeseph:
You are totally right! This belongs in the 3.0.0 branch. Looks like the [breaking change](https://github.com/gcanti/fp-ts/blob/3bba660ce79f77392608d531a7a3cd97e962ce15/src/TaskEither.ts#L130) has already taken place there. My bad, should've looked at that first.

anthonyjoeseph:
new PR w/o breaking change: https://github.com/gcanti/fp-ts/pull/1378

PR #1378, opened 2021-01-17
tryCatch impl (rxjs parity)

anthonyjoeseph:
Logically equivalent semantic change in the implementation of tryCatch for parity with the rxjs `tryCatch` https://github.com/gcanti/fp-ts-rxjs/pull/48 as per @OliverJAsh suggestion https://github.com/gcanti/fp-ts-rxjs/issues/31#issuecomment-760982846

Semantically, `promise.then(a).catch(e)` is closer to `observable.pipe(catchError(e))` than promise.then(a, e). This should have no bearing on tests or behavior, since E.right will never throw an error.

Difference btw/n `then`'s `onRejected` and `catch` documented [here](https://www.codingame.com/playgrounds/347/javascript-promises-mastering-the-asynchronous/the-catch-method), though they are logically equivalent in this case.

### DIFF
--- a/src/TaskEither.ts
+++ b/src/TaskEither.ts
@@ -127,7 +127,7 @@ export const fromEither: FromEither2<URI>['fromEither'] =
  * @category constructors
  * @since 3.0.0
  */
-export const tryCatch = <A>(f: Lazy<Promise<A>>): TaskEither<unknown, A> => () => f().then(E.right, E.left)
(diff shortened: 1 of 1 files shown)


anthonyjoeseph:
It goes without saying, but I'd be perfectly willing to close this PR. I just wanted to be sure to follow up on the discussion we had [here](https://github.com/gcanti/fp-ts-rxjs/issues/31#issuecomment-760982846)

anthonyjoeseph:
Closing

PR #1385, opened 2021-01-27
version 2.10.0

### LINKED ISSUE #1362: Either merge to union function
### Current Behavior

There is no `merge` function or equivalent that I could find.

### Desired Behavior

```typescript
const merge = <E, A>(ma: E.Either<E, A>): E | A => E.fold<E, A, E | A>(identity, identity)(ma);

const e: E.Either<string, number> = E.left("a");
const m: string | number = merge(e)
```

### Suggested Solution

This could probably also apply to other `Either` derivatives.

This would be used by anyone wanting to convert an `Either` type into a standard TypeScript union type.

You can always just write  `E.fold<E, A, E | A>(identity, identity)(ma)` but it would be nice to have a convenience function.

### Your environment

| Software   | Version(s) |
| ---------- | ---------- |
| fp-ts      |       2.9.1     |
| TypeScript |     4.1       |

gcanti:
Version `2.10.0-rc.3` is released, `npm i fp-ts@rc`


gcanti:
FYI just published the release candidate of fp-ts 2.10 (`npm install fp-ts@rc`)

cyberixae:
The new `These` type seems interesting. How does one convert from a `These` to an `Either`? I'm thinking about a situation where non-fatal errors are purposefully made fatal to provide a stricter version of the same operation.

mlegenhausen:
Have tried to integrate the new RC in a bigger code base and the `J.stringify` function exploded in my face. I don't think that it is a good idea to throw at runtime instead of using a `Left`. I would expect to see in the signature the behaviour of the function. When I see an `Either` I expect to get all errors via the `Left`. The Left part is already `unknown` so throwing a new unknown error at runtime instead of simply return a `left(new Error(...))` seem unreasonable. 

I think that would also align with the normal usage when converting the error from `unknown` to some defined error. I normally have an `UnknownError` type as catch all type.

gcanti:
@mlegenhausen it shouldn't definitely happen (the implementation uses a [`tryCatch`](https://github.com/gcanti/fp-ts/blob/020e6a01d4be3823e916cb6714321b06708025d7/src/Either.ts#L1401) internally), could you please provide a repro?

mlegenhausen:
@gcanti never mind, my fault, sorry just read the javascript output in a hurry and overlooked the `tryCatch` block and it looked like a runtime exception in my application.

mlegenhausen:
@gcanti I use the `time` function from `fp-ts-contrib` which required a `MonadIO` instance which is not defined for example on `ReaderTaskEither` how would I use this function with the new granular instances? Currently I can only used the deprecated `readerTaskEither` instance.

gcanti:
@mlegenhausen you are right, looks like the following instances are missing (in all `Task`-based data types):

- `Chain` (with parallel `ap`)
- `Monad` (with parallel `ap`)
- `MonadIO` (with parallel `ap`)
- `MonadTask` (with parallel `ap`)

other?

mlegenhausen:
`MonadThrow` for `TaskEither` and `ReaderTaskEither`?

gcanti:
For what concerns: 

- MonadIO
- MonadTask
- MonadThrow

I would like to ditch them in favour of:

- FromIO
- FromTask
- FromEither

which are future proof.

In `fp-ts-contrib` the only function using `MonadIO` is `time`, I wonder whether I can tweak its definition to make it work with both `MonadIO` and `Chain + FromIO`.

mlegenhausen:
@gcanti Thanks for the fixes!

mlegenhausen:
There is a typo in the Changelog in the `Strong` section you wrote `splitString` instead of `splitStrong`.

PR #1386, opened 2021-01-30
feat: add monoidSumBigInt and monoidProductBigInt

EricCrosson:
The added monoids and semigroups behave the same as monoidSum
and monoidProduct for numbers.

### DIFF
--- a/docs/modules/Monoid.ts.md
+++ b/docs/modules/Monoid.ts.md
@@ -55,8 +55,10 @@ Added in v2.0.0
   - [monoidAll](#monoidall)
   - [monoidAny](#monoidany)
   - [monoidProduct](#monoidproduct)
+  - [monoidProductBigInt](#monoidproductbigint)
   - [monoidString](#monoidstring)
   - [monoidSum](#monoidsum)
+  - [monoidSumBigInt](#monoidsumbigint)
   - [monoidVoid](#monoidvoid)
 - [type classes](#type-classes)
   - [Monoid (interface)](#monoid-interface)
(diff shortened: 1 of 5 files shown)


EricCrosson:
Since these monoid values are constants and not functions, overloading is not an option to provide a consistent interface between the `number` and `bigint` monoids. I resorted to identifiers that include the word `BigInt` in the name.

I'm open to other alternatives, this is just a first take

gcanti:
@EricCrosson I wonder whether this is a breaking change, what if the runtime doesn't support the `BigInt` constructor?

EricCrosson:
Good point. I am not sure which runtimes that would be, but it is good you are thinking about it.

This [MDN compatibility chart](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/BigInt#browser_compatibility) makes it look like every environment but IE is covered. I think this may require `esnext.bigint` to some TypeScript lib configurations though

gcanti:
I think it's better to move bigint support to an external package `fp-ts-bigint`, just to be safe.

EricCrosson:
Here is the separate library: https://github.com/EricCrosson/fp-ts-bigint

gcanti:
Nice, let's move there the `Eq` / `Ord` instances too.

EricCrosson:
:+1: I'll open a PR to delete from fp-ts when ready

gcanti:
> to delete from fp-ts 

No need thank you, I already did it (I'm working on v2.10, see https://github.com/gcanti/fp-ts/pull/1385)

EricCrosson:
Yep, I see it. I like the new layout btw

Will fp-ts 3 include native bigint support?

gcanti:
> Will fp-ts 3 include native bigint support?

I don't think so, I'm fine with your eternal package if it's okay with you

PR #1387, opened 2021-02-02
add Tree.unfoldTreeM and Tree.unfoldForestM

jituanlin:
Migrate `Tree.unfoldTreeM` and `Tree.unfoldForestM` from 2.x


gcanti:
Thanks @jituanlin

PR #1391, opened 2021-02-02
Array, ReadonlyArray: change scanLeft & scanRight to return NonEmpty

willheslam:
Given scanLeft/Right, unlike reduce, always returns an array that has at least one element (the seed),
it makes sense to me to be more specific about the result.

What do you think? Are there any unintended consequences I'm missing here?

### DIFF
--- a/src/ReadonlyArray.ts
+++ b/src/ReadonlyArray.ts
@@ -25,6 +25,7 @@ import { Monad1 } from './Monad'
 import { Monoid } from './Monoid'
 import * as O from './Option'
 import { fromCompare, getMonoid as getOrdMonoid, Ord, ordNumber } from './Ord'
+import { NonEmptyArray } from './NonEmptyArray'
 import { ReadonlyNonEmptyArray } from './ReadonlyNonEmptyArray'
 import { Show } from './Show'
 import { PipeableTraverse1, Traversable1 } from './Traversable'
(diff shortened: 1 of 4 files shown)


gcanti:
> it makes sense to me to be more specific about the result

Thanks @willheslam, I agree.

src/ReadonlyArray.ts
```diff
@@ -299,11 +300,11 @@ export function foldRight<A, B>(
  * @category combinators
  * @since 2.5.0
  */
-export function scanLeft<A, B>(b: B, f: (b: B, a: A) => B): (as: ReadonlyArray<A>) => ReadonlyArray<B> {
+export function scanLeft<A, B>(b: B, f: (b: B, a: A) => B): (as: ReadonlyArray<A>) => ReadonlyNonEmptyArray<B> {
   return (as) => {
     const l = as.length
     // tslint:disable-next-line: readonly-array
```
gcanti, line 306: we can remove this comment I guess

src/ReadonlyArray.ts
```diff
@@ -323,11 +324,11 @@ export function scanLeft<A, B>(b: B, f: (b: B, a: A) => B): (as: ReadonlyArray<A
  * @category combinators
  * @since 2.5.0
  */
-export function scanRight<A, B>(b: B, f: (a: A, b: B) => B): (as: ReadonlyArray<A>) => ReadonlyArray<B> {
+export function scanRight<A, B>(b: B, f: (a: A, b: B) => B): (as: ReadonlyArray<A>) => ReadonlyNonEmptyArray<B> {
   return (as) => {
     const l = as.length
     // tslint:disable-next-line: readonly-array
```
gcanti, line 330: idem

gcanti (changes requested):

src/ReadonlyArray.ts
willheslam: good spot - done!

gcanti (approved):

Issue #1389, opened 2021-02-02
[FR] Functor & Bifunctor instances for Separated


YBogomolov:
### Current Behavior

Currently `Separated` data type from `Compactable` module doesn't have any useful instances — like Functor, Bifunctor, etc.

### Desired Behavior

It would be nice for `Separated` to export instances for Functor and Bifunctor.

### Suggested Solution

`fp-ts` could either:
1. Define and export instances for Functor and Bifunctor for `Separated`. Probably, this will require moving `Separated` to its own module.
2. Change way of defining `Separated` type as just a tuple: `type Separated<L, R> = [L, R]`. This approach will allow reusing instances from `Tuple` module.

### Who does this impact? Who is this for?

Advanced

### Describe alternatives you've considered

My current alternative looks like this file I'm moving between my projects:

```ts
import { pipeable } from "fp-ts/pipeable";
import { Separated } from "fp-ts/Compactable";
import { Bifunctor2 } from "fp-ts/Bifunctor";
import { Functor2 } from "fp-ts/Functor";

export const URI = "Separated";
export type URI = typeof URI;

declare module "fp-ts/HKT" {
    interface URItoKind2<E, A> {
        readonly Separated: Separated<E, A>;
    }
}

export const separated: Bifunctor2<URI> & Functor2<URI> = {
    URI,
    map: <A, B, C>(fa: Separated<A, B>, f: (b: B) => C): Separated<A, C> => ({ ...fa, right: f(fa.right) }),
    mapLeft: <A, B, C>(fa: Separated<A, C>, f: (a: A) => B): Separated<B, C> => ({ ...fa, left: f(fa.left) }),
    bimap: <A, B, C, D>(fa: Separated<A, C>, g: (A: A) => B, f: (c: C) => D): Separated<B, D> =>
        ({ left: g(fa.left), right: f(fa.right) })
};

const { map, mapLeft, bimap } = pipeable(separated);

export {
    map,
    mapLeft,
    bimap
};
```

### Additional context

I would be happy to work on this feature, but I need to discuss the desired design and approach (tuples vs. record encoding of `Separated`).

### Your environment

| Software   | Version(s) |
| ---------- | ---------- |
| fp-ts      |2.9.3|
| TypeScript |4.1.2|

gcanti:
@YBogomolov I'd prefer solution 1 in order to avoid a big breaking change.

PR #1392, opened 2021-02-02
Implement #1389: Functor & Bifunctor instances for Separated

YBogomolov:
Related issue: #1389


src/Compactable.ts
```diff
@@ -32,21 +32,12 @@ import {
...
 // -------------------------------------------------------------------------------------
 // model
 // -------------------------------------------------------------------------------------
 
-/**
- * A `Separated` type which holds `left` and `right` parts.
- *
- * @since 2.0.0
- */
-export interface Separated<A, B> {
```
gcanti, line 45: We can't remove this interface, would be a breaking change. My suggestion is to re-export from the `Separated` module

src/Separated.ts
```diff
@@ -0,0 +1,131 @@
...
+import { Bifunctor2 } from './Bifunctor'
+
+// -------------------------------------------------------------------------------------
+// model
+// -------------------------------------------------------------------------------------
+
+/**
+ * A `Separated` type which holds `left` and `right` parts.
+ *
+ * @since 2.0.0
```
gcanti, line 24: 2.0.0 -> 2.10.0

src/Separated.ts
```diff
@@ -0,0 +1,131 @@
...
+// -------------------------------------------------------------------------------------
+
+/**
+ * `map` can be used to turn functions `(a: A) => B` into functions `(fa: F<A>) => F<B>` whose argument and return types
+ * use the type constructor `F` to represent some computational context.
+ *
+ * @category Functor
+ * @since 2.10.0
+ */
+export const map = <B, C>(f: (b: B) => C) => <A>(fa: Separated<A, B>): Separated<A, C> => ({
```
gcanti, line 65: I'd use the `separated` constructor (everywhere), so `separated(fa.left, f(fa.right))`

src/Separated.ts
```diff
@@ -0,0 +1,131 @@
...
+ * @category constructors
+ * @since 2.10.0
+ */
+export const separated = <A, B>(left: A, right: B): Separated<A, B> => ({ left, right })
+
+// -------------------------------------------------------------------------------------
+// non-pipeables
+// -------------------------------------------------------------------------------------
+
+const map_ = <A, B, C>(fa: Separated<A, B>, f: (b: B) => C): Separated<A, C> => ({ ...fa, right: f(fa.right) })
```
gcanti, line 45: I'd derive `map_` (`mapLeft_`, `bimap_`) from the corresponding pipeable function

```ts
const map_: Functor2<URI>['map'] = (fa, f) => pipe(fa, map(f))
```

gcanti (changes requested):

docs/modules/Compactable.ts.md
```diff
@@ -231,20 +218,29 @@ Added in v2.0.0
...
+Re-exported interface for backwards compatibility.
 
 **Signature**
 
 ```ts
-export interface Separated<A, B> {
-  readonly left: A
-  readonly right: B
-}
+export declare const Separated: any
```
YBogomolov, line 228: @gcanti I haven't found a way to re-export an interface in such way that `docs-ts` mark it as an interface.

src/Compactable.ts
```diff
@@ -200,6 +181,26 @@ export function separate<F, G>(
...
+   * @since 2.10.0
+   */
+  Separated,
+
+  /**
+   * Re-exported constructor for backwards compatibility
+   *
+   * @since 2.10.0
+   */
+  separated
```
gcanti, line 201: this can be removed as was added for 2.10.0

gcanti (changes requested):

docs/modules/Compactable.ts.md
gcanti: Ah that's a pity, let's keep the original definition then (it will be deprecated anyway)

src/Compactable.ts
YBogomolov: I believe this will break quite a few modules, as they are importing `separated` from `Compactable` (Either, Filterable, ReadonlyArray/Map/Set/Record)

gcanti: They should import `separated` from the new `Separated` module

gcanti (approved):

gcanti:
Thank you @YBogomolov 👍

PR #1393, opened 2021-02-05
Add the `flap` function

williamareynolds:
Flap is a convenience that allows functions under structure to be applied to values. It could be compared to `ap`, but with the argument unwrapped. I write this PR after some conversation about having to wrap an argument with `Option` before being able to use `ap`. 

Slack Thread: https://functionalprogramming.slack.com/archives/CPKPCAGP4/p1612453287148200

### DIFF
--- a/docs/modules/Array.ts.md
+++ b/docs/modules/Array.ts.md
@@ -71,6 +71,7 @@ Added in v2.0.0
   - [dropLeftWhile](#dropleftwhile)
   - [dropRight](#dropright)
   - [duplicate](#duplicate)
+  - [flap](#flap)
   - [flatten](#flatten)
   - [intersection](#intersection)
   - [intersperse](#intersperse)
@@ -715,6 +716,16 @@ export declare const duplicate: <A>(wa: A[]) => A[][]
 ```
 
 Added in v2.0.0
+
+## flap
+
+**Signature**
+
+```ts
+export declare const flap: <A>(a: A) => <B>(fab: ((a: A) => B)[]) => B[]
+```
+
+Added in v2.10.0
 
 ## flatten
 
--- a/docs/modules/Const.ts.md
+++ b/docs/modules/Const.ts.md
@@ -25,6 +25,8 @@ Added in v2.0.0
   - [contramap](#contramap)
 - [Functor](#functor)
   - [map](#map)
+- [combinators](#combinators)
+  - [flap](#flap)
 - [constructors](#constructors)
   - [make](#make)
 - [instances](#instances)
@@ -103,6 +105,18 @@ export declare const map: <A, B>(f: (a: A) => B) => <E>(fa: Const<E, A>) => Cons
 ```
 
 Added in v2.0.0
+
+# combinators
+
+## flap
+
+**Signature**
+
+```ts
+export declare const flap: <A>(a: A) => <E, B>(fab: Const<E, (a: A) => B>) => Const<E, B>
+```
+
+Added in v2.10.0
 
 # constructors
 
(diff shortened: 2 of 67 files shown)


williamareynolds:
@gcanti I want to at least include `flap` for `Option` and `Either`. I imagine it might make sense in other structures as well, but I wasn't sure if you'd want those in a different PR, or here. I have to look at the others (`Reader`, `Task`, `State`, etc.) to see if this makes any sense there.

gcanti:
@williamareynolds I don't know if you are comfortable with constrained functions but here's a suggestion: we could define a generic `flap` function in `Functor.ts` constrained by a `Functor` instance

```ts
// Functor.ts

// derivables

/**
 * @category derivables
 * @since 2.10.0
 */
export function flap<F extends URIS4>(
  F: Functor4<F>
): <A>(a: A) => <S, R, E, B>(fab: Kind4<F, S, R, E, (a: A) => B>) => Kind4<F, S, R, E, B>
export function flap<F extends URIS3>(
  F: Functor3<F>
): <A>(a: A) => <R, E, B>(fab: Kind3<F, R, E, (a: A) => B>) => Kind3<F, R, E, B>
export function flap<F extends URIS3, E>(
  F: Functor3C<F, E>
): <A>(a: A) => <R, B>(fab: Kind3<F, R, E, (a: A) => B>) => Kind3<F, R, E, B>
export function flap<F extends URIS2>(
  F: Functor2<F>
): <A>(a: A) => <E, B>(fab: Kind2<F, E, (a: A) => B>) => Kind2<F, E, B>
export function flap<F extends URIS2, E>(
  F: Functor2C<F, E>
): <A>(a: A) => <B>(fab: Kind2<F, E, (a: A) => B>) => Kind2<F, E, B>
export function flap<F extends URIS>(F: Functor1<F>): <A>(a: A) => <B>(fab: Kind<F, (a: A) => B>) => Kind<F, B>
export function flap<F>(F: Functor<F>): <A>(a: A) => <B>(fab: HKT<F, (a: A) => B>) => HKT<F, B>
export function flap<F>(F: Functor<F>): <A>(a: A) => <B>(fab: HKT<F, (a: A) => B>) => HKT<F, B> {
  return (a) => (fab) => F.map(fab, (f) => f(a))
}
```

and then you can use it in each module like this

```ts
// Option.ts

import { flap as flap_ } from './Functor'

/**
 * @category combinators
 * @since 2.10.0
 */
export const flap =
  /*#__PURE__*/
  flap_(Functor)
```

gcanti:
> I don't know if you are comfortable with constrained functions

btw if the answer is no and you are interested, I can explain how this (admittedly verbose) machinery works

> I imagine it might make sense in other structures as well

Theoretically any data structure which admits a `Functor` instance, you can search for the `"export const Functor:"` string in `src`

williamareynolds:
> I don't know if you are comfortable with constrained functions

I actually learned about this library from an article you wrote on building the higher-kinded type functionality into typescript, so this all looks great to me. I knew it didn't belong in the functor instance, but hadn't considered defining the generic function in the Functor module. I'll be happy to do this.

Would it also be desirable to add this to the instance of Functor on these structures? I will admit I haven't worked much with the instances, but I imagine including it there would allow a definition of `flap` to be composed when building a transformer. Is that accurate?

williamareynolds:
> Would it also be desirable to add this to the instance of Functor on these structures? I will admit I haven't worked much with the instances, but I imagine including it there would allow a definition of `flap` to be composed when building a transformer. Is that accurate?

I realized immediately after writing this that another interface would be necessary to provide `flap`. I'm not sure how important it is, and seems like a larger decision.

gcanti:
> Would it also be desirable to add this to the instance of Functor on these structures?

You can't unless you add `flap` to the `Functor` interface itself (but I would keep `Functor` as minimal as possible).

williamareynolds:
> but I would keep `Functor` as minimal as possible

Agreed. I'll make the updates for the generic definition later today. Thank you!

gcanti (approved):

gcanti:
Thanks @williamareynolds

PR #1400, opened 2021-02-12
Simplify implementations for pipe and flow

lionel-rowe:
- [x] Fork [the repository](https://github.com/gcanti/fp-ts) and create your branch from `master`.
- [x] Run `npm install` in the repository root.
- [x] (N/A) If you've fixed a bug or added code that should be tested, add tests!
- [x] Ensure the test suite passes (`npm test`).

This PR simplifies implementations for `pipe` and `flow`. Type signatures can't be simplified due to current TS limitations, but implementations definitely can!

### DIFF
--- a/src/function.ts
+++ b/src/function.ts
@@ -198,54 +198,12 @@ export function flow<A extends ReadonlyArray<unknown>, B, C, D, E, F, G, H, I, J
   hi: (h: H) => I,
   ij: (i: I) => J
 ): (...a: A) => J
-export function flow(
-  ab: Function,
-  bc?: Function,
-  cd?: Function,
-  de?: Function,
-  ef?: Function,
-  fg?: Function,
-  gh?: Function,
-  hi?: Function,
-  ij?: Function
-): unknown {
-  switch (arguments.length) {
-    case 1:
-      return ab
-    case 2:
-      return function (this: unknown) {
-        return bc!(ab.apply(this, arguments))
+export function flow(...fns: ReadonlyArray<Function>): unknown {
+  return fns.length === 0 || fns.length > 9
+    ? undefined
+    : function (this: unknown, ...args: ReadonlyArray<unknown>) {
+        return fns.slice(1).reduce((prev, fn) => fn(prev), fns[0].apply(this, args))
       }
-    case 3:
-      return function (this: unknown) {
-        return cd!(bc!(ab.apply(this, arguments)))
-      }
-    case 4:
-      return function (this: unknown) {
-        return de!(cd!(bc!(ab.apply(this, arguments))))
-      }
-    case 5:
-      return function (this: unknown) {
-        return ef!(de!(cd!(bc!(ab.apply(this, arguments)))))
-      }
-    case 6:
-      return function (this: unknown) {
-        return fg!(ef!(de!(cd!(bc!(ab.apply(this, arguments))))))
-      }
-    case 7:
-      return function (this: unknown) {
-        return gh!(fg!(ef!(de!(cd!(bc!(ab.apply(this, arguments)))))))
-      }
-    case 8:
-      return function (this: unknown) {
-        return hi!(gh!(fg!(ef!(de!(cd!(bc!(ab.apply(this, arguments))))))))
-      }
-    case 9:
-      return function (this: unknown) {
-        return ij!(hi!(gh!(fg!(ef!(de!(cd!(bc!(ab.apply(this, arguments)))))))))
-      }
-  }
-  return
 }
 
 /**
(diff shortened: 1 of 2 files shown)


gcanti:
Thanks @lionel-rowe, could you please run your `flow` implementation against this benchmark? https://github.com/gcanti/fp-ts/blob/d70f487cf84008bef6a7700515210fa85b8fa3c8/perf/function/flow.ts

lionel-rowe:
> Thanks @lionel-rowe, could you please run your `flow` implementation against this benchmark? https://github.com/gcanti/fp-ts/blob/d70f487cf84008bef6a7700515210fa85b8fa3c8/perf/function/flow.ts

Just ran it, `flowWithoutSwitch` is ~1.5× faster than my implementation.

I made a similar benchmark with my `pipe` implementation, the `switch` version is faster at ~53,106,463 ops/sec vs. ~10,692,970 ops/sec for my implementation.

lionel-rowe:
Happy to close if perf is considered more important than file size here, or I can make a separate PR with the `pipe` benchmark for future reference.

gcanti:
@lionel-rowe ok, let's optimize for file size

src/function.ts
```diff
@@ -198,54 +198,12 @@ export function flow<A extends ReadonlyArray<unknown>, B, C, D, E, F, G, H, I, J
...
-    case 1:
-      return ab
-    case 2:
-      return function (this: unknown) {
-        return bc!(ab.apply(this, arguments))
+export function flow(...fns: ReadonlyArray<Function>): unknown {
+  return fns.length === 0 || fns.length > 9
+    ? undefined
+    : function (this: unknown, ...args: ReadonlyArray<unknown>) {
+        return fns.slice(1).reduce((prev, fn) => fn(prev), fns[0].apply(this, args))
```
OliverJAsh, line 205: Something to consider: we used something like this in [`pipe-ts`](https://github.com/unsplash/pipe-ts/blob/920d878046c0e7b719ee9c154b97bff1e2868215/src/index.ts#L68) but the problem we found was that it added a lot of noise to stack traces. Example:

```
Error: Expected Some but got None
  at onNone(./shared/facades/option/operators.ts:31:15)
  at apply(./node_modules/fp-ts/es6/Option.js:144:49)

 ---------  PIPE STARTS -----------------
  at ? (./node_modules/pipe-ts/target/index.js:15:23)
  at Array.reduce(<anonymous>)
  at pipe(./node_modules/pipe-ts/target/index.js:13:20)
  at e.pipeWith(./node_modules/pipe-ts/target/index.js:25:12)
 ---------  PIPE END --------------------
  
  at apply(./shared/facades/option/operators.ts:27:3)
  at ? (./node_modules/pipe-ts/target/index.js:15:23)
  at Array.reduce(<anonymous>)
```

When you use `pipe`/`flow` everywhere this becomes really important, especially if you consider that by default Node limits stack traces to 10 lines.

The existing `switch` implementation doesn't add any uneccessary noise to the stack trace.

We actually switched from `pipe-ts` to using `fp-ts` `pipe`/`flow` for this very reason.

gcanti: @OliverJAsh what about the alternative implementation I'm using in the perf folder?

```ts
export function flow(...fns: ReadonlyArray<Function>): Function | undefined {
  const len = fns.length - 1
  return len > 8
    ? undefined
    : function (this: any, ...x: ReadonlyArray<any>) {
        let y = fns[0].apply(this, x)
        for (let i = 1; i <= len; i++) {
          y = fns[i].call(this, y)
        }
        return y
      }
}
```

the stack trace seems the same as the existing implementation

gcanti:
numbers on my machine

```
flow2 x 4,363,226 ops/sec ±1.69% (87 runs sampled)
flow x 16,030,750 ops/sec ±1.39% (84 runs sampled)

pipe2 x 6,976,069 ops/sec ±0.70% (89 runs sampled)
pipe x 26,323,744 ops/sec ±1.35% (86 runs sampled)
```

using

```ts
export function flow2(...fns: ReadonlyArray<Function>): Function {
  const len = fns.length
  return function (this: any, ...x: ReadonlyArray<any>) {
    let y = fns[0].apply(this, x)
      y = fns[i].call(this, y)
    }
    return y
  }
}

export function pipe2(a: unknown, ...fns: ReadonlyArray<Function>): unknown {
  let out: unknown = a
  for (let i = 0; i < fns.length; i++) {
    out = fns[i](out)
  }
  return out
}
```

@lionel-rowe I'd keep the existing implementations (a 4x performance hit seems too much for a function that is used  everywhere)

src/function.ts
OliverJAsh: Yes that should be fine.

PR #1403, opened 2021-02-20
Add Env comonad

danieljharvey:
Hi!

I've been working through Chris Penner's excellent [Comonads by Example](https://github.com/ChrisPenner/comonads-by-example) and realised that although `Traced` and `Store` are already here, `Env` is not, so thought I'd pop it in.

I have tried to keep everything as consistent with `Store` and similar modules as possible, but happy to make changes.


src/Env.ts
```diff
@@ -0,0 +1,186 @@
...
+import { Comonad2 } from './Comonad'
+import { identity, pipe, flow } from './function'
+import { Functor2 } from './Functor'
+import { Extend2 } from './Extend'
+
+// -------------------------------------------------------------------------------------
+// model
+// -------------------------------------------------------------------------------------
+
+// tslint:disable:readonly-array
```
danieljharvey, line 32: Noticed this is what `Writer` does, as when requiring `readonly` everywhere it it feels a bit awkward to use, happy to change though.

src/Env.ts
```diff
@@ -0,0 +1,186 @@
...
+ */
+export interface Env<E, A> {
+  (): [E, A]
+}
+// tslint:enable:readonly-array
+
+/**
+ * Returns the current environment
+ *
+ * @since 3.0.0
```
danieljharvey, line 45: I'm not sure what version to put here, so guessed an arbitrary one, happy to update.

src/Env.ts
```diff
@@ -0,0 +1,186 @@
...
+}
+
+/**
+ * `extendK` is like `extend` flipped around to make composition easier.
+ * This is like the =<= operator in https://hackage.haskell.org/package/comonad-5.0.8/docs/Control-Comonad.html
+ *
+ * @category Extend
+ * @since 3.0.0
+ */
+export const extendK: <E, B, C>(wbc: (wa: Env<E, B>) => C) => <A>(wab: (wb: Env<E, A>) => B) => (wa: Env<E, A>) => C = (
```
danieljharvey, line 133: This is the only addition that I feel might be contentious - I haven't noticed Kleisli versions of any of the other Comonads. There are tests that use both this and the regular `extend` function side by side - personally using this version feels more intuitive with `pipe` etc.

PR #1405, opened 2021-02-22
Remove matechs-effect from ecosystem.md

pashutk:
Since the matechs library was renamed to effect-ts it's probably a good idea to update links in the ecosystem.md file


mikearnaldi:
I am thinking it would be more appropriate to remove the link, matechs-effect, the previous version was based on the fp-ts typeclasses and it was an extension of the system, effect-ts is completely separated so it would not be correct to advise it as an ecosystem package

pashutk:
Pretty reasonable, I've changed PR so it's removing matechs-effect from ecosystem.md.

@gcanti @mikearnaldi if you agree upon removal then PR is ready to merge, and if you plan to add a "Similar projects" section then close this PR please.

Thanks.

PR #1419, opened 2021-03-03
feat(EitherT): adds bracketT

### LINKED ISSUE #1418: Deriving bracket fromMonadThrow
## 📖 Documentation

How would we derive `bracket` from `MonadThrow`?
I can't see a way to structurally derive it.

The docs say it can be done for `TaskEither`, but the implementation doesn't use it.

Only thing I'd say I could've overlooked is using `ap`, but it's basically just chain, right?

```ts
export interface BracketT<F> {
  <S, A, B>(
    aquire: Lazy<HKT<F, S>>,
    release: R.Reader<S, (fa: HKT<F, A>) => HKT<F, B>>
  ): (use: R.Reader<S, HKT<F, A>>) => HKT<F, B>;
}

export interface BracketT2<F extends URIS2> {
  <S, E, A, B>(
    aquire: Lazy<Kind2<F, E, S>>,
    release: R.Reader<S, (fa: Kind2<F, E, A>) => Kind2<F, E, B>>
  ): (use: R.Reader<S, Kind2<F, E, A>>) => Kind2<F, E, B>;
}

export function bracketT<F>(F: MonadThrow<F>): BracketT<F>;
export function bracketT<F extends URIS2>(F: MonadThrow2<F>): BracketT2<F>;

export function bracketT<F>(F: MonadThrow<F>): BracketT<F> {
  // implementation...
}
```

waynevanson:
Adds `bracketT`, for acquiring and releasing resources that are extended from `Either`, just like the other functions in `EitherT`.

## Summary

- [x] Adds `EitherT.bracketT`
- [x] Adds overload for  `EitherT.BracketT` implementation into `IOEither`
- [x] tests and docs

An edge case where `use` and `release` fail,  but `acquire` succeeds. To account for this, the current implementation is to catch errors into a `ReadonlyArray`, which in this case is always non empty.

### DIFF
--- a/docs/modules/EitherT.ts.md
+++ b/docs/modules/EitherT.ts.md
@@ -21,6 +21,7 @@ Added in v2.0.0
   - [altValidation](#altvalidation)
   - [ap](#ap)
   - [bimap](#bimap)
+  - [bracketT](#brackett)
   - [chain](#chain)
   - [getOrElse](#getorelse)
   - [left](#left)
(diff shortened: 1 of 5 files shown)


waynevanson:
@gcanti No longer a WIP and is now ready for review 👍🏿

gcanti:
@waynevanson

- following the current convention, the name can be just `bracket` IMO
- the implementation looks quite complicated, I'm not sure what are the new overloadings for, I was expecting something like:

```ts
export function bracket<M>(
  M: Monad<M>
): <E, A, B>(
  acquire: HKT<M, E.Either<E, A>>,
  use: (a: A) => HKT<M, E.Either<E, B>>,
  release: (a: A, e: E.Either<E, B>) => HKT<M, E.Either<E, void>>
) => HKT<M, E.Either<E, B>> {
  const left = EitherT.left(M)
  return (acquire, use, release) =>
    M.chain(
      acquire,
      E.fold(left, (a) =>
        M.chain(use(a), (e) =>
          M.chain(
            release(a, e),
            E.match(left, () => M.of(e))
          )
        )
      )
    )
}
```

waynevanson:
Complex because the acquiring and releasing the resource should be independent of using the resource.

Another way to make it simpler is to ignore the fact that `use` and `release` aren't always mutually exclusive in their error handling, and only return the error from `release`.

I'll also change the name to bracket when I am available.

gcanti:
> Another way to make it simpler is to ignore the fact that use and release aren't always mutually exclusive in their error handling, and only return the error from release.

AFAIK there's only one error: the `E` in the return type `HKT<M, E.Either<E, void>>`, here's the logic:

- if `acquire` throws `e1`, rethrow `e1`
- else:
  - if `use` throws `e2`, `release` is called with `a: A` and `left(e2)`, the `release` action can then choose whether rethrow `e2` or not
  - else: 
    `release` is called with `a: A` and `right(b)`

waynevanson:
I see what you mean.

I found that when I want to acquire and release a resource, I want that logic to be separate to how I use it.

In my implementation, the kleisli is separate from acquire and release.
This is useful when the resource might be something `getFreshFileDescriptor`, where the resource is different with each call and I want to map over an array with the function. A simple partial application of the function and it can be used.

 I think a midway solution is to allow the release function to handle the error thrown from use. That way it doesn't have to return and array.

What do you think? I'll change the signature this week (on phone currently)

gcanti:
IIUC you want to change the signature of `bracket` in `EitherT.ts` from:

```ts
export declare function bracket<M>(
  M: Monad<M>
): <E, A, B>(
  acquire: HKT<M, E.Either<E, A>>,
  use: (a: A) => HKT<M, E.Either<E, B>>,
  release: (a: A, e: E.Either<E, B>) => HKT<M, E.Either<E, void>>
) => HKT<M, E.Either<E, B>>
```

to

```ts
export declare function bracket<M>(
  M: Monad<M>
): <E, A, B>(
  acquire: HKT<M, E.Either<E, A>>,
  release: (a: A, e: E.Either<E, B>) => HKT<M, E.Either<E, void>>
) => (use: (a: A) => HKT<M, E.Either<E, B>>) => HKT<M, E.Either<E, B>>
```

right?

waynevanson:
Almost. The release signature could be nicer to work with.

I have the following changes in mind:

- Curried `release` signature
  - Easier to compose. Users can use their modules `chain` or `ap` functions to compose if they feel like it.
- result from `release` is `HKT<M, Either<E,B>` over `Either<E, B>`
  - User would be able to just use the module they got bracket from instead of importing `Either` too
- pipe `bracket` so `use` is at the end.
- Widen the bracket type (or create `bracketW` instead)

```ts
export declare function bracket<M>(
  M: Monad<M>
): (
  acquire: HKT<E, Either<E,A>>,
  release: (a: A) => (result: HKT<M, Either<E, B>>) => HKT<M, Either<E, void>
): (use: (a: A) => HKT<M, Either<E, B>>) =>  HKT<M, Either<E, B>>
```

gcanti:
> Curried release signature. Easier to compose. Users can use their modules chain or ap functions to compose if they feel like it. 

mmhh looks awkward

```ts
declare const bracket2: <E, A, B>(
  acquire: IOEither<E, A>,
  release: (a: A, e: Either<E, B>) => IOEither<E, void>
) => (use: (a: A) => IOEither<E, B>) => IOEither<E, B>

declare const acquire: IOEither<string, number>
declare const release: (a: number) => IOEither<string, void>
declare const use: (a: number) => IOEither<string, boolean>

pipe(use, bracket2(acquire, release))

declare const bracket3: <E, A, B>(
  acquire: IOEither<E, A>,
  release: (a: A) => (e: Either<E, B>) => IOEither<E, void>
) => (use: (a: A) => IOEither<E, B>) => IOEither<E, B>

pipe(
  use,
  bracket3(acquire, (a) => () => release(a))
)
```

> result from release is HKT<M, Either<E,B> over Either<E, B>

again looks awkward, a user can always lift the value if really needed

> pipe bracket so use is at the end

:+1:

> or create bracketW instead

:+1:

waynevanson:
Alright I'll change it to the classic, thanks for consideration.

waynevanson:
I cannot get the piped overload to infer the type from use with release, because release depends on the E of use.

I'm considering closing this as the current implementation is the next best thing.

Issue #1420, opened 2021-03-03
NonEmptyArray and ReadonlyNonEmptyArray are missing functions from Array and ReadonlyArray


jleider:
## 🐛 Bug report

There are a number of functions that exist within `Array` and `ReadonlyArray` that are absent from `NonEmptyArray` and `ReadonlyNonEmptyArray`. 

Some of them may not be applicable to their `NonEmpty` variants or may return an `Option<NonEmptyArray>` or a normal `Array` from a `NonEmptyArray` such as with `filter` style functions. 

At this point its a design decision and I'm not sure if its better to have these `filter` style functions acting directly upon a `NonEmptyArray` which potentially return an `Option<NonEmptyArray>`,  implicitly convert to an `Array`, or if its clearer to explicitly convert a `NonEmptyArray` to a normal `Array` before running `filter` style functions. Ideally converting from a `NonEmptyArray` to `Array` would incur little to no runtime cost. 

That said, a quick glance shows these are the missing functions/instances:

- Compactable
  - compact
  - separate
- Foldable
  - foldMap
- FoldableWithIndex
  - foldMapWithIndex
- FilterableWithIndex 
  - All functions missing
- Traversable
  - All functions missing
- Unfoldable
  - All functions missing
- Witherable
  - All functions missing
- Combinators
  - Lots of inconsistencies here, while the majority of functions exist in both, some only exist in `Array` while others only exist within `NonEmptyArray`
  - `foldMap` and `foldMapWithIndex` listed here instead of under `Foldable` and `FoldableWithIndex`
- Destructors
  - All `find*`, `fold*` and `match*` missing from `NonEmptyArray`
- Instances
  - getMonoid  
  - getOrd
- Unsafe
  - Contains no unsafe methods
- Utils
  - Many util functions missing from `NonEmptyArray`

Also of note, the order of the documentation is different for `Array` vs `NonEmptyArray` which also further complicates comparison between the two different Array types. 

### Your environment

| Software   | Version(s) |
| ---------- | ---------- |
| fp-ts      |       2.9.5     |
| TypeScript |     4.1       |

gcanti:
Thanks @jleider this kind of analysis is gold!

> Ideally converting from a NonEmptyArray to Array would incur little to no runtime cost.

Note that you don't need any conversion, a `NonEmptyArray` is an `Array`, so all `Array` APIs can be used with a `NonEmptyArray` without further ado.

[...]

PR #1421, opened 2021-03-04
version 2.10.0 rc.4

gcanti:
Related: https://github.com/gcanti/fp-ts/issues/1420

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,12 +18,18 @@ high state of flux, you're at risk of it changing without notice.
 # 2.10.0-rc.4
 
 - **Deprecations**
+  - `Array`
+    - deprecate `empty`
+  - `Map`
+    - deprecate `empty`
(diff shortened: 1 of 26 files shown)


SRachamim:
Why did we deprecated the `empty` instances? They were very useful when you want to create a stable reference for an empty array.

```typescript
import * as A from 'fp-ts/ReadonlyArray'
import PureComponent from 'PureComponent'

interface Props<A> {
  readonly as: Array<A>
}
const Component = <A>({ as = A.empty }: Props<A>) => (
  <PureComponent as={as} />
)
```

Using `[]` as recommended is naive and won't work here.

What's the story with removing it? What is the trigger?

gcanti:
@SRachamim `ReadonlyArray`'s `empty` is not deprecated

SRachamim:
Oh wow! That makes sense now!

Greate change! :)

PR #1424, opened 2021-03-10
Assert arrays as non-empty when using some()

thewilkybarkid:
`Array.prototype.some()` will always returns `false` if the array is empty, so can be used as a type guard.

(Refs https://github.com/sciety/sciety/blob/dea3e024b36c6c0c7799dc3376f1930f8f93f114/src/article-page/handle-article-version-errors.ts#L31-L34)


gcanti:
@thewilkybarkid nice! thank you

Issue #1404, opened 2021-02-22
Calling Task.traverseArray on a ReadonlyNonEmptyArray returns a ReadonlyArray


thewilkybarkid:
### Current Behavior

Calling `Task.traverseArray` on a `ReadonlyNonEmptyArray` returns a `ReadonlyArray`.

### Expected behaviour

That it returns a `ReadonlyNonEmptyArray`.

### Reproducible example

```typescript
import { readonlyNonEmptyArray as RNEA, task as T, function as f } from 'fp-ts';

declare const rats: ReadonlyArray<T.Task<string>>
declare const rneats: RNEA.ReadonlyNonEmptyArray<T.Task<string>>

const ra: T.Task<ReadonlyArray<string>> = f.pipe(rats, T.sequenceArray)
const rnea: T.Task<RNEA.ReadonlyNonEmptyArray<string>> = f.pipe(rneats, T.sequenceArray)
```

[TS Playground](https://www.typescriptlang.org/play?#code/JYWwDg9gTgLgBAbzlApgQwCYQHYBsCeAcjgKLgz4CCUUa+caAznAEqEmUA0cMTA1g2YAVbgDMArtgDGMYDkFxRcAL6KoEEHADkosAFoYjLQG4AUKYwopuNKjhScjeLUMAuVuix4qNOgB4hADohfj8nKGBsAHMAPhiLKxs7B2wnZGx0N1Z2SkCWTxwCYmwyMApqWnwA4NDwyNj40xS02ncgkMY+P3zMQp9KsJgI6Li4AF5FQLBgMBQAChdGbiDGFABHcRRpFAq6AEomx2cMtDaazu6cvILvYtLy3yq6kZjxyenZhZPDZcDVja2Uh2jwOQA); [Original source](https://github.com/sciety/sciety/blob/85185d93476ac0519b2e3ecc27bcfe5042d038f9/src/home-page/render-editorial-communities.ts#L30-L33)

### Suggested solution(s)

Overloading if possible (don't think generics would work); otherwise separate functions?

### Additional context

Applies to various other functions where the length of the array doesn't change (eg `T.sequenceArray`).

I've had a quick go with first generics then overloading, but seems a bit more complicated than hoped.

### Your environment

| Software   | Version(s) |
| ---------- | ---------- |
| fp-ts      | 2.9.5           |
| TypeScript |  4.1.5          |

gcanti:
[...]

PR #1436, opened 2021-03-22
Add appendW, prependW and concatW to arrays

thewilkybarkid:
Adds `appendW` and `prependW` to `Array` and `ReadonlyArray`, and `concatW` to `NonEmptyArray` and `ReadonlyNonEmptyArray`.

PR #1439, opened 2021-03-22
Return non-empty arrays where possible

thewilkybarkid:
For #1404.

I've started with the `Task` cases for feedback; there are a lot more modules that can also be updated.

### DIFF
--- a/docs/modules/Task.ts.md
+++ b/docs/modules/Task.ts.md
@@ -595,7 +595,9 @@ Equivalent to `ReadonlyArray#sequence(ApplicativePar)`.
 **Signature**
 
 ```ts
-export declare const sequenceArray: <A>(arr: readonly Task<A>[]) => Task<readonly A[]>
+export declare const sequenceArray: <A, B>(
+  arr: readonly Task<A>[] & { readonly 0?: Task<B> | undefined }
+) => Task<Task<B> extends Task<A> ? ReadonlyNonEmptyArray<A> : readonly A[]>
 ```
 
 Added in v2.9.0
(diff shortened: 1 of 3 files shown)


docs/modules/Task.ts.md
```diff
@@ -595,7 +595,9 @@ Equivalent to `ReadonlyArray#sequence(ApplicativePar)`.
 **Signature**
 
 ```ts
-export declare const sequenceArray: <A>(arr: readonly Task<A>[]) => Task<readonly A[]>
+export declare const sequenceArray: <A, B>(
+  arr: readonly Task<A>[] & { readonly 0?: Task<B> | undefined }
+) => Task<Task<B> extends Task<A> ? ReadonlyNonEmptyArray<A> : readonly A[]>
```
thewilkybarkid, line 600: This signature is probably confusing to read...

gcanti: Yeah, I prefer two distinct functions with clean signatures

thewilkybarkid:
I think the type problems are the same as https://github.com/gcanti/fp-ts/pull/1436#discussion_r599383444. I'll have another look then see what happens when applying it to v3, which I suspect would work.

If that's the case, I'll probably reopen this with a nicer PR to v3, then decide whether it's worth adding some temporary functions to v2.

thewilkybarkid:
Just seen that the separate functions exist in v3: https://github.com/gcanti/fp-ts/blob/cf9f33bd512382c04d9b388c870e8018da48ffeb/src/Task.ts#L414-L433

Guess backporting is preferred?

gcanti:
@thewilkybarkid yes, I'm still not sure about the names though

thewilkybarkid:
👍  I'll close this then.

PR #1444, opened 2021-03-25
update Apply example

christiantakle:
Make the example slightly more declarative.

The current use of the `ternary operator` is equal to `String(x)` or `x.toString()`

Node repl session
```
> c = true
> (c ? 'true' : 'false') === String(c)
true

> c = false
> (c ? 'true' : 'false') === String(c)
true
```

Note: I changed the source example and ran `npm run docs`


gcanti:
Thanks @christiantakle

Issue #1360, opened 2020-12-04
[Feature Request] Error Handlers for Map Left


ryanleecode:
### Current Behavior

If I have a `TaskEither` and I want to handle an error in `mapLeft` with another `TaskEither`, I can't without adding my own utility function.

### Desired Behavior

Add a utility function to TaskEither and Either that lets you handle `left` with another TaskEither/Either.

### Suggested Solution

Add a function like this to `TaskEither`

```ts
const handleError = <E, U>(fe: (e: E) => TE.TaskEither<E, U>) => <A>(
  te: TE.TaskEither<E, A>,
): TE.TaskEither<E, A> => {
  return pipe(
    TE.taskEither.map(te, TE.right),
    TE.swap,
    TE.chainFirstW(flow(fe, TE.mapLeft(TE.left))),
    TE.swap,
    TE.flatten,
  )
}
```

#### Usage

```ts
pipe(
  TE.right(123),
  TE.chain(() => TE.left(new Error('oh no!'))),
  handleError(flow(Console.error, (io) => TE.fromIO(io))),
)
```

### Who does this impact? Who is this for?

TypeScript users

### Your environment

| Software   | Version(s) |
| ---------- | ---------- |
| fp-ts      |       2.9.0     |
| TypeScript |      4.1.2      |

gcanti:
@ryanleecode looks like `orElse`

ryanleecode:
Yeah but `orElse` takes the return value of my error handler which I don't want. I need something like a `orElseFirst` if that makes sense.

ryanleecode:
This doesn't compile

```ts
pipe(
  TE.right(123),
  TE.chain(() => TE.left(new Error('oh no!'))),
  TE.orElse(flow(Console.error, TE.fromIO)),
  TE.map((x: number) => `${x}`),
)
```

but this would

```ts
pipe(
  TE.right(123),
  TE.chain(() => TE.left(new Error('oh no!'))),
  handleError(flow(Console.error, TE.fromIO)),
  TE.map((x: number) => `${x}`),
)

```

mohaalak:
When you write `TE.chain(() => TE.left(new Error("oh no!'))` you are changing your data, you no longer have 123 in right position you have a left Error, so when you want to `handleError` you should return a new number from it.

```
pipe(
  TE.right(123),
  TE.chain(() => TE.left(new Error("oh no!"))),
  handleError(flow(Console.error, (io) => TE.fromIO(io))),
[...]

PR #1445, opened 2021-03-26
Add swapped/swappedW to TaskEither

thewilkybarkid:
Similar to https://github.com/gcanti/fp-ts/issues/1360, I've been looking for a good way to be able to write log messages while on the `left` side of an `Either`. `chainFirstIOK` will be perfect for this, but there's no equivalent for the left side (where more logging will happen). Currently, I'm [using `swap` twice](https://github.com/sciety/sciety/blob/1b8e0ca613deb81bad019c04ce878ff418b68c19/src/infrastructure/fetch-static-file.ts#L14-L21) to allow the use of `chainFirst` etc (as opposed to `mapLeft` and returning the original value).

This adds a `swapped`/`swappedW` function which applies a function with the `left` and `right` sides swapped, then swaps them back again.

Leaving this as WIP for feedback, in case there's a better approach, name etc. Would need rolling out to the other `Either` categories too.

### DIFF
--- a/docs/modules/TaskEither.ts.md
+++ b/docs/modules/TaskEither.ts.md
@@ -63,6 +63,8 @@ Added in v2.0.0
   - [orElse](#orelse)
   - [orElseW](#orelsew)
   - [swap](#swap)
+  - [swapped](#swapped)
+  - [swappedW](#swappedw)
 - [constructors](#constructors)
   - [fromEither](#fromeither)
   - [fromIO](#fromio)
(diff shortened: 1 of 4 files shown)


gcanti:
@thewilkybarkid my plan would be to backport [these](https://github.com/gcanti/fp-ts/blob/f7ae980afcab8f68d46b801f35462e644a64a65c/src/EitherT.ts#L485-L550) two helpers from v3

thewilkybarkid:
Ah, brillant. I really should look through the v3 branch more. 😅 

Happy to do the leg work if you like.

gcanti:
@thewilkybarkid thanks, I made public my todo list for v2.11 (https://github.com/gcanti/fp-ts/issues/1446)

thewilkybarkid:
Closing in favour of #1463.

PR #1447, opened 2021-03-27
Make 3.0.0 snapshots installable via git

thought2:
Currently it's hard to use an unpublished snapshot from the `3.0.0` branch inside inside a npm project. Installing `fp-ts` as git dependency does not work.

The reason for this is that the build is run only from the `release` script. When installing an npm dependency via git (`npm install github:<project>/<repo>#<ref>`) a [prepare script](https://docs.npmjs.com/cli/v7/using-npm/scripts#life-cycle-scripts) can be used to build the package on the user's side.

This PR adds a `prepare` script that mimics the publishing process.

Here are the steps to try it out:

```sh
mkdir my-project
cd my-project
npm init -y
npm install github:no-day/fp-ts#feature/3.0.0-git-installable
```

in a source file you can now use new features like:
```
import * as J from "fp-ts/Json
```

It's hard to test, but I'm quite sure that it does not interfere with other tasks.

Note, reasons unknown to me, this does not work with `yarn` yet. They seem to ignore the `prepare` script. Maybe I'll find out later or someone else knows.


thought2:
Ohhh, now I see that we can use `fp-ts@next`.
Well, then I think we don't need this PR.

PR #1451, opened 2021-03-27
Expand struct Module

anthonyjoeseph:
This PR adds a few functions to the `struct` module
 
Notes:
- `traverseS_` is an un-curried version of `traverseS` where the transformation struct can be defined separately. It's named after `Foldable.traverse_`, which looks a bit similar. We can remove this if desired, since `Foldable.traverse_` is deprecated in 3.0.0, but I thought it might be useful since `traverseS` won't work outside of `pipe`
- some functions accept an `Ord<string>` for handling keys, following the [convention in 3.0.0](https://github.com/gcanti/fp-ts/commit/3861138ef9ce6fed494191a42a3eea1884e9c52b#diff-2a28f9bcbc06ac11c7768967193244d8f9ca05a4e96225ca8449930aa26870b0)
- `sequenceS` is omitted, again [mirroring 3.0.0](https://github.com/gcanti/fp-ts/commit/0b76cec70d1b155363ca602138d19e08d4163911#diff-07b5152745a16434cfad0bfbf677f4f684c84a4bf2142f8f6c1895001484dc9a)
- `mapS` is [evolve](https://github.com/gcanti/fp-ts/blob/f1884ce7f064bc62002acbacd6bc51019934771e/src/struct.ts#L57) - I'm certainly willing to rename/move it, but to me `mapS` seems like it fits the conventions of this module
- `Show.struct`, `Eq.struct`, `Semigroup.struct` & `Monoid.struct` have all been relocated into `Struct` - I hope this is OK, since they're not in a release yet (only release candidates)
- Most importantly - `Struct` is not a HKT, so it doesn't necessarily fit the ethos of fp-ts. It's _almost_ a member of many typeclasses (Functor, Traversable, Witherable), but it doesn't fit any of the traditional function signatures. For this reason, it might not be a great fit. However, the idea of a `struct` already exists in the library - this module simply consolidates all its relevant functions and adds to them. I think a `Struct` module makes sense as part of fp-ts but I think it could be happy living elsewhere if that's best

Questions:
- Atm this PR is against `master`. Should it be against `3.0.0`? It requires at least Typescript version 3.6 which might be reason enough

Ideas:
- we could add [pick & omit](https://github.com/samhh/fp-ts-std/blob/e5f465778d984f6618d92b3ac419cae8e03551a7/src/Record.ts#L58) from `fp-ts-std`

### DIFF
--- a/docs/modules/Eq.ts.md
+++ b/docs/modules/Eq.ts.md
@@ -23,7 +23,6 @@ Added in v2.0.0
 - [Contravariant](#contravariant)
   - [contramap](#contramap)
 - [combinators](#combinators)
-  - [struct](#struct)
   - [tuple](#tuple)
   - [~~getStructEq~~](#getstructeq)
   - [~~getTupleEq~~](#gettupleeq)
@@ -62,16 +61,6 @@ Added in v2.0.0
 
 # combinators
 
-## struct
-
-**Signature**
-
-```ts
-export declare const struct: <A>(eqs: { [K in keyof A]: Eq<A[K]> }) => Eq<{ readonly [K in keyof A]: A[K] }>
-```
-
-Added in v2.10.0
-
 ## tuple
 
 Given a tuple of `Eq`s returns a `Eq` for the tuple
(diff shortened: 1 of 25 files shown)


anthonyjoeseph:
Closing in favor of https://github.com/gcanti/fp-ts/pull/1456 and https://github.com/gcanti/fp-ts/pull/1457

PR #1454, opened 2021-03-28
Add object.prop function

cdimitroulas:
New feature for `object` module:
- added `prop` function 

Reasoning: Small and handy utility function which is also something that many folks in the JS ecosystem are used to and learn when getting started with functional programming. Usually they use something like `get` from lodash or they see the `prop` function from the Mostly Adequate Guide to FP (https://mostly-adequate.gitbook.io/mostly-adequate-guide/appendix_c#prop).

**Question**: Do you prefer to add overloads so that we can have both curried/uncurried versions of this? (e.g. `prop('myKey', myObject)` or `prop('myKey')(myObject)`)

### DIFF
--- a/docs/modules/struct.ts.md
+++ b/docs/modules/struct.ts.md
@@ -89,8 +89,11 @@ Accesses a property of an object
 **Signature**
 
 ```ts
-export declare function prop<Path extends string, Value, Rest>(path: Path, obj: Record<Path, Value> & Rest): Value
-export declare function prop<Path extends string, Value, Rest>(path: Path): (obj: Record<Path, Value> & Rest) => Value
+export declare const prop: <Path extends keyof A extends never ? string : keyof A, A>(
+  path: Path
+) => <B extends { [k in Path]: unknown }>(
+  obj: keyof A extends never ? B : A
+) => Path extends keyof A ? A[Path] : B[Path]
 ```
 
 **Example**
--- a/docs/modules/struct.ts.md
+++ b/docs/modules/struct.ts.md
@@ -89,7 +89,8 @@ Accesses a property of an object
 **Signature**
 
 ```ts
-export declare const prop: <Path extends string>(path: Path) => <Value, Rest>(obj: Record<Path, Value> & Rest) => Value
+export declare function prop<Path extends string, Value, Rest>(path: Path, obj: Record<Path, Value> & Rest): Value
+export declare function prop<Path extends string, Value, Rest>(path: Path): (obj: Record<Path, Value> & Rest) => Value
 ```
 
 **Example**
--- a/docs/modules/struct.ts.md
+++ b/docs/modules/struct.ts.md
@@ -16,6 +16,7 @@ Added in v2.10.0
   - [getAssignSemigroup](#getassignsemigroup)
 - [utils](#utils)
   - [evolve](#evolve)
+  - [prop](#prop)
 
 ---
 
@@ -77,6 +78,34 @@ assert.deepStrictEqual(
   ),
   { a: 1, b: 2 }
 )
+```
+
+Added in v2.11.0
+
+## prop
+
+Accesses a property of an object
+
+**Signature**
+
+```ts
+export declare const prop: <Path extends string>(path: Path) => <Value, Rest>(obj: Record<Path, Value> & Rest) => Value
+```
+
+**Example**
+
+```ts
+import { pipe } from 'fp-ts/function'
+import { prop } from 'fp-ts/struct'
+
+type Person = {
+  readonly name: string
+  readonly age: number
+}
+
+const person: Person = { name: 'Jane', age: 62 }
+assert.deepStrictEqual(prop('name')(person), 'Jane')
+assert.deepStrictEqual(pipe(person, prop('age')), 62)
 ```
 
 Added in v2.11.0
--- a/src/struct.ts
+++ b/src/struct.ts
@@ -63,3 +63,24 @@ export const evolve = <A, F extends { [K in keyof A]: (a: A[K]) => unknown }>(tr
   }
   return out as any
 }
+
+/**
+ * Accesses a property of an object
+ *
+ * @example
+ * import { pipe } from 'fp-ts/function'
+ * import { prop } from 'fp-ts/struct'
+ *
+ * type Person = {
+ *   readonly name: string
+ *   readonly age: number
+ * }
+ *
+ * const person: Person = { name: 'Jane', age: 62 }
+ * assert.deepStrictEqual(prop('name')(person), 'Jane')
+ * assert.deepStrictEqual(pipe(person, prop('age')), 62)
+ *
+ * @since 2.11.0
+ */
+export const prop = <Path extends string>(path: Path) => <Value, Rest>(obj: Record<Path, Value> & Rest): Value =>
+  obj[path]
--- a/src/struct.ts
+++ b/src/struct.ts
@@ -82,12 +82,8 @@ export const evolve = <A, F extends { [K in keyof A]: (a: A[K]) => unknown }>(tr
  *
  * @since 2.11.0
  */
-export function prop<Path extends string, Value, Rest>(path: Path, obj: Record<Path, Value> & Rest): Value
-export function prop<Path extends string, Value, Rest>(path: Path): (obj: Record<Path, Value> & Rest) => Value
-export function prop<Path extends string, Value, Rest>(path: Path, obj?: Record<Path, Value & Rest>) {
-  if (typeof obj === 'object') {
-    return obj[path]
-  }
-
-  return (obj: Record<Path, Value> & Rest) => obj[path]
-}
+export const prop = <Path extends keyof A extends never ? string : keyof A, A>(path: Path) => <
+  B extends { [k in Path]: unknown }
+>(
+  obj: keyof A extends never ? B : A
+) => (obj as B)[path] as Path extends keyof A ? A[Path] : B[Path]
--- a/src/struct.ts
+++ b/src/struct.ts
@@ -82,5 +82,12 @@ export const evolve = <A, F extends { [K in keyof A]: (a: A[K]) => unknown }>(tr
  *
  * @since 2.11.0
  */
-export const prop = <Path extends string>(path: Path) => <Value, Rest>(obj: Record<Path, Value> & Rest): Value =>
-  obj[path]
+export function prop<Path extends string, Value, Rest>(path: Path, obj: Record<Path, Value> & Rest): Value
+export function prop<Path extends string, Value, Rest>(path: Path): (obj: Record<Path, Value> & Rest) => Value
+export function prop<Path extends string, Value, Rest>(path: Path, obj?: Record<Path, Value & Rest>) {
+  if (typeof obj === 'object') {
+    return obj[path]
+  }
+
+  return (obj: Record<Path, Value> & Rest) => obj[path]
+}
--- a/test/struct.ts
+++ b/test/struct.ts
@@ -36,4 +36,15 @@ describe('struct', () => {
     x.b = 1
     U.deepStrictEqual(pipe(x, _.evolve({ b: (b) => b > 0 })), { b: true })
   })
+
+  it('prop', () => {
+    interface Person {
+      readonly name: string
+      readonly age: number
+    }
+
+    const person: Person = { name: 'Jane', age: 62 }
+    U.deepStrictEqual(_.prop('name')(person), 'Jane')
+    U.deepStrictEqual(pipe(person, _.prop('age')), 62)
+  })
 })


gcanti:
Thanks @cdimitroulas, two questions:

- what's the purpose of `Rest`?
- is there a way to improve the DX? In the following two use cases typescript doesn't provide any suggestion:

```ts
import { flow, pipe } from 'fp-ts/function'

export declare const prop: <Path extends string>(path: Path) => <Value, Rest>(obj: Record<Path, Value> & Rest) => Value

interface Person {
  readonly name: string
  readonly age: number
}

declare const p: Person

declare const f: () => Person

// use case 1
const g = flow(f, prop('')) // <= no suggestion, it would be nice to get "name" and "age" as suggestions
// use case 3
export const name = pipe(p, prop('')) // <= idem
```

cdimitroulas:
The `Rest` property is needed otherwise in some cases the compiler won't let you pass an object with additional properties:

For example, if we pass the object directly instead of explicitly defining it as the `Person` interface before I get this type error if I remove the `Rest` from the type signature:
```ts
prop('name')({ name: "hello", age: 62 })
// Argument of type '{ name: string; age: number; }' is not assignable to parameter of type 'Record<"name", string>'.
// Object literal may only specify known properties, and 'age' does not exist in type 'Record<"name", string>'.
```

I'll play around with it to see if we can get autocompletion for the key argument to improve the DX, though I'm not entirely sure how to achieve it

thewilkybarkid:
Ignoring the uncurried side for the minute:

```ts
const prop = <Obj, Path extends keyof Obj>(path: Path) => (obj: Obj): Obj[Path] => obj[path];
```

seems close, as autocompletion on the path exists. It does mean having to duplicate the property name when not actually passing in an object, unfortunately:

```ts
const getStatusText = prop<AxiosResponse, 'statusText'>('statusText');
```

cdimitroulas:
Thanks for looking into it @thewilkybarkid 

It seems like we need to make a tradeoff here. Your solution helps to autocomplete the path (though when I tried it, it doesn't list all the possible keys for me when I open the quotes, but it does autocomplete when I start typing e.g. when I type "n" it autocompletes "name" for me). On the other hand it makes it more awkward to reuse the partially applied function for different objects because the `Obj` generic is fixed after the first argument (the key) is passed to `prop`.

For example:
```
const prop = <Obj, Path extends keyof Obj>(path: Path) => (obj: Obj): Obj[Path] => obj[path];

interface Person {
  readonly name: string
  readonly age: number
}

const getName = prop<Person, "name">("name")

// This no longer works because it requires me to pass in a `Person`
getName({ name: "hello" })
```

samhh:
Until the type system improves might we best solve this with two different functions?:

```ts
// not checked yet, I'm saying accept anything with a 'k' property
prop('k')

// checked immediately against the provided generic
propFrom<T>()('k')
```

thewilkybarkid:
> Thanks for looking into it @thewilkybarkid
> 
> It seems like we need to make a tradeoff here. Your solution helps to autocomplete the path (though when I tried it, it doesn't list all the possible keys for me when I open the quotes, but it does autocomplete when I start typing e.g. when I type "n" it autocompletes "name" for me). On the other hand it makes it more awkward to reuse the partially applied function for different objects because the `Obj` generic is fixed after the first argument (the key) is passed to `prop`.

You could make the two object types separate:

```ts
const prop = <Obj, Path extends keyof Obj>(path: Path) => <RealObj extends Record<Path, unknown>>(obj: RealObj): RealObj[Path] => obj[path];
```

This means that the type of the property can be different. If not desired, that type can be reused:

```ts
const prop = <Obj, Path extends keyof Obj>(path: Path) => <Value extends Obj[Path], RealObj extends Record<Path, Value>>(obj: RealObj): Value => obj[path];
```

But not sure if you should be able to pass in a partial object anyhow. (Sounds like the original generic object should be smaller.)

cdimitroulas:
Neither of those two implementations work for me:

Screenshot from 2021-03-31 17-32-54

anthonyjoeseph:
Could something like [this](https://www.typescriptlang.org/play?#code/KYDwDg9gTgLgBAYwgOwM7zASzMOBeOAHgEEAaOAIQD4AKAQwC44y46AjJ+p4gSnyso8mFfqzb0eAKEmhIsOADMArsgQxMKOGCgQwJcgAU6MABZxQMYMgAmqOAGtgATwgLmtMMZNMjpoXBoINgArbj48AWIAbV8TAF0ZcGh4ZVV1TW1dQljzEEsbO3QoTGQAc3JiXPzbOABvOCiAaTgSuFi4phV7ZAgAd2Q4AF8PLx8vciDggH4w7hivBNlkxRU1DQHMsBoAOl26KFLUJiiAJWAkKGtCIpLyuC6e-qo4uAAfBpuy8jOLq8+7h59ZDPfyA-p1SRwFpuegHVDbAA2VlKpnweAIACY+LVIVC4FBgDAlFABvtDlEAIxxKJk1BRAAMcQSUMG0ihBKJJICjicTH+4QEtIZ1J5CVZkiQaHgjDgyCUAFs2MAoPgtNhgDRajKAMza8gcOAAcm1hsG5C0Oi2hrohp4Ukl6DgBrliuVqs2NGthvI9R1eqdTENGMNQz4AHow0QALRRxDEgnIeCYeVgJHyqwwYzrJ0Euj2OwmZXACUoR0ITrIbpA92Wz028ORwgxuCmTCFYDAeV2CD2FsQODpoA) work? 

I'm not sure this is possible in typescript ([source](https://github.com/microsoft/TypeScript/issues/37181)):

```ts
const a: <A extends Record<'a', unknown>>(a: A) => A['a']  = prop('a')
```

The current implementation yields this, which seems less useful:
```ts
const a: (a: Record<'a', unknown>) => unknown  = prop('a')
```

thewilkybarkid:
> Neither of those two implementations work for me:

Sorry, meant need to combine with https://github.com/gcanti/fp-ts/pull/1454#issuecomment-811109210. Think this covers all the cases:

```ts
declare function prop<Obj, Path extends keyof Obj>(path: Path): {
  <RealObj extends Record<Path, unknown>>(obj: RealObj): RealObj[Path]
  (obj: Obj): Obj[Path]
}
declare function prop<Path extends string>(path: Path): <Obj extends Record<Path, unknown>>(obj: Obj) => Obj[Path]

declare function propW<Obj, Path extends keyof Obj>(path: Path): {
  <RealObj extends Record<Path, unknown>>(obj: RealObj): RealObj[Path]
}
declare function propW<Path extends string>(path: Path): <Obj extends Record<Path, unknown>>(obj: Obj) => Obj[Path]
```

[Try this playground](https://www.typescriptlang.org/play?ssl=3&ssc=1&pln=12&pc=116#code/JYWwDg9gTgLgBAbzmYYCmcC+cBmUIhwDkOYAtDAM4D0ANsAEbU4CuAdgMYzARtEDcAKEEATNB1oBDKBladuvZPjAAeAPIMAVgBo4ABUkwAFnDQAPGGjYjKcANZoAnhBxwNmgHwAKMIaMAufT8ASkCEQTg4FQAlNElad1MLKxs4WI5oERUDY112OzYIAHc2D28ILUDY+PdQtLiErQBtHKMAXQi4LwrNQNq+5taOzFFxKRlcdi4eNiUIVVaky2tbShgoYDYAc29fY0DWuvUtJZTbdMzsvzy2AuLS8sq3LWC4AF4PZ80Wvw7RiWksimClmYGUAHVjjogsZTit7E4XF9dn4DiEwp0Yg1EuZlqkLlAsq0bncSmVuk9qo1NHUqe4fsZhv9xkD5DM5mBIYtcWc4GsNtsUfsYUYjjjkvCCUTrnB8oUyY9el9Xh8vgz2sIYI50Po0FBKIo3ohOmxJCA0IF+Zstp1JFsLXA2CwQAw9YIRszAXAMmw1sgDnqDWxPRMfX7OWFHWaHU6XXqsMIw-B7TA9IHeAA5aPvDnZdNsXREU3mojeIvRojBQRJuAptP6zPR8E5sHzLn5wvFtClrzlktVmsprPmlvKXtdyvV3h+odN0dt8cVgfT5NoGAAMQgEHnYF7OC3k8Ha83EGbRtbnL3B6rU99yYAjJb1taW6g0D5dBfF-3l3fawAmJ8BS2V90A-Ws13rINhzQX8ZwAZiAl9zzfcDZ3NODkwAFiQ7ZQPfMBPwhb9u2CTDawAVlwkCULAwiINTfMYPBcitgANmo-C0LXZjWIAdkCWNXSgLjOSIhc+1I1iAA5BOdYTRPBXQ6yYptWIATjkuMRNogilIY3jBBQOjlOPLdXmoagknQLg0BEXBJGAWgWBkIzUPolMTxYuBLOs8RLHsnBHOc1ygA)

cdimitroulas:
>Could something like this work?

@anthonyjoeseph I personally find it weird that it infers as `unknown` when you write something like `prop('a')` but maybe that's just me :D

@thewilkybarkid what's `propW` supposed to do beyond `prop`? In your playground link in `Edit 2` it seems identical to `prop`. Thanks for all the effort to find the best iteration of this function btw!

cdimitroulas:
Nice! The only issue I see with that is that we default to `any` for the `Path` generic, which means that we lose type safety.

For example, this compiles even though the key does not exist in `Person`:
```ts
const getPersonName = prop<Person>('banana')
```

cdimitroulas:
Nice one, this looks quite good in terms of how well it works with the various use-cases.

It's a bit scary that we need to use two assertions to `any` (`(obj as any)[path as any])` :scream: ) but maybe it's fine as I couldn't find any case where this actually impacted on type-safety with a few attempts that I made.

anthonyjoeseph:
[That can be solved](https://www.typescriptlang.org/play?#code/KYDwDg9gTgLgBAYwgOwM7zASzMOBeOAHgEEAaOAIQD4AKAQwC44y46AjJ+p4gSnyso8mFfqzb0eAKElI0GKBDD4ikuHAAKdGAAs4oGMGQATVHBoBrYAE8IAM2Z6QB46eTAAbsChwA-HHRQmMgA5nBMljb2vOSqzJK0YFraTJo6fHgChLEi+oYmcADecADa5nBBGkkAukwArsjmyBAA7shwAL7xNLEQbABW4dZ2DrkucG6e3n4i3JLpAjS9faymFDzFiTpVK5U6js75EcPEvszFqdrbwufV0rLorMpYODQFjHAAzOQccADkH792uQwAowDRfnRfjwpPd4GwnqDwZCeK9WEwvnAfv9fh0YSgHggEYokVCZPj4EYnthgK93hiEExsUDEFIgA) with a couple less scary assertions

cdimitroulas:
Perfect! This works great. I'll update the PR with this new implementation. Are we happy to drop the uncurried overload? Personally I'm ok with it

thewilkybarkid:
I've just had a play and looks like the best option. :+1:

[Couldn't simplify the type any further.](https://www.typescriptlang.org/play?#code/JYWwDg9gTgLgBAbzmYYCmcC+cBmUIhwDkOYAtDAM4D0ANsAEbU4CuAdgMYzARtEDcAKEEATNB1oBDKBg69K8MPjAAuOAB5BcOAEEANFrgAFSTAAWcNAA8YaNiMpwA1mgCeEHLss27DuGzQANzQoOAB+OAUoYDYAczg1F3dPHUEAPgAKMFMzNRNzAEo4AF40jUMAIW9be0cAJXFoEXV8sz04dic2CAB3NjT0jMMIBgArRLcPL2savwDg0IiqtVSi0uMc6t9HJKmdcN0AbVaAXQS4CuOck+ExCWlZeUU1BEM2SRA0NSiY2MNJWJffwsEAMEKCTDCORsBRwQEwAByHwwxWQygyRHenyIBUE0Nh8IAYhAICU0RAwBicCScVCnnCAIzfGDROJklDoLLtJQUjFYtA4orUaiRMwQFi0ERwSQsGAQOTgWhoWx4+mxABMzNZ8VRHLQXPJlJINIKQpF1nQXDQUpwkmAtBYMlVMPgsQAzFrfuzUPqwO14UjPrj8a6ACyetm6n0Gokks3eS22G12h1OoA)

cdimitroulas:
@gcanti for some reason I got a bunch of unrelated docs changes after I rebased my branch and I reran `npm run test`

thewilkybarkid:
Turned my playground into a dtslint file (https://github.com/cdimitroulas/fp-ts/pull/1).

gcanti:
@cdimitroulas ah sorry, I'm not generating the docs on every commit as it slowed me down.

I was thinking about the usefulness of a `prop` function, what are the use cases?

This use case is not compelling

```ts
interface Person {
  readonly name: string
  readonly age: number
}

declare const p: Person

pipe(p, prop('name'))
```

as you can just project with the same effort (and standard autocompletion)

```ts
pipe(p, (_) => _.name)
```

If I want a generic "accept anything with a 'name' property", I would expect a polymorphic function as a result:

```ts
// const getName: <A>(obj: Record<"name", A>) => A
const getName = prop('name')
```

Another use case that would be nice is to make `contramap`s easier to define:

```ts
import * as S from 'fp-ts/string'
import * as Ord from 'fp-ts/Ord'
import * as Eq from 'fp-ts/Eq'

interface Person {
  readonly name: string
  readonly age: number
}

declare const p: Person

const eq = pipe(
  S.Eq,
  Eq.contramap((p: Person) => p.name) // <= awkward
)

// const myEq: Eq.Eq<Record<"name", string>>
const myEq = pipe(S.Eq, Eq.contramap(prop('name'))) // nice

// const myOrd: Ord.Ord<Record<"name", string>>
const myOrd = pipe(S.Ord, Ord.contramap(prop('name'))) // nice
```

Other use cases?

cdimitroulas:
>This use case is not compelling {...} as you can just project with the same effort (and standard autocompletion)
>`pipe(p, (_) => _.name)`

I can understand your argument, but I think many in the ecosystem find this to be slightly awkward as well, and would prefer to do it with a function like `prop`. Especially those who are fans of point-free style.

The contramap use-case is nice :+1:

anthonyjoeseph:
> I would expect a polymorphic function as a result

That's the current behavior - would you like to remove the autocomplete functionality to simplify it?

cdimitroulas:
I tested out the `contramap` use-case you mentioned @gcanti and it works only if we add the explicit type annotations to `contramap` like so:
```ts
// const myEq: Eq.Eq<Record<"name", string>>
const myEq = pipe(Eq.eqString, Eq.contramap<string, Person>(prop('name'))) // nice

// const myOrd: Ord.Ord<Record<"name", string>>
const myOrd = pipe(Ord.ordString, Ord.contramap<string, Person>(prop('name'))) // nice
```

So it doesn't seem like the `prop` function really helps to keep this terse that much unfortunately

anthonyjoeseph:
Good catch @cdimitroulas ! I think I've solved this ([source](https://www.typescriptlang.org/play?ssl=21&ssc=36&pln=21&pc=40#code/JYWwDg9gTgLgBAbzmYYCmcC+cBmUIhwDkOYAtDAM4D0ANsAEbU4CuAdgMYzARtEBQoSLDgAqOAENKcAMq58hEuSp1G1SjCjA2AcwFDo8cVLgBRAI7yCxUhRr0mFgfzQAPYfFaduvZPjAAzAA8-HBwAAoSMAAWcG4waGwAJtIA1mgAnhA4cACCca4JydJsaABuaFBwAPxwGlq6cABccOlZObkANHCheZ29AML8AHwAFGBR0S2RMQCULeypbBAA7mwFRSlwQ7VBAEIbiVtIANqpcNoRkwC6C2xLq+uYY70QDABWLW3ZeYfFcKUKlVagcWrl+LM4ABeYZXGJ-LbfDo1PInGbRa7NOB7NE3HotUavD5fTI-fLxI4lcqVFGnc6XdG3bZYLHgyEwuGxCn-JG-Wq5XExTEtAaIXpQNAwFhQdajUZvT6SNgZdmwhUnCZCyEmCTK-iYfj8Di8DQFACM0OQqDQoxkADoLN0LHbjWxNBIQBIwON-AFRkQ2B60ERZqG4NRqOaWhYgkhAyA0C16todFhhkaTfA3AAmS1gX3++PByERgrZlr7BHSONBu4PNZp+XE7Gq7EnANBojXDNsU1uAJ562jBBFloBTDdfMQQKFzth0sSFgwCDG8C0SUYHDQYhFohAA)) - please forgive my `any` casts, I'm still working on that

raveclassic:
Hey there, my 2 cents: shouldn't this be a part of a 3rd-party library? This reminds me lenses a lot, [monocle-ts](https://github.com/gcanti/monocle-ts) for instance. 

I'm concerned about the boundaries of `fp-ts`. There are some questions that I'd first like to discuss: 
 - will we add an `index` accessor for arrays? If not, why not? We already have a "prop" accessor. If yes, then how about other helpers?
 - are we going to reimplement libraries like `ramda` and similar in `fp-ts` core? It seems that the main goal of `fp-ts` - emulation of HKTs in TypeScript - diverges with such proposal

cdimitroulas:
@raveclassic there's a lot of discussion over at https://github.com/gcanti/fp-ts/issues/1460 relating to similar `struct` util functions, so I think that would be a good place to bring this up as I believe your concerns would carry over to those other functions as well

anthonyjoeseph:
@raveclassic very good points, definitely important stuff to consider.

I know this wasn't your point, but you made me realize that `prop` could support tuples as well - [POC](https://www.typescriptlang.org/play?#code/JYWwDg9gTgLgBAbzmYYCmcC+cBmUIhwDkOYAtDAM4D0ANsAEbU4CuAdgMYzARtEBQoSLDgAqOAENKcAMq58hEuSp1G1SjCjA2AcwFDo8cVLgBRAI7yCxUhRr0mFgfzQAPYfA68NyfGDgAvHAAPPxwcAAKEjAAFnBuMGhsACbSANZoAJ4QOHAAkmw4aFDxrokp0mxoAG7FcAD8cBpaunAAXHAZ2bkFRVAANHBh+YXF-cMASmgw-AB8ABRg0TEdUbEAlIGzcPPzEAwAVh0SbJmbAdv7BwDaS7EAupsm7GlsEADubKXlqXBTM+FGsEIhBaJlvklfkhrmk4NpIst7h0Xm9PlgFsNwlcOl0ciM+hCKnAqrUSo0QWD2vjisNzts1nEEpD0lk8b06o12VBrgz7lSKZkeYihuEOntDh1obD4byOv8sHSEbFCb9cT1RmTqdzeVT-vx+F42D43ABGQLIVBoeYyAB0FkGFhths0EhAEjAiz88yIbFdaCI60DcGo1FKJo6FmCSF9IDQHWa2h06IN3ngbgATOawF6fX6A8HQxmOsEAEIq6TRv3ItivD5fTALbFwEuKkvXXOxoj3FNGtOuADMWct8wQMbjcH7mEG2YgHo7-qDIckLBgEC84Fo0wwOGgxDHAkNxtcABYh+h5u2dxAu5JpIeYNOcwAGAObJcSFdrghgTeJXC7+ciCAA)

samhh:
Slightly unrelated but @raveclassic I've got [this document](https://samhh.github.io/fp-ts-std/ramda) that attempts to keep track of Ramda equivalency in fp-ts[-std], it might prove useful.

anthonyjoeseph:
@cdimitroulas would you mind updating this PR to be against `master`? I think it got auto-closed along with the `2.11` branch

cdimitroulas:
Done @anthonyjoeseph

anthonyjoeseph:
I've published a library called [spectacles-ts](https://github.com/anthonyjoeseph/spectacles-ts) that mirrors this functionality - it'd be great to get feedback from all you smart folks on this thread!

```ts
import { get } from 'spectacles-ts'
import * as Eq from 'fp-ts/Eq'
import { Eq as StringEq } from 'fp-ts/string'

const piped: number = pipe({ a: { b: 123 } }, get('a', 'b'))
const polymorphic: number = get('a', 'b')({ a: { b: 123 } })
const inferred: Eq.Eq<{ a: { b: string } }> = pipe(
  StiringEq,
  Eq.contramap(get('a', 'b'))
)
```

Issue #1446, opened 2021-03-26
v2.11 Roadmap


gcanti:
(pretty much a big backporting from v3)

- **Deprecation**
  - array utils?: (see also #1404) to avoid a combinatorial explosion of functions add a single `traverseReadonlyArrayWithIndex` (and `traverseReadonlyArrayWithIndexSeq` when makes sense) to all data types and deprecate:
    - `traverseArrayWithIndex`
    - `traverseArray`
    - `sequenceArray`
    - `traverseSeqArrayWithIndex`
    - `traverseSeqArray`
    - `sequenceSeqArray`
  - `Array` / `ReadonlyArray`
    - deprecate `range`, use `NonEmptyArray` / `ReadonlyNonEmptyArray` APIs
  - `Traversable`
    - deprecate `sequence`?
  - `ReadonlyTuple`
    - add `Tuple2` module and deprecate `ReadonlyTuple`?
  - `Foldable`
    - deprecate `traverse_`?
  - rename APIs that end with `M` to `E`: example `reduceM` -> `reduceE`
- **Internal**
  - `internal` module
- **New Feature**
  - `EitherT`
    - add `bracket` and refactor:
      - `IOEither`
      - `ReaderEither`
      - `ReaderTaskEither`
      - `TaskEither`
    - add (ref: https://github.com/gcanti/fp-ts/issues/1360):
      - `orLeft`
      - `orElseFirst`
      - add them to (also `orElseFirstW`):
        - `IOEither`
        - `ReaderEither`
        - `ReaderTaskEither`
        - `TaskEither`
  - `Witherable`
    - add `wiltDefault`, `witherDefault`
      - and use them in:
        - `Either`
        - `Option`
        - `ReadonlyArray`
        - `ReadonlyMap`
        - `ReadonlyRecord`
  - `These`
    - add `ApT`
    - (internal) remove `Option` dependency
    - add `traverseReadonlyNonEmptyArrayWithIndex` e `traverseReadonlyArrayWithIndex`
  - `TaskThese`
    - add `ApT`
    - add `traverseReadonlyNonEmptyArrayWithIndex` e `traverseReadonlyArrayWithIndex`, `traverseReadonlyNonEmptyArrayWithIndexSeq` e `traverseReadonlyArrayWithIndexSeq`
  - `function`
    - add `SK`
  - https://github.com/gcanti/fp-ts/issues/1404
    - `Either`
    - `IO`
    - `IOEither`
    - `Option`
[...]

PR #1455, opened 2021-03-28
Backport ReadonlyRecord & Record

anthonyjoeseph:
Backports `ReadonlyRecord` from 3.0.0 based on the 2.11 roadmap https://github.com/gcanti/fp-ts/issues/1446

I agree w/ the `Ord` idea outlined in the roadmap, but I can change if desired.

Notes:
- I used `Ord` constraints in `Record` as well, I can change it to use `Ord.trivial` if desired
- I wasn't sure why `ReadonlyRecord.wilt` and `ReadonlyRecord.wither` [use `Ord.trivial`  in 3.0.0](https://github.com/gcanti/fp-ts/blob/f7ae980afcab8f68d46b801f35462e644a64a65c/src/ReadonlyRecord.ts#L756), but I mirrored that - I'm not sure if that was correct
- I also added `Ord` instances to these, which were unspecified - we can use `Ord.trivial` instead if desired
  - `traverse`
  - `sequence`
  - `getShow`
  - typeclass instances (`Traversable` -> `getTraversable: (O: Ord<string>) => ...`)

### DIFF
--- a/docs/modules/ReadonlyRecord.ts.md
+++ b/docs/modules/ReadonlyRecord.ts.md
@@ -43,18 +43,27 @@ Added in v2.5.0
   - [Compactable](#compactable-1)
   - [Filterable](#filterable-1)
   - [FilterableWithIndex](#filterablewithindex)
-  - [Foldable](#foldable-1)
-  - [FoldableWithIndex](#foldablewithindex)
   - [Functor](#functor)
   - [FunctorWithIndex](#functorwithindex)
-  - [Traversable](#traversable)
-  - [TraversableWithIndex](#traversablewithindex)
   - [URI](#uri)
   - [URI (type alias)](#uri-type-alias)
-  - [Witherable](#witherable-1)
+  - [getDifferenceMagma](#getdifferencemagma)
   - [getEq](#geteq)
+  - [getFoldable](#getfoldable)
+  - [getFoldableWithIndex](#getfoldablewithindex)
+  - [getIntersectionSemigroup](#getintersectionsemigroup)
   - [getMonoid](#getmonoid)
   - [getShow](#getshow)
+  - [getTraversable](#gettraversable)
+  - [getTraversableWithIndex](#gettraversablewithindex)
+  - [getUnionMonoid](#getunionmonoid)
+  - [getUnionSemigroup](#getunionsemigroup)
+  - [getWitherable](#getwitherable)
+  - [~~FoldableWithIndex~~](#foldablewithindex)
+  - [~~Foldable~~](#foldable)
+  - [~~TraversableWithIndex~~](#traversablewithindex)
+  - [~~Traversable~~](#traversable)
+  - [~~Witherable~~](#witherable)
   - [~~readonlyRecord~~](#readonlyrecord)
 - [interop](#interop)
   - [fromRecord](#fromrecord)
@@ -63,6 +72,7 @@ Added in v2.5.0
   - [ReadonlyRecord (type alias)](#readonlyrecord-type-alias)
 - [utils](#utils)
   - [collect](#collect)
+  - [difference](#difference)
   - [elem](#elem)
   - [empty](#empty)
   - [every](#every)
@@ -71,6 +81,7 @@ Added in v2.5.0
   - [fromFoldable](#fromfoldable)
   - [fromFoldableMap](#fromfoldablemap)
   - [has](#has)
+  - [intersection](#intersection)
   - [isEmpty](#isempty)
   - [isSubrecord](#issubrecord)
   - [keys](#keys)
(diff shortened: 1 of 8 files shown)


gcanti:
> I agree w/ the Ord idea outlined in the roadmap

We can't just change the signatures though, this PR contains the following breaking changes:

- `keys`
- `collect`
- `toReadonlyArray`
- `toUnfoldable`
- `reduceWithIndex`
- `foldMapWithIndex`
- `reduceRightWithIndex`
- `traverseWithIndex`
- `traverse`
- `sequence`
- `reduce`
- `foldMap`
- `reduceRight`
- `getShow`
- `Foldable` -> `getFoldable`
- `FoldableWithIndex` -> `getFoldableWithIndex`
- `Traversable` -> `getTraversable`
- `TraversableWithIndex` -> `getTraversableWithIndex`
- `Witherable` -> `getWitherable`

> I wasn't sure why ReadonlyRecord.wilt and ReadonlyRecord.wither...

This was a mistake, thanks for pointing out (fixed now in v3)

anthonyjoeseph:
Oh, ok - I misunderstood the roadmap & the release process, my bad. Do you think it would make sense to new functions w/ `*WithOrd` signatures (e.g. `keysWithOrd`)? Or maybe it's best to leave all that out for the moment?

gcanti:
Yeah, I'm sorry, the roadmap wan't clear on this point, I put a "?" on the item list because I'm not even sure is possible to backport the new constraint.

In general we can do whatever we want in the `3.0.0` branch (with an eye on the migration path though) but in the `2.x` series we can't introduce breaking changes.

In the past I used the following technique:

- add a new signature as an overload to the API
- deprecate the old signature

Let's pick `keys` for example:

```ts
export const keys = <K extends string>(r: ReadonlyRecord<K, unknown>): ReadonlyArray<K> =>
  (Object.keys(r) as any).sort()
```

The new definition would be something like:

```ts
declare const isOrd: (r: Ord<string> | ReadonlyRecord<string, unknown>) => r is Ord<string>

/**
 * @since 2.5.0
 */
export function keys(O: Ord<string>): <K extends string>(r: ReadonlyRecord<K, unknown>) => ReadonlyArray<K>
/**
 * Use the other overload instead
 *
 * @deprecated
 */
export function keys<K extends string>(r: ReadonlyRecord<K, unknown>): ReadonlyArray<K>
export function keys(
  or: Ord<string> | ReadonlyRecord<string, unknown>
): ((r: ReadonlyRecord<string, unknown>) => ReadonlyArray<string>) | ReadonlyArray<string> {
  return isOrd(or) ? keys(or) : (Object.keys(or) as any).sort()
}
```

The problem is the `isOrd` guard: how can we reliably distinguish between an `Ord` and a `ReadonlyRecord`?

anthonyjoeseph:
No worries! Oh good question. Maybe something like this?

```ts
const isOrd = <A>(a: Record<string, unknown> | Ord<A>): a is Ord<A> =>
  keys(a as Record<string, unknown>).length === 2
  && 'equals' in a
  && typeof a.equals === 'function'
  && 'compare' in a && typeof a
  && typeof a.compare === 'function'
```

I know it's super kludgy, but we could _probably_ assume that would work

gcanti:
I would:

- keep as is:
  - `keys` (you can always sort the output, or, if you want the insertion order, use `Object.keys`)
  - `toReadonlyArray` (you can always sort the output, or, if you want the insertion order, use `Object.values`)
  - `toUnfoldable`, `traverseWithIndex`, `traverse`, `sequence` (it doesn't worth it)
- apply the technique to: 
  - `collect` (this first parameter is a function so we are safe)
  - `reduceWithIndex`, `reduceRightWithIndex`, `reduce`, `reduceRight` (checking the `arguments` length I guess)
  - `foldMapWithIndex`, `foldMap`, `getShow` (checking `compare`? ugly...)
- add a new API and deprecate the old one (which we keep as is to not introduce breaking changes):
  - `Foldable` -> `getFoldable`
  - `FoldableWithIndex` -> `getFoldableWithIndex`
  - `Traversable` -> `getTraversable`
  - `TraversableWithIndex` -> `getTraversableWithIndex`
  - `Witherable` -> `getWitherable`

gcanti:
> I would...

btw I know this is an horrible task, so feel free to close this PR or just clean it up by keeping only the new additions like `union`, `intersection`, etc...

anthonyjoeseph:
Not at all - it's my pleasure! It should be good now.

By the way, I saw your comment in the Roadmap - please feel free to close this or steal from it if that'd be easier for you. Or I can remove the `Ord` stuff if that's easiest

gcanti:
@anthonyjoeseph awesome, thank you!

PR #1456, opened 2021-03-28
Relocate Struct Instances into the Struct Module

anthonyjoeseph:
This PR relocates `Show.struct`, `Eq.struct`, `Semigroup.struct` and `Monoid.struct` into the new `struct.ts` module.

The deprecated `getStruct*` functions are implemented separately to avoid a circular dependency


gcanti:
@anthonyjoeseph I thought about this for v2.10 but decided to keep them where they are

anthonyjoeseph:
Ok - understood!

PR #1457, opened 2021-03-28
Add `struct` functions

anthonyjoeseph:
This PR adds a few functions to the `struct` module, mostly with an `S` suffix (taking inspiration from `sequenceS`)
 
Notes:
- `traverseS_` is an un-curried version of `traverseS` where the transformation struct can be defined separately. It's named after `Foldable.traverse_`, which looks a bit similar. We can remove this if desired, since `Foldable.traverse_` is deprecated, but I thought it might be useful since `traverseS` won't work outside of `pipe`
- [evolve](https://github.com/gcanti/fp-ts/blob/f1884ce7f064bc62002acbacd6bc51019934771e/src/struct.ts#L57) has been renamed to `mapS` - I'm certainly willing to keep `evolve`, but `mapS` seems to fit the conventions of the module
- Most importantly - `Struct` is not a HKT, so it doesn't necessarily fit the ethos of fp-ts. It's _almost_ a member of many typeclasses (Functor, Traversable, Witherable), but it doesn't fit any of the traditional function signatures. For this reason, it might not be a great fit

Ideas:
- could we add [pick & omit](https://github.com/samhh/fp-ts-std/blob/e5f465778d984f6618d92b3ac419cae8e03551a7/src/Record.ts#L58) from `fp-ts-std`?
- maybe there could be similar functions for `tuple` (e.g. `tuple.evolve` -> `tuple.mapT`)?

### DIFF
--- a/docs/modules/struct.ts.md
+++ b/docs/modules/struct.ts.md
@@ -6,18 +6,969 @@ parent: Modules
 
 ## struct overview
 
-Added in v3.0.0
+A 'struct' is a heterogeneous `ReadonlyRecord`, often an `interface`.
+Many of these functions have a `Record._WithIndex` counterpart.
+e.g. struct.mapS <-> Record.mapWithIndex
+
+Added in v2.10.0
 
 ---
 
 <h2 class="text-delta">Table of contents</h2>
 
+- [combinators](#combinators)
+  - [compactS](#compacts)
+  - [filterMapS](#filtermaps)
+  - [filterS](#filters)
+  - [foldMapS](#foldmaps)
+  - [mapS](#maps)
+  - [partitionMapS](#partitionmaps)
+  - [partitionS](#partitions)
+  - [reduceS](#reduces)
+  - [separateS](#separates)
+  - [traverseS](#traverses)
+  - [traverseS\_](#traverses_)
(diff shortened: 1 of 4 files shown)


gcanti:
> mostly with an S suffix

We can remove the suffix IMO, the module is already in itself a namespace.

The return type of `keys` looks problematic:

```ts
type A = {
  readonly a: string
}

declare const x: { readonly a: string; readonly b: number }
const y: A = x
const ykeys = keys(S.Ord)(y) // => readonly "a"[]
```

> evolve has been renamed to mapS

If we remove the S suffix we end up with `map` which is not correct IMO, this is not the `Functor`'s `map` (btw I picked "evolve" from [ramda](https://ramdajs.com/docs/#evolve), perhaps people is already accustomed to this name).

In general: I can see the value of something like `evolve`, `compactS` / `unCompact` (<= for interop) and maybe `pick` / `omit`, but is there a valid use case for all the other functions?
I don't want to add (and maintain) something that is rarely used in the wild.

anthonyjoeseph:
Re: `keys` - yikes! Good call, I'll remove that one at least.

Re: use cases - I guess I should have made a feature request first and gathered some input - that was my bad. Closing this for now in favor of that.

PR #1470, opened 2021-04-05
Do notation docs

cdimitroulas:
Opening a draft PR to get some initial feedback on some documentation for the `Do` notation. The location and name of this file is not final, I wasn't sure where exactly we wanted this to live (in `guides` for example) so I just created it at the top level within `docs` for now with the idea that we can move it somewhere more appropriate once it's ready.

@pfgray feel free to add some more detailed examples at the end as you mentioned to me on Slack


docs/do-notation.md
```diff
@@ -0,0 +1,79 @@
...
+Here's how we can write `main` with do notation (we'll call it `mainDo`):
+```ts
+const mainDo: T.Task<{ x: string, y: string }> = pipe(
+  T.Do,
+  T.bind('x', () => readLine),
+  T.bind('y', () => readLine),
+  T.chainFirst(({ x }) => print(x)),
+  T.chainFirst(({ y }) => print(y)),
+)
+```
```
vinassefranche, line 42: First, thanks for this doc, it would have been really useful for me a few months ago :-)
I find the `bindTo` method also useful sometimes (specially if the first Task needs to be built with a couple of function) so I think it would be nice to add it here as another way of writing the same function:
```
const mainDo: T.Task<{ x: string, y: string }> = pipe(
  readLine,
  T.bindTo('x'),
  T.bind('y', () => readLine),
  T.chainFirst(({ x }) => print(x)),
  T.chainFirst(({ y }) => print(y)),
)
```
What do you think?

docs/do-notation.md
```diff
@@ -0,0 +1,79 @@
...
+Let's take a look at an example of how do notation can help to simplify our code. Here we have
+a bit of code which reads two values from the command line, prints them and stores them
+in an object with `x` and `y` properties.
+
+```ts
+import { pipe } from 'fp-ts/function'
+import * as T from 'fp-ts/Task'
+
+declare const print: (s: string) => T.Task<void>
+declare const readLine: T.Task<string>
```
vinassefranche, line 19: Since `readLine` is already a Task and not a lazy function returning a Task, I think the `apS` method is more appropriate here:
```
const mainDo: T.Task<{ x: string, y: string }> = pipe(
  T.Do,
  T.apS('x', readLine),
  T.apS('y', readLine),
  T.chainFirst(({ x }) => print(x)),
  T.chainFirst(({ y }) => print(y)),
)
```
I'm not sure how to introduce both concepts but maybe we can update your example with `readline` being:
```
declare const readLine: () => T.Task<string>
```
and the bind being used like: 
```
T.bind('x', () => readLine()),
// or even:
T.bind('x', readLine),
```
and below, we could add that if you have `Tasks` not wrapped in a lazy function, you can use `apS`.
What do you think?

gcanti: In the case of `Task`, you can use the default exported `apS` when sequentiality doesn't matter. 

For example while running the following program:

```ts
const parallel: T.Task<{ x: string, y: string }> = pipe(
  T.Do,
  T.apS('x', readLine),
  T.apS('y', readLine)
)
```

`y` **may** be requested before `x`. However in the following program:

```ts
const sequential: T.Task<{ x: string, y: string }> = pipe(
  T.Do,
  T.apS('x', readLine),
  T.bind('y', () => readLine)
)
```

`y` is guaranteed being requested after `x`.

cdimitroulas: Interesting, thanks for the explanation.

I think that sticking with `bind` for the initial example is the best as it's the closest to the do notation from Haskell/Purescript and so serves as a direct translation.

The idea is to tack on a bunch of other examples at the end of this page so I think that would be a good place to add examples with `apS`, `bindTo` etc.

@pfgray tagging you as well for visibility

vinassefranche: I don't know if it's worth mentioning on the initial example, but another big difference between `bind` and `apS` is that you can use what's currently in the Monad to compute the value with `bind` while you can't with `apS`.
Here's an example:
```ts
import * as T from 'fp-ts/Task'

declare const print: (s: string) => T.Task<void>
declare const readLine: T.Task<string>
declare const formatString: (s: string) => T.Task<string>

const mainDo: T.Task<{ x: string, y: string }> = pipe(
  T.Do,
  T.bind('x', () => readLine),
  T.bind('y', ({ x }) => formatString(x)),
  T.chainFirst(({ x }) => print(x)),
  T.chainFirst(({ y }) => print(y)),
)
```
We can put in with the other examples at the end the page anyway.

docs/do-notation.md
cdimitroulas: I added this as an example at the end

docs/do-notation.md
```diff
@@ -0,0 +1,115 @@
...
+)
+```
+
+Performing actions in parallel with `apS`:
+
+```ts
+import { pipe } from 'fp-ts/function'
+import * as T from 'fp-ts/Task'
+
+declare const encryptValue: (val: string): T.Task<string>
```
vinassefranche, line 107:
```suggestion
declare const encryptValue: (val: string) => T.Task<string>
```

docs/do-notation.md
```diff
@@ -0,0 +1,115 @@
...
+
+pipe(
+  T.Do,
+  T.apS('x', () => encryptValue("hello")),
+  T.apS('y', () => encryptValue("world")),
```
vinassefranche, line 112:
```suggestion
  T.apS('x', encryptValue("hello")),
  T.apS('y', encryptValue("world")),
```
`Task.apS` takes a `Task` as a second argument contrary to `bind` which takes a function returning a `Task`

docs/do-notation.md
vinassefranche: Thanks!

docs/do-notation.md
cdimitroulas: Thanks

gcanti:
> I wasn't sure where exactly we wanted this to live

@cdimitroulas I'd say in `guides`

cdimitroulas:
Moved to `guides`

gcanti:
@cdimitroulas thanks, let me know when you want me to merge this.

I think we must add a header, something like

```
---
title: Do notation
parent: Guides
nav_order: 5
---
```

cdimitroulas:
Shall we put it above the v2 migration guide in the ordering?

cdimitroulas:
I think we can go ahead with this - more examples can be added later if needed

gcanti:
> I think we can go ahead with this - more examples can be added later if needed

@cdimitroulas I agree

I saw you added a new file `Do.md`, shouldn't the header go into `do-notation.md`?

cdimitroulas:
Whoops, my bad that was a mistake. Fixed now

PR #1473, opened 2021-04-11
add Ix{Apply,Chain,StateT,State}

waynevanson:
# Indexed Monads

I've been looking at indexed state monads and decided to try implement it as a typeclass. This is largely inspired by gcanti's hyper-ts.

In depth explanation of Indexed Monads https://stackoverflow.com/questions/28690448/what-is-indexed-monad

### IxApply

A typeclass has a function `iap` which acts similarly to `ap`, but keeps track of the index.

Also has transformers for `iapFirst` and `iapSecond`.

### IxChain
A typeclass that has a function `ichain`, which acts simarly to `chain` but keeps track of the index.
It also inherits `iap` from `IxApply`.

Also has transformers for Do notation and `ichainFirst`

## IxStateT (IndexedState)

Polymorphic state values, with state tracked via an index.

Included are transformers for the IxChain and IxApply, as well as transformers for `IxState`

## Questions

- I've done two overloads in `declare module './HKT` so `Pointed` works correctly for `StateT`. Is this allowable?

commits can be rebasable instead of merged.

### DIFF
--- a/src/IxApply.ts
+++ b/src/IxApply.ts
@@ -1,0 +1,78 @@
+/**
+ * @since 2.10.0
+ */
+import { pipe } from './function'
+import { Functor3, Functor4 } from './Functor'
+import { HKT3, Kind3, Kind4, URIS3, URIS4 } from './HKT'
+
+/**
+ * @category type classes
+ * @since 2.10.0
+ */
+export interface IxApply4<F extends URIS4> extends Functor4<F> {
+  readonly iap: <O, Z, E, A>(
+    fa: Kind4<F, O, Z, E, A>
+  ) => <I, B>(fab: Kind4<F, I, O, E, (a: A) => B>) => Kind4<F, I, Z, E, B>
+}
+
+/**
+ * @category type classes
+ * @since 2.10.0
+ */
+export interface IxApply3<F extends URIS3> extends Functor3<F> {
+  readonly iap: <O, Z, A>(fa: Kind3<F, O, Z, A>) => <I, B>(fab: Kind3<F, I, O, (a: A) => B>) => Kind3<F, I, Z, B>
+}
+
+/**
+ * @category type classes
+ * @since 2.10.0
+ */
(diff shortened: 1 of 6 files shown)


derrickbeining:
@waynevanson I was just working on porting `purescript-react-basic-hooks` to TS and implemented some of this myself recently. Glad to see other folks are interested in having indexed types in fp-ts. It seems they were included at one point but were removed. Not sure why.

I'd really like to see you work merged. I see you only did IxApply and IxChain. Are you interested in including `IxPointed`, `IxApplicative`, `IxFunctor`, and `IxMonad` as well?

waynevanson:
The scope you're wishing to amend makes the result quite large. We're basically making ix copies of existing modules, which may pollute the environment. 

Should we break this down into a separate package? fp-ts-indexed?

derrickbeining:
We _can_ make a separate package if @gcanti isn't interested in having these added to fp-ts core. But it'd be nice to have in core if there's not any particular reason for excluding it.

If you and @gcanti are interested in having the rest of the type classes included, I've opened a PR targeting this branch to add the remaining classes https://github.com/waynevanson/fp-ts/pull/1.

derrickbeining:
Alternatively, would `fp-ts-contrib` be a preferable home for these contributions?

gcanti:
@derrickbeining @waynevanson a separate package (or fp-ts-contrib) would be preferable :+1:

derrickbeining:
@gcanti I've put up a PR for the type classes here https://github.com/gcanti/fp-ts-contrib/pull/79

derrickbeining:
Hey y'all, update: I just went ahead and published [fp-ts-indexed-monad](https://www.npmjs.com/package/fp-ts-indexed-monad) as a separate package. Do we want to close this PR and the one at `fp-ts-contrib`?

waynevanson:
Yes please, sounds good.

Thank you!

PR #1476, opened 2021-04-12
Add string.toLowerCase and use string manipulation type in toUpperCase

thewilkybarkid:
Adds `string.toLowerCase` and improves the return type of `string.toUpperCase`.

`toLowerCase` can be pack ported to `2.11`, but the string manipulation types require TypeScript 4.1.

I've left this as WIP while a `@ts-ignore` is removed.

### DIFF
--- a/test/Const.ts
+++ b/test/Const.ts
@@ -22,6 +22,7 @@ describe('Const', () => {
     })
 
     it('mapLeft', () => {
+      // @ts-ignore
       const f = _.mapLeft(S.toUpperCase)
       U.deepStrictEqual(pipe(_.make('a'), f), _.make('A'))
     })
--- a/docs/modules/string.ts.md
+++ b/docs/modules/string.ts.md
@@ -22,6 +22,7 @@ Added in v3.0.0
   - [empty](#empty)
   - [isEmpty](#isempty)
   - [size](#size)
+  - [toLowerCase](#tolowercase)
   - [toUpperCase](#touppercase)
 
 ---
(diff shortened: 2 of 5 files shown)


test/Const.ts
```diff
@@ -22,6 +22,7 @@ describe('Const', () => {
     })
 
     it('mapLeft', () => {
+      // @ts-ignore
       const f = _.mapLeft(S.toUpperCase)
```
thewilkybarkid, line 26: Fails to infer the input type as `string`.

thewilkybarkid:
```suggestion
      const f = pipe(S.toUpperCase, _.mapLeft)
```

works, as does

```suggestion
      const f = _.mapLeft<string, string>(S.toUpperCase)
```

Neither feels ideal though.

src/Bifunctor.ts
```diff
@@ -26,7 +26,7 @@ export interface Bifunctor<F> {
 export interface Bifunctor2<F extends URIS2> {
   readonly URI?: F
   readonly bimap: <E, G, A, B>(f: (e: E) => G, g: (a: A) => B) => (fea: Kind2<F, E, A>) => Kind2<F, G, B>
-  readonly mapLeft: <E, G>(f: (e: E) => G) => <A>(fea: Kind2<F, E, A>) => Kind2<F, G, A>
+  readonly mapLeft: <E, G, A>(f: (e: E) => G) => (fea: Kind2<F, E, A>) => Kind2<F, G, A>
```
thewilkybarkid, line 29: Not ideal to change this here, required by https://github.com/gcanti/fp-ts/pull/1476#discussion_r611437693.

PR #1478, opened 2021-04-13
Add ReaderTask.chainFirstW and Reader.chainFirstW

vinassefranche:


gcanti:
Thanks @vinassefranche

PR #1481, opened 2021-04-15
Workaround issue with `hasOwnProperty`

OliverJAsh:
Re. https://github.com/gcanti/fp-ts/issues/1249#issuecomment-820232156

### DIFF
--- a/src/internal.ts
+++ b/src/internal.ts
@@ -25,7 +25,7 @@ export const isLeft = <E, A>(ma: Either<E, A>): ma is Left<E> => ma._tag === 'Le
 // -------------------------------------------------------------------------------------
 
 /** @internal */
-export const hasOwnProperty = Object.prototype.hasOwnProperty
+export const has = Object.prototype.hasOwnProperty
 
 // -------------------------------------------------------------------------------------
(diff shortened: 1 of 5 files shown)


src/internal.ts
```diff
@@ -24,8 +24,17 @@ export const isLeft = <E, A>(ma: Either<E, A>): ma is Left<E> => ma._tag === 'Le
 // Record
 // -------------------------------------------------------------------------------------
 
-/** @internal */
-export const hasOwnProperty = Object.prototype.hasOwnProperty
```
gcanti, line 28: @OliverJAsh we can just change the name to `has` here

gcanti (changes requested):

src/internal.ts
OliverJAsh: Good idea, done!

gcanti (approved):

OliverJAsh:
Thanks! I see a 2.10.1 tag has been pushed to GitHub but the release doesn't seem to exist on npm yet?

gcanti:
@OliverJAsh should be a matter of seconds, let me know if it's ok

PR #1482, opened 2021-04-15
version 2.10.2

gcanti:
@OliverJAsh this should fix the last issue


gcanti:
@OliverJAsh what do you think? Maybe you can try to temporary patch the `.js` file locally with this change to double check the fix before publishing, just to be safe

OliverJAsh:
@gcanti Sorry I had to run an errand. I've tested locally and this definitely fixes it. Thank you!

gcanti:
Patch released 👍

PR #1489, opened 2021-04-20
fix altW function signature in all Either modules, fixes #1488

### LINKED ISSUE #1488: Either altW 'left' is too wide
## 🐛 Bug report

The function signature for `Either` `altW` is as follows:

```
export declare const altW: <E2, B>(that: Lazy<Either<E2, B>>) => <E1, A>(fa: Either<E1, A>) => Either<E1 | E2, A | B>
```

From my understanding of `alt`, `Left<E1>` can never be produced, so I believe the signature should be as follows:
```
export declare const altW: <E2, B>(that: Lazy<Either<E2, B>>) => <E1, A>(fa: Either<E1, A>) => Either<E2, A | B>
```

### Current Behavior

```
const foo = pipe(
    E.left<number, number>(1),
    E.altW(() => E.right<string, string>('a')),
);
```

In this case, `foo`'s type is `Either<string | number, string | number>`.

### Expected behavior

I believe `foo`'s type should be `Either<string, string | number>`.

### Reproducible example

See above.

### Suggested solution(s)

Remove E1 from the return type of `altW`:

```
export const altW: <E2, B>(
    that: Lazy<Either<E2, B>>,
) => <E1, A>(fa: Either<E1, A>) => Either<E2, A | B> = that => fa => (isLeft(fa) ? that() : fa);
```

I could make a PR if you agree.

### Your environment

| Software   | Version(s) |
| ---------- | ---------- |
| fp-ts      |  2.10.2          |
| TypeScript |  4.2.4          |

wmaurer:
Fixes #1488

### DIFF
--- a/docs/modules/Either.ts.md
+++ b/docs/modules/Either.ts.md
@@ -166,7 +166,7 @@ Less strict version of [`alt`](#alt).
 **Signature**
 
 ```ts
-export declare const altW: <E2, B>(that: Lazy<Either<E2, B>>) => <E1, A>(fa: Either<E1, A>) => Either<E2 | E1, B | A>
+export declare const altW: <E2, B>(that: Lazy<Either<E2, B>>) => <E1, A>(fa: Either<E1, A>) => Either<E2, B | A>
 ```
 
 Added in v2.9.0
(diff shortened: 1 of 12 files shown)


gcanti (approved):

gcanti:
Thanks @wmaurer, patch released.

PR #1490, opened 2021-04-20
Add Date.fromString constructor

thewilkybarkid:
As `new Date('not a date')` is valid only for its usage to fail badly, this adds a `Date.fromString` constructor (along with `Date.isValid`).

### DIFF
--- a/src/Date.ts
+++ b/src/Date.ts
@@ -2,8 +2,11 @@
  * @since 2.0.0
  */
 import * as E from './Eq'
+import { flow } from './function'
 import { IO } from './IO'
-import * as O from './Ord'
+import * as O from './Option'
+import { Ord as Ord_, ordDate } from './Ord'
+import { not, Predicate } from './Predicate'
 
 // -------------------------------------------------------------------------------------
 // instances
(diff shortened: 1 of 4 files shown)


src/Date.ts
```diff
@@ -2,8 +2,11 @@
  * @since 2.0.0
  */
 import * as E from './Eq'
+import { flow } from './function'
 import { IO } from './IO'
-import * as O from './Ord'
+import * as O from './Option'
+import { Ord as Ord_, ordDate } from './Ord'
```
thewilkybarkid, line 8: Preference on naming collisions here?

src/Date.ts
```diff
@@ -50,12 +53,28 @@ export const eqYear: E.Eq<Date> = {
...
+ */
+export const isValid: Predicate<Date> = flow((date) => date.getTime(), not(isNaN))
+
+/**
+ * Returns a `Date` if the input is a valid date time string.
+ *
+ * @category constructors
+ * @since 2.11.0
+ */
+export const fromString: (value: string) => O.Option<Date> = flow((value) => new Date(value), O.fromPredicate(isValid))
```
EricCrosson, line 76: This is an elegant idea. Dates are tricky though, from the [MDN Documentation](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date#examples)

> Note: Parsing of date strings with the Date constructor (and Date.parse, they are equivalent) is strongly discouraged due to browser differences and inconsistencies.

[More](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/Date#parameters) from MDN, Section 3

Which may technically mean this isn't a pure function? Well, pure within one environment but not necessarily across two environments. 

The consistent way to create dates is to go for the 
``` tyepscript
Date.UTC(year, month, day, hour, minute, second, millisecond)
```
parameters, [example](https://github.com/strong-roots-capital/od/blob/7cae089458e4c1a568b8e5ecd5ec5a7e7b386b05/src/date-descriptor.ts#L58), however this function is still not injective

```
$ node --version
v15.6.0

$ npx ts-node
> new Date(1999, 0, 1)
1999-01-01T06:00:00.000Z
> new Date(99, 0, 1)
1999-01-01T06:00:00.000Z
```

This behavior is at least consistent but may be surprising, given how expectable/discoverable the rest of the fp-ts API is.

gcanti: Yeah, honestly I don't want to deal with Dates in the core, they are a nightmare. 
This could be something `fp-ts-contrib` related. 
Or maybe pick a good library and write the bindings for `fp-ts` as a separate package.

PR #1496, opened 2021-04-25
Add flattenW (less strict version of flatten) to all types that also have chainW

jussisaurio:
Eg. for `Either`, allows:

```
function a(input: E.Either<E1, E.Either<E2, A>>): E.Either<E1 | E2, A> {
  return E.flattenW(input)
}
```

instead of:

```
function a(input: E.Either<E1, E.Either<E2, A>>): E.Either<E1 | E2, A> {
  return pipe(input, E.chainW(id => id))
}
```

i.e. allows the Left type to widen.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -89,6 +89,7 @@ high state of flux, you're at risk of it changing without notice.
     - add `isBoolean`
   - `Either`
     - add `chainOptionK`
+    - add `flattenW`
   - `EitherT`
     - add `orElseFirst`
     - add `orLeft`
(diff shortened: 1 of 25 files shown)


jussisaurio:
Would probably make sense to add TaskEither.flattenW to the same PR, right?

gcanti:
@jussisaurio anywhere there's a `chainW` I guess.

p.s.
I'll change the PR target to the `2.11` branch (which is dedicated to the next release)

gcanti:
@jussisaurio nice, thank you! could you please update the changelog too?

PR #1497, opened 2021-04-25
ReaderT asks*K variants

TylorS:
This implements the ability to run an effect using the various `From*` type-classes after asking for something from the environment.

I'm happy to add some tests if they seem needed, but I figured that since it's mainly just composition and type-signatures they might not be too important.


gcanti (approved):

gcanti:
Thanks @TylorS

PR #1500, opened 2021-04-30
Add RTE.chainFirstTaskEitherK

vinassefranche:
We should have them all on ReaderTaskEither now!


gcanti:
Thanks @vinassefranche

PR #1501, opened 2021-04-30
Add RTE.add-rte-chainFirstTaskEitherK(W)

vinassefranche:
port of #1500


gcanti:
Thanks @vinassefranche

PR #1511, opened 2021-06-03
Use Error to replace unknown for the parse and stringify method of Json

aiooly:
# WHAT

The underlying method `parse` and `stringify` of [JSON](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON) will throw an error if the [JSON is invalid](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/parse#exceptions) or [failed to stringify](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify#exceptions). 

In most use cases, we will pipe the `Json.parse` into `Either<Error, Xxx>`, like below:
Screen Shot 2021-06-03 at 14 10 08

So I think we can change the return type from `Either<unknown, Xxx>` to `Either<Error, Xxx>`

### DIFF
--- a/src/Json.ts
+++ b/src/Json.ts
@@ -1,8 +1,7 @@
 /**
  * @since 2.10.0
  */
-import { Either, parseJSON, stringifyJSON } from './Either'
-import { identity } from './function'
+import { Either, parseJSON, stringifyJSON, toError } from './Either'
 
 /**
  * @since 2.10.0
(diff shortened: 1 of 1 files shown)


DenisFrezzato:
I, as a user of this library, prefer to get `unknown`. I use TypeScript also to model the domain errors, in order to discriminate over them, so getting an `Error` would not help me at all and it would be an opinionated choice for a library.

samhh:
If you're already anticipating an `unknown` I can't see how this change could harm you. On the other hand I _think_ the specification says that a `TypeError` should be thrown, at least in the case of `stringify`, and in some cases that may be helpful. It'd be more "correct" which I think is a reasonable goal to strive for.

PR #1512, opened 2021-06-03
Enhance ecosystem

thought2:
- Adds the section "Tooling" (docs-ts + a new package "create-fp-ts-lib")
- Adds two libraries (fp-ts-sized-vectors, fp-ts-number-instances)


gcanti:
Thanks @thought2

PR #1518, opened 2021-06-14
Additional 3.x Functor.map overloads

TylorS:
Adds missing overloads to `Functor.map` for Kind(2/3/4) in the 3.0.0 branch

### DIFF
--- a/src/Functor.ts
+++ b/src/Functor.ts
@@ -113,10 +113,72 @@ export function flap<F>(F: Functor<F>): <A>(a: A) => <B>(fab: HKT<F, (a: A) => B
  * @category combinators
  * @since 3.0.0
  */
+export function map<F extends URIS4, G extends URIS4>(
+  F: Functor4<F>,
+  G: Functor4<G>
+): <A, B>(
+  f: (a: A) => B
+) => <S, R, E>(fa: Kind4<F, S, R, E, Kind4<G, S, R, E, A>>) => Kind4<F, S, R, E, Kind4<G, S, R, E, B>>
+export function map<F extends URIS4, G extends URIS3>(
+  F: Functor4<F>,
+  G: Functor3<G>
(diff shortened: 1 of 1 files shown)


src/Functor.ts
```diff
@@ -113,6 +113,72 @@ export function flap<F>(F: Functor<F>): <A>(a: A) => <B>(fab: HKT<F, (a: A) => B
...
+  G: Functor4<G>
+): <A, B>(f: (a: A) => B) => <S, R, E>(fa: Kind2<F, E, Kind4<G, S, R, E, A>>) => Kind2<F, E, Kind4<G, S, R, E, B>>
+export function map<F extends URIS2, G extends URIS3>(
+  F: Functor2<F>,
+  G: Functor3<G>
+): <A, B>(f: (a: A) => B) => <R, E>(fa: Kind2<F, E, Kind3<G, R, E, A>>) => Kind2<F, E, Kind3<G, R, E, B>>
+export function map<F extends URIS2, G extends URIS2>(
+  F: Functor2<F>,
+  G: Functor2<G>
+): <A, B>(f: (a: A) => B) => <E>(fa: Kind2<F, E, Kind2<G, E, A>>) => Kind2<F, E, Kind2<G, E, B>>
```
gcanti, line 161: I think that the `E` type parameters should be different here

```ts
export function map<F extends URIS2, G extends URIS2>(
  F: Functor2<F>,
  G: Functor2<G>
): <A, B>(f: (a: A) => B) => <EF, EG>(fa: Kind2<F, EF, Kind2<G, EG, A>>) => Kind2<F, EF, Kind2<G, EG, B>>
```

Same for the other overloads

gcanti (changes requested):

src/Functor.ts
TylorS: Thanks for taking a look! Updated in https://github.com/gcanti/fp-ts/pull/1518/commits/691097d2bc9c50e72eded7a8d930f6bfb2ef8131

gcanti (approved):

gcanti:
Thanks @TylorS

Issue #1519, opened 2021-06-14
Provide a `match` function for the `Json` type


thought2:
Most of the data types in fp-ts come with a `match` function. I think this may be useful for the `Json` type, too.

Something like this:

```ts
export const match: <Z>(
  isNull: () => Z,
  isBool: (x: boolean) => Z,
  isNum: (x: number) => Z,
  isStr: (x: string) => Z,
  isArr: (x: Array<Json>) => Z,
  isOb: (x: Record<string, Json>) => Z
) => (j: Json) => Z = (isNull, isBool, isNum, isStr, isArr, isObj) => (j) => {
  if (j == null) return isNull()
  else if (typeof j === 'boolean') return isBool(j)
  else if (typeof j === 'number') return isNum(j)
  else if (typeof j === 'string') return isStr(j)
  else if (Object.prototype.toString.call(j) === '[object Array]')
    return isArr(j)
  else return isObj(j)
}
```

(mainly taken from: https://github.com/purescript-contrib/purescript-argonaut-core/blob/main/src/Data/Argonaut/Core.js#L30)

What do you think? If you agree I can prepare a PR.

PR #1522, opened 2021-06-25
Add match function to Json module

thought2:
I think this may be useful to destructure Json data structures. 

https://github.com/gcanti/fp-ts/issues/1519

### DIFF
--- a/src/Json.ts
+++ b/src/Json.ts
@@ -60,3 +60,48 @@ export const parse = (s: string): Either<unknown, Json> => parseJSON(s, identity
  */
 // tslint:disable-next-line: deprecation
 export const stringify = <A>(a: A): Either<unknown, string> => stringifyJSON(a, identity)
+
+// -----------------------------------------------------------------------------
+// destructors
+// -----------------------------------------------------------------------------
+
+/**
+ * Takes a function for each possible Json type and returns the result of the one from the position that matches the given data.
+ *
+ * @category destructors
+ * @example
+ * import * as J from 'fp-ts/Json'
+ *
+ * const typeOf = J.match(
+ *   () => 'null',
+ *   () => 'boolean',
+ *   () => 'number',
+ *   () => 'string',
+ *   () => 'array',
+ *   () => 'object'
+ * )
+ *
+ * assert.deepStrictEqual(typeOf(32.2), 'number')
+ * assert.deepStrictEqual(typeOf(['a', 'b']), 'array')
+ *
+ *  @since 2.10.6
+ */
+export const match: <Z>(
+  isNull: () => Z,
+  isBool: (x: boolean) => Z,
+  isNum: (x: number) => Z,
+  isStr: (x: string) => Z,
+  isArr: (x: Array<Json>) => Z,
+  isObj: (x: Record<string, Json>) => Z
+) => (j: Json) => Z = (isNull, isBool, isNum, isStr, isArr, isObj) => (j) =>
+  j === null
+    ? isNull()
+    : typeof j === 'boolean'
+    ? isBool(j)
+    : typeof j === 'number'
+    ? isNum(j)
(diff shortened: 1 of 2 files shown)


samhh:
This is broader than this PR but is there a reason we favour pattern matching in the form of `match(f, g)` instead of `match({ f, g })`?

I've had comments from colleagues I was onboarding to fp-ts previously that match/fold is hard to understand because you don't know which callback aligns with which member of the union or sum type. It's manageable for types with fewer members like `Option` where you can memorise the order, but for a case like this it's going to be very difficult. In a sense, where the types overlap, we're also giving up type safety with the current approach.

Here's a comparison in terms of readability (keeping in mind you'd usually not have the type referenced in the return value!):

```ts
J.match(
  () => 'null',
  () => 'boolean',
  () => 'number',
  () => 'string',
  () => 'array',
  () => 'object'
)

J.match({
  onNull: () => 'null',
  onBoolean: () => 'boolean',
  onNumber: () => 'number',
  onString: () => 'string',
  onArray: () => 'array',
  onObject: () => 'object'
})
```

An additional benefit is that this lets you order the callbacks however you'd like. We could additionally experiment with how to support a universal `_` fallback.

DenisFrezzato:
> is there a reason we favour pattern matching in the form of match(f, g) instead of match({ f, g })?

I guess for verbosity's sake. I'm fine with `match(f, g)` for well known members and as long as there are a couple of cases. For custom unions I usually go with an object where the keys are the name of the members. Example of usage:
```ts
type Err = { _type: 'NotFound', id: string } | { _type: 'SomeException', error: Error }

match({
  NotFound: ({ id }) => `Resource not found with ID "${id}"`,
  SomeException: ({ error }) => error.message
})
```
(I actually use [`ts-adt`](https://github.com/pfgray/ts-adt) for pattern matching, which follows this idea).

src/Json.ts
```diff
@@ -60,3 +60,48 @@ export const parse = (s: string): Either<unknown, Json> => parseJSON(s, identity
...
+ *   () => 'object'
+ * )
+ *
+ * assert.deepStrictEqual(typeOf(32.2), 'number')
+ * assert.deepStrictEqual(typeOf(['a', 'b']), 'array')
+ *
+ *  @since 2.10.6
+ */
+export const match: <Z>(
+  isNull: () => Z,
```
DenisFrezzato, line 90: `is*` is usually used for predicates. I would stick with `on*` to be consistent with match functions of other modules.

DenisFrezzato (changes requested):

src/Json.ts
thought2: good point. Changed it.

thought2:
> This is broader than this PR but is there a reason we favour pattern matching in the form of `match(f, g)` instead of `match({ f, g })`?

Personally I'd favor the `{f, g...}` approach, too. I just used the positional one to stay in sync with other match functions.

thought2:
Btw, I guess you have noticed the `as assertion` needed in the last case. I tried this approach:

```ts
export const match: <Z>(
    onNull: () => Z,
    onBool: (x: boolean) => Z,
    onNum: (x: number) => Z,
    onStr: (x: string) => Z,
    onArr: (x: JsonArray) => Z,
    onObj: (x: JsonRecord) => Z
  ) => (j: Json) => Z = (onNull, onBool, onNum, onStr, onArr, onObj) => (j) =>
    j === null
      ? onNull()
      : typeof j === 'boolean'
      ? onBool(j)
      : typeof j === 'number'
      ? onNum(j)
      : typeof j === 'string'
      ? onStr(j)
      : j instanceof Array
      ? onArr(j)
      : onObj(j)
  ```

But unfortunately this did not pass the typelevel tests for TS version 3.5 :)

DenisFrezzato (approved):

PR #1541, opened 2021-07-26
add Tree.findDepthFirst and Tree.findBreadthFirst and Tree.filter

polymona:
Add some utility functions to `Tree` module.

### DIFF
--- a/src/Tree.ts
+++ b/src/Tree.ts
@@ -32,6 +32,7 @@ import { Predicate } from './Predicate'
 import * as RA from './ReadonlyArray'
 import type { Show } from './Show'
 import type { Traversable1 } from './Traversable'
+import * as O from './Option'
 
 // -------------------------------------------------------------------------------------
 // model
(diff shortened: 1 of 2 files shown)


polymona:
Hi @gcanti !
I put lots of effort into the PR I submitted and I would really appreciate it if you can take a look and give me some suggestions on the code if you have any. Thanks!

kalda341:
This would be really helpful for me too.

One thing to consider would be whether the find and filter should be operating on the tree node, or the value of the node.
I think your approach (the tree node) is the better one as it is more flexible (and similar to operations I've built). Maybe it's worth implementing both (findNodeDepthFirst, findDepthFirst, or similar naming convention)?

When it comes to filter (which is really useful to me), it would probably be worth having both a depth first and breadth first option if we're operating on the node, as when filtering an element your predicate may care about how many filtered children remain, for example.

What do you think? I'm happy to help get this across the line.

polymona:
> Maybe it's worth implementing both (findNodeDepthFirst, findDepthFirst, or similar naming convention)?

Yes, It's valuable.

> When it comes to filter (which is really useful to me), it would probably be worth having both a depth first and breadth first option.

I am confused.  
Are you mean: 
When filter on a tree, there should be two ways to traverse the tree:
1. Apply the predicate to the forest of the node, and then the node's own value.
2. Apply the predicate to the node's own value, if the result is true, then filter the forest of the node recursively. (current implementation)

PR #1545, opened 2021-07-26
struct: alias _.hasOwnProperty to _.has for evolve

imcotton:
Follows up #1481.


imcotton:
BTW: I discovered this via **Rollup** bundle warning:

```
(!) Missing exports
https://rollupjs.org/guide/en/#error-name-is-not-exported-by-module

node_modules/fp-ts/es6/struct.js

hasOwnProperty is not exported by node_modules/fp-ts/es6/internal.js

52:     var out = {};
53:     for (var k in a) {
54:         if (_.hasOwnProperty.call(a, k)) {
                  ^
55:             out[k] = transformations[k](a[k]);
56:         }
```

gcanti:
Thanks @imcotton

PR #1547, opened 2021-07-27
Added to Either docs

bravely:
Adding more examples so certain functions are easier to understand for those new to the library.

Specifically:
* `bimap`
* `chain`
* `chainNullableK`
* `map`
* `mapLeft`
* `tryCatchK`
* `traverseArray`
* `toUnion`

Also, two ideas for doc generation:
* Linking to the function source
* Linking to the function's actual tests, **especially** if there is no example.

I've been finding great joy with this lib, so anything I can do to widen its use would be a boon.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,7 +16,7 @@
 **Note**: A feature tagged as Experimental is in a
 high state of flux, you're at risk of it changing without notice.
 
-# 2.11.0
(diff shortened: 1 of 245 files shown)


samhh:
The two doc generation ideas are good ones, FYI that'd probably be within the remit of [docs-ts](https://github.com/gcanti/docs-ts).

PR #1549, opened 2021-07-28
Object prop function

cdimitroulas:
Reopening this PR as it was automatically closed in https://github.com/gcanti/fp-ts/pull/1454

New feature for `object` module:
- added `prop` function 

Reasoning: Small and handy utility function which is also something that many folks in the JS ecosystem are used to and learn when getting started with functional programming. Usually they use something like `get` from lodash or they see the `prop` function from the Mostly Adequate Guide to FP (https://mostly-adequate.gitbook.io/mostly-adequate-guide/appendix_c#prop).

**Question**: Do you prefer to add overloads so that we can have both curried/uncurried versions of this? (e.g. `prop('myKey', myObject)` or `prop('myKey')(myObject)`)

### DIFF
--- a/docs/modules/struct.ts.md
+++ b/docs/modules/struct.ts.md
@@ -89,8 +89,11 @@ Accesses a property of an object
 **Signature**
 
 ```ts
-export declare function prop<Path extends string, Value, Rest>(path: Path, obj: Record<Path, Value> & Rest): Value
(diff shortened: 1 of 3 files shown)


cdimitroulas:
@gcanti should we close this if there's no interest in having this in the library?

PR #1555, opened 2021-08-05
Add/fix pure comments

OliverJAsh:
After upgrading from 2.10.4 to 2.11.0, our bundle size increased by ~2 KB gzipped. I ran a comparison to see which bundled modules contributed to this change. This inspired me to investigate whether we had any missing or broken pure comments.


gcanti (approved):

gcanti:
Good catch, thank you @OliverJAsh

PR #1560, opened 2021-08-15
The bracket should ignore releasing type on the right

imcotton:
Set to `unknown` would avoid unnecessarily aligning to `void`, so this works:

```ts
import {
    task as T,
    taskEither as TE,
} from 'fp-ts'



const foobar = TE.bracket(
    TE.of(42),
    (a   ) => TE.of(a + 1),
    (_, e) =>  T.of(e),
)

```

### DIFF
--- a/docs/modules/IOEither.ts.md
+++ b/docs/modules/IOEither.ts.md
@@ -1177,10 +1177,10 @@ whether the body action throws (\*) or returns.
 **Signature**
 
 ```ts
-export declare const bracket: <E, A, B>(
+export declare const bracket: <E, A, B, C>(
   acquire: IOEither<E, A>,
   use: (a: A) => IOEither<E, B>,
-  release: (a: A, e: E.Either<E, B>) => IOEither<E, void>
+  release: (a: A, e: E.Either<E, B>) => IOEither<E, C>
 ) => IOEither<E, B>
 ```
 
--- a/docs/modules/ReaderTaskEither.ts.md
+++ b/docs/modules/ReaderTaskEither.ts.md
@@ -1802,10 +1802,10 @@ whether the body action throws (\*) or returns.
 **Signature**
 
 ```ts
-export declare function bracket<R, E, A, B>(
+export declare function bracket<R, E, A, B, C>(
   aquire: ReaderTaskEither<R, E, A>,
   use: (a: A) => ReaderTaskEither<R, E, B>,
-  release: (a: A, e: Either<E, B>) => ReaderTaskEither<R, E, void>
+  release: (a: A, e: Either<E, B>) => ReaderTaskEither<R, E, C>
 ): ReaderTaskEither<R, E, B>
 ```
 
--- a/docs/modules/TaskEither.ts.md
+++ b/docs/modules/TaskEither.ts.md
@@ -1454,10 +1454,10 @@ whether the body action throws (\*) or returns.
 **Signature**
 
 ```ts
-export declare const bracket: <E, A, B>(
+export declare const bracket: <E, A, B, C>(
   acquire: TaskEither<E, A>,
   use: (a: A) => TaskEither<E, B>,
-  release: (a: A, e: E.Either<E, B>) => TaskEither<E, void>
+  release: (a: A, e: E.Either<E, B>) => TaskEither<E, C>
 ) => TaskEither<E, B>
 ```
 
--- a/src/IOEither.ts
+++ b/src/IOEither.ts
@@ -841,10 +841,10 @@ export const fromEitherK: <E, A extends ReadonlyArray<unknown>, B>(
  *
  * @since 2.0.0
  */
-export const bracket = <E, A, B>(
+export const bracket = <E, A, B, C>(
   acquire: IOEither<E, A>,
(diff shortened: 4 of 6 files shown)


DenisFrezzato:
I think using `TaskEither<E, void>` is a better solution, because it signals that the release handler and `bracket` do nothing with whatever is in the `Right` and I'm happy with the compiler asking me to "ignore" the value in the `Right` in an explicit way, if needed.

imcotton:
For example: `chainFirst` or `apFirst` from **TaskEither** doesn't require second be `void` on right part to be ignored later on, seems overly restricted by **bracket** here.

mlegenhausen:
If we choose `chainFirst` as reference we should actually use a new generic variable that is not used in the result or should align all `First` function to use `unknown`.

```ts
export declare const bracket: <E, A, B, C>(
  acquire: TaskEither<E, A>,
  use: (a: A) => TaskEither<E, B>,
  release: (a: A, e: E.Either<E, B>) => TaskEither<E, C>
) => TaskEither<E, B>
```

DenisFrezzato:
I could argue that a type variable like that is useless, `unknown` would be equivalent.

imcotton:
> I could argue that a type variable like that is useless, `unknown` would be equivalent.

That's what I've had in mind while submitting the patch, however on second thought (as @mlegenhausen pointed out) maybe putting into generic would be better because it stays the same rhythm within the code base as a whole.

imcotton:
Apply same changes onto:

- IOEither
- TaskEither
- ReaderTaskEither

xuanduc987:
Personally I don't like this change.
By using `chainFirst`, I am **knowingly** ignore the previous value, the intention was in the function name.
But with this change, there could be a hidden bug due to **unknowingly** ignoring the type problem in release function. 

contrived example

```typescript
declare const acquireBoth: TaskEither<Error, [number, number]>
declare const releaseOne: (s: number) => TaskEither<never, void>

const foobar = TE.bracket(
    acquire,
    ([a, b]) => TE.of(a + b),
    ([a, b]) => pipe(releaseOne(a), TE.map(() => releaseOne(b)),
)
```

imcotton:
I don't think the `void` could save you much here

```ts
pipe(
    releaseOne(a),
    TE.map(() => releaseOne(b)),
    TE.apSecond(releaseOne(a + b)),
)
```

_If you go down that road, one may argue that every **functor** should calling **join** automatically (so we have **Promise**)._

xuanduc987:
As I said, it's just a contrived example to illustrate the problem.

I got bitten too many times because it's too easy to lose value with a language support mixed of statement and expression like JavaScript/TypeScript, thus this change makes me feel uncomfortable.

imcotton:
OK, let's take another angle:

```ts
declare const acquire: () => TaskEither<never, string>
declare const use: (id: string) => TaskEither<Error, Uint8Array>
declare const release: (id: string) => TaskEither<Error, number> // how much resource left

// Error: Type 'number' is not assignable to type 'void'
const job = TE.bracket(acquire, use, release)
```

so you need to additionally do:

```ts
const job = TE.bracket(acquire, use, pipe(release, TE.map(num => void num)))
```

Question: does this extra step actually been beneficial something in the system?

xuanduc987:
The type error seems like a good thing to me.

Because `release` signature doesn't fit, developer must read the document/or the code of `release` to decide what to do with the returned value.

If it deems safe to ignore that value, then they can ignore that easily enough.
But it's also possible that the `release` must be retried until there are 0 resources left, or that the program must panic and exit early.

The point is the type system force the developer to make a conscious choice instead of implicitly swallow a potential problem.

imcotton:
I agree with you, I'm anything but against static typing, it just this very `void` type constrain acting weird, for the referencing:

### **Fluture** ([doc], [dts]) 

```haskell
hook :: Future a b -> (b -> Future c d) -> (b -> Future a e) -> Future a e
```

```ts
export function hook <L, H> (acquire: FutureInstance<L, H>):
                            (dispose: (handle: H) => FutureInstance<any, any>) =>
                        <R> (consume: (handle: H) => FutureInstance<L, R>) =>
                            FutureInstance<L, R>
```

### **ZIO** ([src][zio-src])

```scala
  def bracket[E, A, B](acquire: IO[E, A], release: A => UIO[Any], use: A => IO[E, B]): IO[E, B] =
    ZIO.bracket(acquire, release, use)
```


### **Haskell** ([doc][haskell-doc])

```haskell
bracket :: IO a
  -> (a -> IO b)
  -> (a -> IO c)
  -> IO c
```


[dts]: https://github.com/fluture-js/Fluture/blob/7651bf4a0879a4d5da4fe4d582e0cb0204fd0819/index.d.ts#L134-L135
[doc]: https://github.com/fluture-js/Fluture#hook
[haskell-doc]: https://hackage.haskell.org/package/base-4.15.0.0/docs/Control-Exception-Base.html#v:bracket
[zio-src]: https://github.com/zio/zio/blob/3a354737ae2bff037a77e701bbc8b493e5e88f9a/core/shared/src/main/scala/zio/IO.scala#L49-L53

Issue #1583, opened 2021-09-15
Bind and BindW improperly passes params through


feydan:
## 🐛 Bug report
The following code improperly raises a type error:
```Typescript
  const func1 = (i: {arg1: string}) => TE.right('result1')
  const func2 = (i: {arg2: string, result1: string}) => TE.right('result2')

  pipe(
    TE.right({arg1: 'arg1', arg2: 'arg2'}),
    TE.bindW('result1', func1),
    TE.bindW('result2', func2),
  )
```
This is because this line improperly rewrites the parameter list that gets passed into the next function in the pipe `TE.bindW('result1', func1),`

### Current Behavior

Improper type error

### Expected behavior

The params should be passed through using the return type of the previous pipe call rather than the param list of the function passed in.

### Reproducible example
Example above

### Suggested solution(s)

I'll open an MR to fix this following this bug issue.

### Your environment

fp-ts 2.11.2

| Software   | Version(s) |
| ---------- | ---------- |
| fp-ts      |      2.11.2      |
| TypeScript |     4.2.3       |

jrmdayn:
+1

I keep hitting the same problem. This has been open for quite some time, with a PR to fix it.. What is preventing the fix to be released?

Thx

gcanti:
Reopening because the fix causes a regression (https://github.com/gcanti/fp-ts/pull/1584#issuecomment-1252053927)

thewilkybarkid:
@gcanti The ts-pattern problem is confirmed as a limitation in TypeScript (https://github.com/gvergnaud/ts-pattern/issues/111#issuecomment-1261961683). I have a workaround for when the change is made, and it's definitely one to do. (Just not sure whether it should be left for v3.)

georgegebbett:
The release notes for v2.13.1 suggest that this was fixed in that release but I am still seeing this behaviour after updating and the comments here suggest the fix was removed from the release?

Could you confirm that it is the release notes that are incorrect and I am not going mad lol?

gcanti:
@georgegebbett sorry, I forgot to amend the changelog after reverting the change (the fix was actualy removed from the release)

georgegebbett:
That makes what I am seeing make sense! Thanks for clarifying!

chao7150:
With TypeScript 5.4, `f: (a: NoInfer<A>) => TaskEither<E2, B>` might solve the problem.

For example
```typescript
export const bindW: <N extends string, A, E2, B>(
[...]

PR #1584, opened 2021-09-15
Ensuring that bind and bindW pass the correct params through rather t…

feydan:
…han get rewritten by the function passed in

Fixes https://github.com/gcanti/fp-ts/issues/1583

The following code improperly raises a type error:
```Typescript
  const func1 = (i: {arg1: string}) => TE.right('result1')
  const func2 = (i: {arg2: string, result1: string}) => TE.right('result2')

  pipe(
    TE.right({arg1: 'arg1', arg2: 'arg2'}),
    TE.bindW('result1', func1),
    TE.bindW('result2', func2),
  )
```
This is because this line improperly rewrites the parameter list that gets passed into the next function in the pipe `TE.bindW('result1', func1),`

This MR changes the passed in function to have its own generic that extends the param list.  This allows the params to flow through as expected.

edit: formatting


feydan:
Unfortunately, this seems to cause another issue:
```typescript
// proper param passthrough
    const c = (p: { a: number }) => _.right<string, number>(p.a)
    const d = (p: { b: string }) => _.right<string, string>(p.b)
    U.deepStrictEqual(
      await pipe(
        _.right<string, number>(1),
        _.bindTo('a'),
        _.bind('b', () => _.right('b')),
        _.bind('c', c),
        _.bind('d', d),
        _.bind('e', _.fromIOK(i => () => 'io')) // Type 'A2' is not assignable to type 'A2'. Two different types with this name exist, but they are unrelated.
      )(),
      E.right({ a: 1, b: 'b', c: 1, d: 'b', e: 'io' })
    )
```
Looking into alternatives.

feydan:
> Unfortunately, this seems to cause another issue:
> 
> ```ts
> // proper param passthrough
>     const c = (p: { a: number }) => _.right<string, number>(p.a)
>     const d = (p: { b: string }) => _.right<string, string>(p.b)
>     U.deepStrictEqual(
>       await pipe(
>         _.right<string, number>(1),
>         _.bindTo('a'),
>         _.bind('b', () => _.right('b')),
>         _.bind('c', c),
>         _.bind('d', d),
>         _.bind('e', _.fromIOK(i => () => 'io')) // Type 'A2' is not assignable to type 'A2'. Two different types with this name exist, but they are unrelated.
>       )(),
>       E.right({ a: 1, b: 'b', c: 1, d: 'b', e: 'io' })
>     )
> ```
> 
> Looking into alternatives.

I think I found a solution -- the change is committed.

src/Chain.ts
```diff
@@ -115,49 +115,49 @@ export function bind<M extends URIS4>(
   M: Chain4<M>
 ): <N extends string, A, S, R, E, B>(
   name: Exclude<N, keyof A>,
-  f: (a: A) => Kind4<M, S, R, E, B>
+  f: <A2 extends A>(a: A | A2) => Kind4<M, S, R, E, B>
```
gcanti, line 118: if `A2 extends A` isn't `A | A2` redundant?

```diff
-f: <A2 extends A>(a: A | A2) => Kind4<M, S, R, E, B>
+f: <A2 extends A>(a: A2) => Kind4<M, S, R, E, B>
```

gcanti (changes requested):

src/Chain.ts
feydan: Thank you for the feedback @gcanti.  I tried removing A according to your suggestion here, and I got the following error in line 633 of `test/TaskEither.ts`.

```
Argument of type '<E>(i: A2) => TaskEither<E, number>' is not assignable to parameter of type '<A2 extends { readonly a: number; readonly b: string; readonly c: number; readonly d: string; }>(a: A2) => TaskEither<unknown, number>'.
  Types of parameters 'i' and 'a' are incompatible.
    Type 'A2' is not assignable to type 'A2'. Two different types with this name exist, but they are unrelated.
      'A2' is assignable to the constraint of type 'A2', but 'A2' could be instantiated with a different subtype of constraint '{ readonly a: number; readonly b: string; readonly c: number; readonly d: string; }'
```

I also tried `<A2 extends A = A>(a: A2) => Kind4<M, S, R, E, B>`, but got the same error.  I unfortunately can't claim to understand the nuance here.

feydan: I played around with your suggestion a bit more.  I changed the test to use `fromOptionK` instead of `fromIOK`, and it works  with your suggested changes as expected now.  I'm not sure I understand the signature for fromIOK.

I've updated all occurrences of `A | A2` to be just `A2`.

gcanti: > I changed the test to use fromOptionK instead of fromIOK

If you remove the type annotation here

```diff
-_.bind('e', _.fromOptionK(() => 'err')((p: { readonly c: number }) => some(p.c)))
+_.bind('e', _.fromOptionK(() => 'err')((p) => some(p.c)))
```

`fromOptionK` raises an error too. 

Looks like I was wrong and `A | A2` is necessary, though the reason why is not clear to me

feydan: Thank you for the follow-up!  I changed the type signature back to use `A | A2`, separated out the param passthrough test, and added a case to provide coverage for `bindW`

gcanti:
`npm run fix-prettier` should fix the CI failure

gcanti (approved):

gcanti:
Thank you @feydan

thewilkybarkid:
I'm trying out `2.13.0-rc.4`, and this change seems to cause a type of failure in an existing (working) project. I'm seeing if I can create a reproducible failure/fix.

thewilkybarkid:
So the failure is related to inference and ts-pattern in  https://github.com/PREreview/prereview.org/blob/b9f3de1477e2904fcbeee2ed70cc7e1aa7a24a6e/src/infrastructure.ts#L145-L153

Seems like the union type causes the exhaustiveness check to fail.

I can extract a typed function which gets around the issue, but this wasn't possible before this change. 😃

thewilkybarkid:
[Reproduction with ts-pattern](https://www.typescriptlang.org/play?#code/JYWwDg9gTgLgBAbzhMNgQHZwIYGc4DycAvnAGZQQhwDkZYAtDLjQFCiSyJwjYwDGACxLlK1GswZg+MAKZQMbdhjlQy2frLgAVAJ5gtCVnDhRZ2ACaYANrrjALALlqZZNOAB9aMAO4Q2xKysAPTBcADq0ADWuAA0cABGAK7w-NhJuLIWcGCUCdayIPjWwFFagjAwYLiOoQDmwDCCSQkAdPxUwXVpKsDB9Ey4wcC4uEmyQwCMAKwAHADM9lgDzHAAPABMrZPzrGRJGPxomHBW2oLAGHWTABTYznoGAJTOBK0EqOgYaxhJIAnyAB8iGMplkMCSCh4fCEdyeoJMrR8jUENyQDmcNFc7mI8RuTzgAF5gW9cFRZDdJk94SZEcimmj7E5vH4cXiCcTCK0ySAKRtqQi4K1ZAAPQTpXBoABuFPhgRCYQASrJchALEkjl9kGQ4E0tOpgNZIVpLnAKlUavUUS12p1uthev1GMxgmAktZrMEZrMACwAYhGY1kHRAvJUDEmG2mGwADAsfT7dvtDscsGcLlcNmtHrINnBRXIMBZ8DnAXcHvotF4c-zXu9Ppgfn8AVBgUYTGYIVDeAJUdgabSkSjGRiXBg3CR2USSdzyZSBbShfTUejmRJWZO4Pjp1yeXyF4PReKMtLZax5aE4ABhB1wNIZLSSaSVeRYGAQcjYQ1wZdwABWJ66oIWh1LI45QMA-CtDoFz4LgggQO62QAj+0QAPxbjQDjuD0EiJFoDq6E0lx1MgepQEBt5YuO7jQCy-hofCyaaic6YkfM2aVnmBZgcWOiVmW9z8QYtZch8qZNv8QIgh24KQlgPawv2gpDgyq6Ytim7bpypJzlSA50sO6n0WyW4cjOe43PyBlCkeEqnvi56sEAA)

I'll open a ts-pattern issue due to the problem there with generics.

Issue #1507, opened 2021-05-16
Reduce the supported number of arguments to pipe function


pbadenski:
### Current Behavior

85b51d1 increased number of arguments to pipe function. However, it seems that there's an odd performance cliff in V8  when more 10 "case" statements are used. I am not entirely clear what's going on under the hood and I inquired on v8-users google group for details (https://groups.google.com/g/v8-users/c/_Add2xOVmfU). Performance loss is substantial - I noticed between 2 to 4 times deterioration.

### Reproducible example

```
const foo = (obj: { a: { b: string } }) => {
  return pipe(
    option.fromNullable(obj.a),
    option.map(x => x.b)
  );
}

const bar = (obj) => {
  return pipe(
    foo(obj),
    option.chain(x => option.some(x))
  );
}

let x = bar(obj);
```

Results:
```
pipe; Cases = 10 x 35,191,983 ops/sec ±0.64% (89 runs sampled)
pipe; Cases > 10 x 14,563,165 ops/sec ±0.74% (87 runs sampled)
```

### Suggested solution(s)

Personally I'd be interested in seeing number of supported pipe arguments reduced back to 10, especially since the performance impact seems substantial. I understand that this would be a backwards incompatible change. I would argue that pipe with more than 10 arguments are rather infrequent, but other people may have a different view.

### Additional context

This is especially relevant since non-pipeable methods delegate to pipe, ie:
```
var _chain = function (ma, f) { return function_1.pipe(ma, exports.chain(f)); };
```

### Your environment

| Software   | Version(s) |
| ---------- | ---------- |
| fp-ts      | 2.10.4           |
| TypeScript |      4.1.5      |

steida:
@pbadenski Is it possible to warn with the eleventh arg?

pbadenski:
I actually thought of alternative suggestion - keep 1-9 case hand-rolled, and add default case to handle list of arguments by reducing them with a for-loop (similar to current `sequenceS` and `sequenceT` implementations)

kylegoetz:
Is there a reason it needs to use case at all? When I implemented something like `pipe` years ago, IIRC it was basically just `pipe = [arg1, ..., argn].reduce((acc, fn) => fn(acc), arg0)`.

steida:
@kylegoetz Types

kylegoetz:
@steida export a bunch of declared functions but the actual real function type does not get exported. I thought the codebase already did this for pipe. My mistake.

[...]

PR #1585, opened 2021-09-16
pipe: fix v8 performance cliff for >10 case stmts

pbadenski:
V8 has a performance cliff when there's more than 10 case blocks in a switch
statement. This fix reduces the number of fast specialised case statements
to 9 + extra one for general case.

More details here: https://groups.google.com/g/v8-users/c/_Add2xOVmfU/m/IQHh2IAVBAAJ

Fixes: https://github.com/gcanti/fp-ts/issues/1507

### DIFF
--- a/src/function.ts
+++ b/src/function.ts
@@ -651,18 +651,7 @@ export function pipe(
   ef?: Function,
   fg?: Function,
   gh?: Function,
-  hi?: Function,
-  ij?: Function,
-  jk?: Function,
-  kl?: Function,
-  lm?: Function,
-  mn?: Function,
-  no?: Function,
-  op?: Function,
-  pq?: Function,
-  qr?: Function,
-  rs?: Function,
-  st?: Function
+  hi?: Function
 ): unknown {
   switch (arguments.length) {
     case 1:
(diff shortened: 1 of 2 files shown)


src/function.ts
```diff
@@ -683,30 +672,17 @@ export function pipe(
...
-      return op!(no!(mn!(lm!(kl!(jk!(ij!(hi!(gh!(fg!(ef!(de!(cd!(bc!(ab!(a)))))))))))))))
-    case 17:
-      return pq!(op!(no!(mn!(lm!(kl!(jk!(ij!(hi!(gh!(fg!(ef!(de!(cd!(bc!(ab!(a))))))))))))))))
-    case 18:
-      return qr!(pq!(op!(no!(mn!(lm!(kl!(jk!(ij!(hi!(gh!(fg!(ef!(de!(cd!(bc!(ab!(a)))))))))))))))))
-    case 19:
-      return rs!(qr!(pq!(op!(no!(mn!(lm!(kl!(jk!(ij!(hi!(gh!(fg!(ef!(de!(cd!(bc!(ab!(a))))))))))))))))))
-    case 20:
-      return st!(rs!(qr!(pq!(op!(no!(mn!(lm!(kl!(jk!(ij!(hi!(gh!(fg!(ef!(de!(cd!(bc!(ab!(a)))))))))))))))))))
+    default:
```
gcanti, line 675: what about

```ts
      let ret = arguments[0]
      for (let i = 1; i < arguments.length; i++) {
        ret = arguments[i](ret)
      }
      return ret
```

?

pbadenski: duh.. yes, I'll change to your suggestion. I knew I was missing something as I was writing it.

gcanti:
Thanks @pbadenski, released

PR #1601, opened 2021-10-12
Edit docs for Foldable in Array.ts.md

Nickphammond:
Current documentation implies for the Foldable methods, the array is curried when it isn't.

**Note**. If you've fixed a bug please link the related issue or, if missing, open an issue before sending a PR.

**Note**. If you want to send a PR related to `fp-ts@1.x` please create your branch from `1.x`

### DIFF
--- a/docs/modules/Array.ts.md
+++ b/docs/modules/Array.ts.md
@@ -410,7 +410,7 @@ Added in v2.0.0
 **Signature**
 
 ```ts
-export declare const foldMap: <M>(M: Monoid<M>) => <A>(f: (a: A) => M) => (fa: A[]) => M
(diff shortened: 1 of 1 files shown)


DenisFrezzato:
The documentation is generated from the source code. If we change it manually, the change will be lost next time it's generated.

Anyway, the current documentation seems correct to me. What version of `fp-ts` are you using?

Nickphammond:
Oh ok that makes sense.

I'm not sure what I'm doing wrong then. My version is 2.11.2. 
As it is currently written, shouldn't I be able to do something like:
`A.Foldable.reduce("s", (str:string, num:number) => str)([1])`

Which fails, instead I have to do:
`A.Foldable.reduce([1], "s", (str:string, num:number) => str)`

PR #1605, opened 2021-10-20
chore: improve task execution order

mikearnaldi:
Some have noticed that the order of execution changed after the microtask PR got introduced, while there is no guarantee of order in case of parallel combinators it would still be good to preserve it if possible.

This PR introduce an internal task scheduler that uses mictotasks but avoids oversuspension, it basically suspends once and executes everything that can be scheduled in a loop until nothing else is to be done, this mechanisms preserves the order of scheduling because scheduling is sync within a single suspension


mikearnaldi:
cc @StefanoMagrassi @gcanti

PR #1618, opened 2021-11-14
`StateT`: Add missing category marker to `StateT3`

cwstra:
StateT3 is in the utils section of the documentation page at present, which I presume is due to it not having a category marker.

### DIFF
--- a/docs/modules/StateT.ts.md
+++ b/docs/modules/StateT.ts.md
@@ -16,8 +16,8 @@ Added in v2.0.0
   - [StateT (interface)](#statet-interface)
   - [StateT1 (interface)](#statet1-interface)
   - [StateT2 (interface)](#statet2-interface)
-- [utils](#utils)
   - [StateT3 (interface)](#statet3-interface)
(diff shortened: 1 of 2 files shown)


grossbart:
Thanks for this patch, @cwstra, this is indeed missing the category. Can you please do two more things:

1. Run `npm run docs` to generate the new Markdown docs.
2. Then commit the generated `StateT.ts.md` file as well.

You can check whether this has the desired effect if you enable GitHub Pages at https://github.com/cwstra/fp-ts/settings/pages and then push to your repo. There you should be able to see the updated docs with StateT3 in the correct category.

cwstra:
Done. I did notice that one of the examples failed when I ran `docs`: `src-Task.ts-function-delay-0.ts`. That _might_ be because I'm running node 12 at present? This PR obviously doesn't touch that.

grossbart:
Thanks @cwstra! Not sure what is going on with the test, it ran through without issues on CI (also on Node 12 …).

PR #1625, opened 2021-12-02
Add Either.traverseOption

vinassefranche:
Let me know if it's ok to add this. I'd be willing to do the same for most of the other modules (`TaskEither`, `Task`, `Reader`....)

### DIFF
--- a/docs/modules/Either.ts.md
+++ b/docs/modules/Either.ts.md
@@ -134,6 +134,7 @@ Added in v2.0.0
   - [toError](#toerror)
   - [traverseArray](#traversearray)
   - [traverseArrayWithIndex](#traversearraywithindex)
+  - [traverseOption](#traverseoption)
   - [traverseReadonlyArrayWithIndex](#traversereadonlyarraywithindex)
   - [traverseReadonlyNonEmptyArrayWithIndex](#traversereadonlynonemptyarraywithindex)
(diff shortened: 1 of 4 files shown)


mohaalak:
I don't think this is a good idea you can easily this function in your app, if we go down this road there will multiple combination of this traverse.

traverseArray is there because of faster and stack safety.

vinassefranche:
@mohaalak Thanks for you message. Here are a few questions/remarks on what you said:

> I don't think this is a good idea you can easily this function in your app, if we go down this road there will multiple combination of this traverse.

I honestly don't think that 'being able to easily add the function in the app' and 'there will multiple combination' are good reasons not to do that. We added `chainReaderK`, `chainReaderTaskK`, `chainReaderEitherK` ... to `ReaderTaskEither` (and same for `ReaderEither`, `ReaderTask`...) and I think that all could have been easily add in each app instead of here and there are multiple combinations.
My main concern about this addition, though, is that it makes `Option` a specific Module for no reason. Because of circular dependencies, it would be impossible to add `Option.traverseEither` (apart from re-writing the traverse method without needing the other module, which I don't think is a good idea) and I don't know why we would prefer to add one and not the other.

> traverseArray is there because of faster and stack safety.

I understand the 'faster' part but could you elaborate about the 'stack safety'? I don't think I understand what you mean.

PR #1637, opened 2021-12-22
Narrow Json error types

thewilkybarkid:
`JSON.parse` and `JSON.stringify` only throw [`SyntaxError`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/parse#exceptions) and [`TypeError`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify#exceptions) respectively, so we can narrow the types in the Json module.

There is one custom error in `stringify`, but this is better represented as a `TypeError` anyway.

### DIFF
--- a/docs/modules/Json.ts.md
+++ b/docs/modules/Json.ts.md
@@ -62,7 +62,7 @@ Converts a JavaScript Object Notation (JSON) string into a `Json` type.
 **Signature**
 
 ```ts
-export declare const parse: (s: string) => Either<unknown, Json>
+export declare const parse: (s: string) => Either<SyntaxError, Json>
 ```
 
 **Example**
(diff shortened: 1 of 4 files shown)


enricopolanski:
I like the proposal, I think though that we are missing a case?

```
JSON.stringify({x: 2n});
// TypeError: BigInt value can't be serialized in JSON
```

Should we handle both the TypeErrors messages rather than only cyclic references?

thewilkybarkid:
Having tried this out for a bit, these types are a bit problematic: `TypeError` and `SyntaxError` are just aliases of `Error`.

So, I'd probably say to create new error types, or just use `Error` (which is better than the original `unknown`).

PR #1638, opened 2021-12-23
Performance improvement for Task

sahincapiec:
I removed some unnecessary promises creations. Thus reducing tick's calls and increasing the performance.

### DIFF
--- a/src/Task.ts
+++ b/src/Task.ts
@@ -88,11 +88,8 @@ export const fromIO: FromIO1<URI>['fromIO'] = (ma) => () => Promise.resolve().th
 export function delay(millis: number): <A>(ma: Task<A>) => Task<A> {
   return (ma) => () =>
     new Promise((resolve) => {
-      setTimeout(() => {
-        // tslint:disable-next-line: no-floating-promises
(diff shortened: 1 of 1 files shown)


xuanduc987:
It was intentional https://github.com/gcanti/fp-ts/pull/1591

sahincapiec:
Yes. I understand, but it's still working with my changes. It's stack-safe but faster. There were many redundant and unnecessary promise creations.

xuanduc987:
Only `chain`, which is not touched by this PR is stack-safe.
You can see the stack blow up with
```ts
let pipeline = T.of(0)
for (let i = 0; i < 55000; i++) {
  pipeline = pipe(
    pipeline,
    T.map(n => n + 1)
  )
}
await pipeline()

```

enricopolanski:
Just a small hint imho: when opening PRs that contain performance changes please provide the numbers in the PR.

pbadenski:
Could someone explain what happened that this PR was closed? I haven't quite understood the conversation. I've been looking at #1591 and wondering about performance impact in promise heavy environment (my preliminary micro-benchmarks show stack-safe impl. to be 2 x slower).

PR #1641, opened 2021-12-25
StateEither monad

thomasvargiu:
Added `StateEither`


thomasvargiu:
Just saw it's implemented in `fp-ts-contrib`

PR #1664, opened 2022-01-20
feat(semigroup): add partial semigroup

JalilArfaoui:
I suggest this `partial` semigroup generator, inspired by `struct` but that can take partial elements


gcanti:
@JalilArfaoui It seems to me that throughout the codebase, the absence of a value is modeled through `Option` rather than optional properties. I won't go into the details of this implementation, but adding support for optional properties would be an exception. 

In any case, development is frozen until a yet-to-be-determined date when I am clear on which direction to take after finishing developments in the new org `@fp-ts`.

JalilArfaoui:
Thanks @gcanti. 

I think I undertand your point, but I find partials are useful to work on … «partial data structures», where the absence of value does not mean `None`, it just mean we don’t care … for instance when trying to merge configuration options or some metadata … 

To me `Option` type and `Partial` serve different purpose …

I didn’t knew some big changes are in development … I’ll go check out on the Discord

JalilArfaoui:
OK I just found out https://github.com/fp-ts/core …

PR #1676, opened 2022-02-14
try/catch in tryCatch

### LINKED ISSUE #1672: TaskEither.tryCatch should either try/catch or be renamed
### Current Behavior

`TE.tryCatch` only handles promises which have been rejected.

### Desired Behavior

It should have the same semantics as a javascript `try`/`catch` construct. 

Or it should be deprecated and renamed to something not misleading, such as `TE.catchRejectedPromise`

### Suggested Solution

Wrapping the code in a try/catch, or using an async function and try/catch block.

All users.

Wrapping `TE.tryCatch` in a try/catch.

### Additional context

I did read the docs where it said that `r` cannot throw.

### Your environment

| Software   | Version(s) |
| ---------- | ---------- |
| fp-ts      |   2.11.8         |
| TypeScript |   4.5.5         |

thewilkybarkid:
This reopens and updates #1001 to allow `f` in `TaskEither` and `TaskOption`'s `tryCatch`/`tryCatchK` to throw an error (as opposed to only returning a resolved/rejected promise).

Fixes #1672.

### DIFF
--- a/src/TaskEither.ts
+++ b/src/TaskEither.ts
@@ -254,8 +254,6 @@ export const getOrElseW: <E, B>(
 /**
  * Transforms a `Promise` that may reject to a `Promise` that never rejects and returns an `Either` instead.
  *
- * Note: `f` should never `throw` errors, they are not caught.
- *
  * See also [`tryCatchK`](#trycatchk).
  *
  * @example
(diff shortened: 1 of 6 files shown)


src/TaskEither.ts
```diff
@@ -272,8 +270,16 @@ export const getOrElseW: <E, B>(
...
  * @since 2.0.0
  */
-export const tryCatch = <E, A>(f: Lazy<Promise<A>>, onRejected: (reason: unknown) => E): TaskEither<E, A> => () =>
-  f().then(_.right, (reason) => _.left(onRejected(reason)))
+export const tryCatch = <E, A>(
+  f: Lazy<Promise<A>>,
+  onRejected: (reason: unknown) => E
+): TaskEither<E, A> => async () => {
+  try {
+    return await f().then(_.right)
```
matthewpflueger, line 278: Isn't return await an anti-pattern and causes performance issues (https://arthur.place/the-cost-of-return-await)?

bwbuchanan: return-await is necessary in this case to catch promise rejections (there's a section about this in the article you cited)

matthewpflueger: Oh, good catch!

PR #1677, opened 2022-02-20
Update learning-resources.md

SimonAM:
Added community documentation. 
Added author names


SimonAM:
I would suggest an open issue for learning resources suggestions that would also allow for some kind of peer review, rather than creating pull requests as i have. Suggestions on learning resources flows in faster via comments than pull requests i would think.  

There is an open issue on documentation
[issue 1103](https://github.com/gcanti/fp-ts/issues/1103)
However, i would suggest a new issue specifically asking for learning resources links.

grossbart:
Thanks for your edits, @SimonAM! Regarding your idea for opening an issue for learning resources: how about a dedicated discussions category instead and linking to there from this page?

PR #1679, opened 2022-03-03
Add intercalate to Monoid

thewilkybarkid:
Adds `intercalate` to `Monoid`, to match `Semigroup`.

### DIFF
--- a/docs/modules/Monoid.ts.md
+++ b/docs/modules/Monoid.ts.md
@@ -44,6 +44,7 @@ Added in v2.0.0
 <h2 class="text-delta">Table of contents</h2>
 
 - [combinators](#combinators)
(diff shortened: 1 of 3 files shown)


gcanti:
You don't get a `Monoid` as output, counterexample:

```ts
import { Monoid } from 'fp-ts/Monoid'
import * as Se from 'fp-ts/Semigroup'
import * as S from 'fp-ts/string'

const intercalate = <A>(middle: A) => (M: Monoid<A>): Monoid<A> => ({
  empty: M.empty,
  concat: Se.intercalate(middle)(M).concat
})

const IM = intercalate(' + ')(S.Monoid)

console.log(IM.concat(IM.empty, 'a')) // => ' + a', should be 'a'
```

PR #1681, opened 2022-03-11
Refinement overload added to 'every' functions

mlegenhausen:
I have added `Refinement` overloads to the `every` function of `ReadonlyArray`,  `ReadonlyRecord`, `ReadonlySet` and their mutable counterparts.

Example

```ts
import { every } from 'fp-ts/ReadonlyArray'
import { isNumber } from 'fp-ts/number'

declare const as: ReadonlyArray<string | number>
every(isNumber)(as) // ReadonlyArray<number>
```

### DIFF
--- a/docs/modules/Array.ts.md
+++ b/docs/modules/Array.ts.md
@@ -3313,7 +3313,10 @@ Added in v2.0.0
 **Signature**
 
 ```ts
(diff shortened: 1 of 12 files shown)


gcanti:
Thanks @mlegenhausen, released.

PR #1693, opened 2022-04-06
Make JsonRecord type safe

thewilkybarkid:
The `JsonRecord`'s type currently allows you to access _any_ string key, but [this isn't type safe](https://www.typescriptlang.org/play?#code/JYWwDg9gTgLgBAbzgKwM4QHZwIargKTgF84AzKCEOAclLAFoZVqAoFgY01XigFNOoAEwBcBAHT50GAEr9oguAF5ELODlHVs1ADQsibThm5wwFML1gBPAIKj8EqUrh8BgsdjgB6T3BgALYDxAsmAMXhweAFcMGFBeDi54UwhzKwAhOwdMJxd5MQAjLx9-YOCMCBgxKpYgA). Looking at the commit history the type came from https://github.com/Microsoft/TypeScript/issues/1897#issuecomment-338650717, and does seem to have been copied over to many other libraries.

This PR changes the values of a `JsonRecord` to also possibly be `undefined`, covering the fact that the key might not be set.

This does mean that it's possible to set keys as `undefined`, even though this type doesn't exist in JSON (`stringify` will throw these keys away). This actually fixes my problem in https://github.com/gcanti/io-ts/pull/628#issuecomment-1084597642 (which was complicated to try and get round in https://github.com/gcanti/io-ts/pull/639).

### DIFF
--- a/docs/modules/Json.ts.md
+++ b/docs/modules/Json.ts.md
@@ -49,7 +49,7 @@ Added in v2.10.0
 
 ```ts
 export interface JsonRecord {
-  readonly [key: string]: Json
+  readonly [key: string]: Json | undefined
 }
 ```
 
--- a/dtslint/ts3.5/Json.ts
+++ b/dtslint/ts3.5/Json.ts
@@ -2,6 +2,15 @@ import * as E from '../../src/Either'
 import { pipe } from '../../src/function'
 import * as _ from '../../src/Json'
 
+declare const jr: _.JsonRecord
+
+//
+// JsonRecord
+//
+
+// $ExpectType string | number | boolean | JsonRecord | JsonArray | null | undefined
+jr.foo
+
 //
 // stringify
 //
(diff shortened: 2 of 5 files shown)


thewilkybarkid:
[type-fest has the same type (in a different syntax).](https://github.com/sindresorhus/type-fest/blob/043b732bf02c2b700245aa6501116a6646d50732/source/basic.d.ts#L22)

thewilkybarkid:
dtslint failed in CI; locally it was only running up to TS 4.1. Looks like the library isn't pinned so.

thewilkybarkid:
Ah, that's definitely something to enable generally (and does fix the accessing problem). Thanks.

It doesn't, however, get around https://github.com/gcanti/io-ts/pull/628. Should I close this and reopen https://github.com/gcanti/io-ts/pull/639?

PR #1696, opened 2022-04-08
add `IOOption` module, closes #1694

### LINKED ISSUE #1694: Make IOOption part of fp-ts
### Current Behavior

ATM `IOOption` is part of `fp-ts-contrib`

### Desired Behavior

Move `IOOption` as part of `fp-ts`

### Suggested Solution

Move the code to `fp-ts`

`fp-ts-contrib` users will be impacted but migration will be painless.

This is for `fp-ts` users to allow easier and better usage of Option and IO for scenarios that implement if/then patterns.

`IOEither` can work as a replacement but the `left` in many cases it is not required and causes a bit of noise in the implementation.

### Additional context

Since `Option` and `Either` both exist in `fp-ts` (and correctly so IMHO) I would argue that `IOOption` and `IOEither` - that are also closely related - should also both exist in `fp-ts`.

Currently only `IOEither` is in `fp-ts` and `IOOption` is in `fp-ts-contrib`

### Your environment

N/A

| Software   | Version(s) |
| ---------- | ---------- |
| fp-ts      |            latest |

gcanti:


src/IOOption.ts
```diff
@@ -0,0 +1,728 @@
...
+export const fromIOEither: NaturalTransformation21<IEURI, URI> =
+  /*#__PURE__*/
+  I.map(O.fromEither)
+
+// -------------------------------------------------------------------------------------
+// destructors
+// -------------------------------------------------------------------------------------
+
+/**
+ * @category destructors
```
gkamperis, line 118: would a `fold` make sense?

gcanti: @gkamperis in all other modules `fold` is just an alias of `match` (or `matchE`) now

gkamperis: @gcanti yes saw that... but since it existed in `IOEither` thought I'd mention it to see if we should keep the pattern...

PR #1721, opened 2022-06-11
feat(Apply/ap): make total overloads `F: 0,1,2,2C,3,3C,4` to `G: 1,2,2C,3,3C,4`

waynevanson:
There are some overloads missing that I needed in in my projects, and figured we may as well add all the overloads for the functions.

Questions:
- What is preferred, A => B or A1 => A2? I prefer the latter because there's no good standard on letters after Q.
- Could we do A1 instead of A? I've been using vim search and replace and it's much easier when they all have numbers.

### DIFF
--- a/src/Apply.ts
+++ b/src/Apply.ts
@@ -110,6 +110,12 @@ export interface Apply4<F extends URIS4> extends Functor4<F> {
  * @category combinators
  * @since 2.10.0
  */
+export function ap<F extends URIS3, G extends URIS2>(
+  F: Apply3<F>,
+  G: Apply2<G>
+): <R, E1, E2, A1>(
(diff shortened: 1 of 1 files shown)


waynevanson:
@gcanti Could I get a rerun on this one?

waynevanson:
@gcanti Please review when you get the chance, it ran.

gcanti:
> What is preferred, A => B or A1 => A2?

For consistency with all the other signatures I'd keep `A => B`

waynevanson:
> 

I copied the pattern that I saw in the Functor module.

waynevanson:
> > What is preferred, A => B or A1 => A2?
> 
> For consistency with all the other signatures I'd keep `A => B`

@gcanti I've used the Functor module as the reference for the syntax.

PR #1725, opened 2022-06-14
Functor: add `bindMap` function as a do-notation helper

Punie:
Hello 👋 

I took the liberty to create a new function on the `Functor` typeclass to help with introducing new values in do-notation context while leveraging pure computations that depend on said context.

ie. I needed to do something akin to this (example greatly simplified from my original genuine use-case):
```ts
pipe(
  option.Do,
  option.apS('x', some(42)),
  option.apS('y', some(1337)),
  option.bind('z', ({ x, y }) => option.of(x * y)),
); // => Option<{ x: number; y: number; z: number }>
```

Another way to achieve the same result would be to just use `option.map` at the last step:
```ts
pipe(
  option.Do,
  option.apS('x', some(42)),
  option.apS('y', some(1337)),
  option.map(factors => ({
    z: factors.x * factors*y,
    ...factors,
  })),
); // => Option<{ x: number; y: number; z: number }>
```

I find both these solutions unsatisfying so I introduced the `bindMap` function on `Functor` specifically for those use-cases. It is to `map` what `bind` is to `chain` and `apS` is to `ap`.

With this, we can rewrite the example above as such (getting rid of the `option.of` or the recreation of the context object):
 ```ts
pipe(
  option.Do,
  option.apS('x', some(42)),
  option.apS('y', some(1337)),
  option.bindMap('z', ({ x, y }) => x * y),
); // => Option<{ x: number; y: number; z: number }>
```

I hope this can be of use to others as well. I'm not too fond of the name `bindMap` but it's the best I could think of. I kinda liked `let` but as a reserved keyword in TS it was a no-go.

### DIFF
--- a/docs/modules/Array.ts.md
+++ b/docs/modules/Array.ts.md
@@ -188,6 +188,7 @@ Added in v2.0.0
   - [Spanned (interface)](#spanned-interface)
   - [apS](#aps)
   - [bind](#bind)
+  - [bindMap](#bindmap)
   - [bindTo](#bindto)
   - [deleteAt](#deleteat)
   - [elem](#elem)
@@ -3287,6 +3288,19 @@ export declare const bind: <N, A, B>(
 ```
 
 Added in v2.8.0
+
+## bindMap
+
+**Signature**
+
+```ts
+export declare const bindMap: <N, A, B>(
+  name: N,
+  f: (a: A) => B
+) => (fa: A[]) => { readonly [K in N | keyof A]: K extends keyof A ? A[K] : B }[]
+```
+
+Added in v2.13.0
 
 ## bindTo
 
--- a/docs/modules/Array.ts.md
+++ b/docs/modules/Array.ts.md
@@ -188,7 +188,6 @@ Added in v2.0.0
   - [Spanned (interface)](#spanned-interface)
   - [apS](#aps)
   - [bind](#bind)
-  - [bindMap](#bindmap)
   - [bindTo](#bindto)
   - [deleteAt](#deleteat)
   - [elem](#elem)
@@ -200,6 +199,7 @@ Added in v2.0.0
   - [insertAt](#insertat)
   - [intercalate](#intercalate)
   - [isOutOfBound](#isoutofbound)
+  - [let](#let)
   - [lookup](#lookup)
   - [modifyAt](#modifyat)
   - [size](#size)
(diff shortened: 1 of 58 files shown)

### CODE REVIEW

**Comment by imcotton** [comment]
previously #1524?

**Comment by samhh** [comment]
@imcotton I think you're right that this covers that use case as well.

For my mental model, I think this PR is equivalent to:

```hs
do
  x <- Just 42
  y <- Just 1337
  let z = x * y -- this line
  etc
```

imcotton:
One thing I noticed from your implementation is allowing redefine existing keys, e.g.

```ts
pipe(
 option.Do,
 option.apS('x', some(42)),
 option.apS('y', some(1337)),
 option.bindMap('y', ({ x, y }) => x * y),  // <-- reassign y here
)
```

Is this intend behavior or oversight in compare to:

https://github.com/gcanti/fp-ts/blob/34b28fa9a6c968529f689db1f71d2ef0a91ea4b6/src/Chain.ts#L114-L119

in L117: `name: Exclude<N, keyof A>`?

Punie:
@imcotton I'll be honest, that is an oversight on my part and I will fix it for consistency with `bind` and `apS`.
However, I'd argue that this restriction is unnecessary for all of these methods as shadowing is a thing in "regular" do-notation.

Good catch, though. Thank you 🙏

imcotton:
> However, I'd argue that this restriction is unnecessary for all of these methods as shadowing is a thing in "regular" do-notation.

IMO, name shadowing from outer scope is reasonable (but I'd avoid it personally), however in case of `let` bind in `do` notation probably a sign of something off.

Overall I'm 👍 for this feature tho, thanks for the effort.

gcanti:
For what concerns the name I'd really like `let` (*), ...or maybe `bindValue` (<- already used by [effect-ts](https://github.com/Effect-TS/core/blob/0f274d99b3c93724945b1b9feff73314a0d25ecd/packages/core/_src/io/Exit/operations/do.ts#L35))

(*)

> as a reserved keyword in TS it was a no-go

```ts
const let_ = ...

export {
  /**
   * @since 2.13.0
   */
  let_ as let
}
```

is it a viable option?

gcanti:
@Punie some `N extends string` are missing I think

gcanti:
Never mind, I was looking at the md files :)

gcanti:
Thank you!

Issue #1216, opened 2020-05-19
Consider changing typeclasses signatures to be always data-last


gcanti:
Pros (from https://github.com/gcanti/fp-ts/pull/942#issuecomment-524912701)

- API inconsistency:  `pipe` is an idiomatic way to use `fp-ts` however its abstract core - the typeclasses - have signatures not suitable for this idiomatic way out of the box and require some messy (redundant IMO) steps like calling `pipeable`, destructuring, reexporting and so on. When they actually could be used _as is_.
- transformers: if I need a "pipeable" version (an idiomatic version actually) of some transformer I need to register its type as a separate monad on the `HKT` dictionary. This doesn't  make much sense because the type is already "bundled" in the transformer instance (for example the result of `getOptionM` already contains info about `Option` and passed `M`.
- `pipeable` module itself is a mess with no sensible profit - it just fixes data-first typeclasses when they could be data-last by design completely eliminating the need for this module. 

Cons

- breaking change
- performance implications?

Prior art

- https://github.com/gcanti/fp-ts/issues/925
- https://github.com/gcanti/fp-ts/pull/942

mikearnaldi:
I did some further testing, it doesn't look so bad performance wise (roughly 30% loss but case dependent), in case of critical places there are still means of optimization by exposing specific data-first functions outside of the instances so I'll also vote for it.

gcanti:
@raveclassic I have a proposal to make this change kind of "backward compatible". Switching to data-last type class is indubitably a breaking change for "power users" (e.g. library authors). However "normal" users always interact with type class instances only as a way to configure constrained functions, for example `A.sequence`

```ts
import * as A from 'fp-ts/lib/Array'
import * as E from 'fp-ts/lib/Either'
import { pipe } from 'fp-ts/lib/pipeable'

pipe(
  [E.right(1), E.right(2), E.right(3)],
  A.sequence(E.either), // <= as a user I really don't care what `E.either` actually is
  E.bimap(
    () => 'error',
    () => 'ok'
  )
)
```

Now `E.either` is pretty bad from a tree-shakeablity point of view so I'm thinking of splitting this kind of mega instances into specialized ones. For example an `applyEither` instance (which is now [internal](https://github.com/gcanti/fp-ts/blob/4471dcbac7f2d7f92be00cec85f8bbb9e2cea970/src/Either.ts#L532-L536), i.e. not yet ufficually exported), an `applicativeEither` instance, an `monadEither` instance, etc...

[...]

PR #1748, opened 2022-08-23
Flip the arguments of a curried function

thewilkybarkid:
I've been using this `flipC` in combination with `Reader`'s `local` to hide implementation details, based on https://andywhite.xyz/posts/2021-01-28-rte-react/#service-layers-and-implementation-hiding (e.g. https://github.com/PREreview/prereview.org/blob/35278c267c541586f7ca505b9bd9f5a634ec3e93/src/app.ts#L81).

This function dates back to @jskeate's suggestion in https://functionalprogramming.slack.com/archives/CPKPCAGP4/p1655736152988389?thread_ts=1655730853.886439&cid=CPKPCAGP4.

Another option could be to overload `flip`, though I've not tried that.


samhh:
This is one of the few cases where I used a conflicting name in [fp-ts-std](https://samhh.github.io/fp-ts-std/modules/Function.ts.html#flip) - I don't think uncurried functions should get the proper name `flip` in a FP library.

Could we perhaps target a breaking change in 3.0 that rewrites `flip` to the curried version in this PR (and optionally renames uncurried `flip` to something else)? v3 includes [other curry-friendly breaking changes](https://github.com/gcanti/fp-ts/issues/1216).

thewilkybarkid:
I've pushed 20268fc39144c5fc6fdca81bf1b9adfcf1d0c507 based on @samhh's suggestion to replace the existing `flip`: I've deprecated two arguments in favour of a new function `flipArgs`.

Not sure which approach you'd rather take @gcanti.

gcanti:
> Could we perhaps target a breaking change in 3.0 that rewrites `flip` to the curried version in this PR

I agree, I would:

- add the `export function flip<A, B, C>(f: (a: A) => (b: B) => C): (b: B) => (a: A) => C` overloading to the existing `flip` function
- deprecate the existing `export function flip<A, B, C>(f: (a: A, b: B) => C): (b: B, a: A) => C` signature
- keep the `@since 2.0.0` tag
- do not export a `flipArgs` function

thewilkybarkid:
@gcanti Should be sorted now.

gcanti (approved):

PR #1780, opened 2022-09-17
Add `getOrDefault` to `Option` and `Either` based types

Punie:
I added the `getOrDefault` function to `Option`, `Either` and associated transformer types.
This function allows to extract the value if present and default to the provided `Monoid`'s `empty` value if not.

**Examples:**
```ts
import { some, none, getOrDefault } from 'fp-ts/Option'
import { MonoidSum, MonoidProduct } from 'fp-ts/number'

assert.strictEqual(pipe(some(42), getOrDefault(MonoidSum)), 42)
assert.strictEqual(pipe(none, getOrDefault(MonoidSum)), 0)
assert.strictEqual(pipe(none, getOrDefault(MonoidProduct)), 1)
```

### DIFF
--- a/docs/modules/Either.ts.md
+++ b/docs/modules/Either.ts.md
@@ -100,6 +100,7 @@ Added in v2.0.0
 - [destructors](#destructors)
   - [fold](#fold)
   - [foldW](#foldw)
+  - [getOrDefault](#getordefault)
   - [getOrElse](#getorelse)
   - [getOrElseW](#getorelsew)
   - [match](#match)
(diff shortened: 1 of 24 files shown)


samhh:
NB this looks like [`fold`](https://hackage.haskell.org/package/base-4.17.0.0/docs/Data-Foldable.html#v:fold).

newswim:
I agree that `fold` is probably preferable, you could also say `getOrElse[W](() => m.empty)`.

Punie:
It is indeed very similar to `fold`/`match` and `getOrElse` but merely provided as a convenience here.

Feel free to close this PR if you think this is unnecessary (it is easily written as `getOrElse(() => m.empty)` as mentioned but I liked having a specific `getOrDefault` in the API).

gcanti:
Thanks @Punie but I agree with

> it is easily written as `getOrElse(() => m.empty)`

Going to close

PR #1787, opened 2022-09-20
package.json `exports` field: move `./*` to the end, closes #1786

### LINKED ISSUE #1786: 2.13 'lib' import errors in Jest
Just tried upgrading [a project](https://github.com/PREreview/prereview.org) to `2.13.0-rc.4` to see what happens, and found that the old-style `fp-ts/lib/` imports in io-ts cause Jest to fail:

```
Test suite failed to run

ENOENT: no such file or directory, open '[...]/node_modules/fp-ts/lib/lib/Either.js'

  at Runtime.readFile (node_modules/jest-runtime/build/index.js:2574:21)
  at Object.<anonymous> (node_modules/io-ts/lib/Decoder.js:27:22)
```

It might happen with other packages too (e.g. logging-ts), but it's probably not reached them yet.

Given the project runs successfully (with Node 16.17.0), this could be a Jest problem (or a problem with our configuration), but I thought it worth flagging here.

gcanti:


thewilkybarkid:
Yep, it was fine. Thanks!

PR #1793, opened 2022-10-02
Dual ESM/CJS Support, Move to Org, Improved Build (for v3)

mikearnaldi:
This PR introduces compatibility with ESM and temporarily dual compatibility with CJS.

- [x] move to a workspace
- [x] adapt import structure
- [x] introduce changesets
- [x] auto add PURE comments
- [x] stricter ts config
- [x] move to `@fp-ts` org
- [ ] port custom additional scripts (docs, linters, etc)

Note: this PR introduces temporarily build-scripts and configs from effect that will be rewritten soon to be generic (plain ts, no deps) and moved to the `@repo-tooling` org (that we control)

Note2: it isn't good practice to include generated content such as `docs` to the repo itself, it would be better to have a dedicated pipeline that generates that content to be published in a branch like https://github.com/mikearnaldi/pirum-meetup-09-2022/blob/main/.github/workflows/deploy.yml


steida:
I recently switched to PNPM and I suppose it's worth considering.

mikearnaldi:
> I recently switched to PNPM and I suppose it's worth considering.

I can give you access to the repo `matechs-garage/fp-ts` where this PR is managed from and you can give it a try, I have no experience with pnpm and yarn has worked nicely in the past but if you think it's better you're welcome to try

steida:
From what I can say so far, PNPM is crazy fast and handy just like Yarn except for no versions and Plug'n'Play confusion 😅 I am using it at https://github.com/evoluhq/evolu with changeset, etc. It seems to be mature and well-supported. I have no time right now (Evolu) to experiment on the repo, but I believe it can be updated later with no or little changes.

kalda341:
I have also found PNPM to be a huge improvement over yarn in every way.

mikearnaldi:
> I have also found PNPM to be a huge improvement over yarn in every way.

Over yarn 3? Can you be precise in saying what was better?

steida:
@mikearnaldi Yarn hacks node_modules with hoist dependencies while pnpm does not. And it's faster. https://divriots.com/blog/switching-to-pnpm

mikearnaldi:
> @mikearnaldi Yarn hacks node_modules with hoist dependencies while pnpm does not. And it's faster. https://divriots.com/blog/switching-to-pnpm

Hoisting dependencies is not a hack, it's perfectly compaiant with the logic of node_modules and is needed to avoid duplicates

kalda341:
@mikearnaldi Mostly support for monorepos. It's been far easier to manage dependencies across them correctly with isolated workspaces. We had a lot of cases with Yarn where the packages were defined incorrectly, but everything worked fine within the monorepo, but were badly broken when used outside. Monorepo support with yarn workspaces has always felt a bit hacky, whereas it's been a breeze with pnpm.

For performance in CI we went down the rabbit-hole of PnP and zero installs with Yarn (version 2 and briefly 3) which led to a lot of time wasted trying to figure out why things weren't working. It also made it very difficult to try out new versions of Typescript for example as the patches weren't necessarily released. PNPM has been fast enough that we haven't missed this feature of Yarn at all.

I've had way fewer issues with pnpm, and when I have had a problem I've found it much easier to debug - I'm not sure whether that's down to the docs, or just a better mental model of how it works.

mikearnaldi:
```bash
gitpod /workspace/fp-ts (chore/restructure-build-esm-support) $ curl -fsSL https://get.pnpm.io/install.sh | sh -
==> Extracting pnpm binaries 7.13.0
Copying pnpm CLI from /tmp/tmp.9iBUHMwYuF/pnpm to /home/gitpod/.local/share/pnpm/pnpm
 ERR_PNPM_UNSUPPORTED_SHELL  Can't setup configuration for "" shell. Supported shell languages are bash, zsh, and fish.
Install Error!
```

Not a great start.

mikearnaldi:
Screenshot 2022-10-03 114946

See the duplicated "fast-check" package? that is why yarn is way better. Duplicating packages is NOT safe even if packages are stateless as the two projects will end up using different instances of the modules (and something like a class defined in one will not be instance of a class defined in the other).

mikearnaldi:
Actually it's fine from node's perspective the issue only occurs in bundlers where symlinks are not properly handled which may be fine to use, I didn't notice any speed improvement over yarn 3 though, what I liked is binary linking works though which may simplify structure, I guess it's a choice between slightly different trade-offs

mikearnaldi:
Moved in https://github.com/fp-ts/fp-ts/pull/1

PR #1800, opened 2022-10-18
Add rte from reader io

vinassefranche:
While upgrading to 2.13.1, I found out that this was missing.
Tell me if something's wrong with the pull request!

And thanks for adding this new module. I was thinking about adding it myself so I was glad to see it in this version :-)


mixedCase:
I know this PR has been around for over a year, but this should be implemented as an alias to rightReaderIO, which fulfills the same purpose but has a less discoverable name.

PR #1815, opened 2022-11-26
Implement generic types for fromEntries of Record and ReadonlyRecord

dilame:
I am heavily use types, and most of my `Record` inputs are typed with generic keys, so i really miss this feature.

I was forced to use `Object.create(null)` because otherwise typescript throwed compile error.


arthurgubaidullin:
Awesome! I need it. I want use branded keys in records!

PR #1834, opened 2023-02-14
feat(option): better return type for Option constructors

JalilArfaoui:
This makes `some()` constructor return a `Some<A>` and `none` return a `None`, more specific than actual `Option<A>`.

This enables this kind of code to work :

```ts
const updateObjectOrCreate = (
  someObject: Option<SomeType>,
  transformer: (session: SomeType) => SomeType
): option.Some<SomeType> =>
  pipe(
    someObject,
    option.getOrElse(
      constant(
        makeObject(someSaneDefaults)
      )
    ),
    transformer,
    option.some
  );
```

without adding à `as option.Some<Session>`

### DIFF
--- a/src/Option.ts
+++ b/src/Option.ts
@@ -140,7 +140,7 @@ export type Option<A> = None | Some<A>
  * @category constructors
  * @since 2.0.0
  */
-export const none: Option<never> = _.none
+export const none: None = _.none
 
 /**
  * Constructs a `Some`. Represents an optional value that exists.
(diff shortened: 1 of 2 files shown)


JalilArfaoui:
Why is that @samhh ? I was under the idea that we want the most precise types we can have, don’t we ?

samhh:
I've yet to come across a scenario with sum types where it's not a code smell because of an issue with code structure somewhere else. Just because we can do something doesn't mean we should.

Less esoterically this might present an issue in some spots where currently widening isn't necessary. I know this was a consideration when designing [sum-types](https://github.com/unsplash/sum-types) though I don't have a specific example to hand.

asjir:
I'm afraid this might break with [typescript 5.5](https://www.totaltypescript.com/type-predicate-inference)
The reason being: define `arr=T[]`  then typescript gives `arr[i]: T` even though it can be `T | undefined` when it's out of bounds, then the type predicate inference might give you:
`O.fromNullable(arr[i]): Some<T>` even though it's `None` if it's out of bounds

Issue #1851, opened 2023-04-21
Backporting from effect/*


gcanti:
- `flatMap` (aka chain / chainW) https://github.com/gcanti/fp-ts/pull/1849
- `tap` (aka chainFirst / chainFirstW)

gcanti:
IMO it always makes sense since you get dual APIs and also a better name (tapError vs orElseFirst)

sukovanej:
`as` and `asUnit` would be also nice, hm?

- `Either`
- `IO`
- `IOEither`
- `IOOption`
- `Option`
- `Reader`
- `ReaderEither`
- `ReaderIO`
- `ReaderTask`
- `ReaderTaskEither`

Does it make sense to implement `as` and `asUnit` for any data type having a functor instance or is there rule / law / practical reasoning for not having it for some data types? I see, in the `effect/data`, it's not implemented for arrays or `Identity`. What about `State` for example?

gcanti:
No, there isn't a specific reason, they just don't seem very useful in the case of ReadonlyArray and Identity

gcanti:
> Are struct and tuple from Product considered stable enough to be backported?

@sukovanej I think it's better to wait a little longer

p.s.
Anything else you would like to add? otherwise we could release 2.16

sukovanej:
Ok. I don't have anything right now.

christophgietl:
How about `tap` for `Identity`? For me it feels a little bit awkward to write `O.tap` and `I.chainFirst`.

PR #1864, opened 2023-05-15
feat: tapEither

sukovanej:
#1851


gcanti (approved):

gcanti:
@sukovanej just one thing, for now I would keep the `tapEither` helper (in `FromEither.ts`) internal (as I did with `Chain.tap` for example) and postpone the decision on whether expose those helpers or not

PR #1874, opened 2023-05-21
Add ReaderTask, TaskEither and IO flatMaps

sukovanej:
- `flatMapReaderTask`
- `flatMapTaskEither`
- `flatMapIO` (+ internal-only new hkt type lambdas and instances)


sukovanej:
@gcanti at this stage, do you prefer introducing the new hkt internally? Now as I noticed it, I might have done the same e.g. for the `tapIO`.

gcanti:
> do you prefer introducing the new hkt internally?

Do you mean compared to the old system with overloading? When possible, yes, mostly for our convenience though since it's all internal stuff.

PR #1882, opened 2023-05-25
feat: Either mapError

sukovanej:
#1851

### DIFF
--- a/docs/modules/Either.ts.md
+++ b/docs/modules/Either.ts.md
@@ -98,7 +98,7 @@ Added in v2.0.0
   - [getApplicativeValidation](#getapplicativevalidation)
   - [getOrElse](#getorelse)
   - [getOrElseW](#getorelsew)
-  - [mapLeft](#mapleft)
+  - [mapError](#maperror)
   - [orElse](#orelse)
   - [orElseW](#orelsew)
 - [filtering](#filtering)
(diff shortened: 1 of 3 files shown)


docs/modules/Either.ts.md
```diff
@@ -98,7 +98,7 @@ Added in v2.0.0
   - [getApplicativeValidation](#getapplicativevalidation)
   - [getOrElse](#getorelse)
   - [getOrElseW](#getorelsew)
-  - [mapLeft](#mapleft)
+  - [mapError](#maperror)
```
gkamperis, line 101: conceptually Either contains 2 values. Don't think there is a certainty for Left to be an "error".

gcanti: Yeah, `mapLeft` in `Either.ts` was intentionally omitted (no `mapLeft` in `@effect/data/Either` as well)

sukovanej: Okay, kinda makes sense. The inconsistency is unfortunate tho as now the TaskEither or ReaderTaskEither have the `mapError` while the `Either` doesn't. 

True, the `Either` doesn't need to represent success / fail and can be used for anything where distinction for two outcomes is desired. But, at least from my experience, it is used for failable computations most of the time.

Issue #1610, opened 2021-10-28
Increase the number of parameters for `flow` from 9 to 20 to make it equivalent to `pipe`.


zanemayo:
### Current Behavior

`flow` supports 9 arguments

### Desired Behavior

`flow` supports 20 arguments (like pipe)

### Suggested Solution

Add some more declarations for `flow` in function.d.ts

### Who does this impact? Who is this for?

This seems helpful for building larger chains of functions. I sometimes have a long list of map, chain and bind.

### Describe alternatives you've considered

I've switched to using pipe for now.

### Your environment

| Software   | Version(s) |
| ---------- | ---------- |
| fp-ts      |    2.11.4        |
| TypeScript |    4.4.3        |

PR #1883, opened 2023-05-27
refactor: recursive type definition of pipe() and flow()

john-batch:
The current implementation and type definition of the pipe() and flow() functions limits the number of arguments you can pass to them. This PR fixes that without compromising with type-safety from the caller perspective.

I know the need for ts-expect-error doesn't make for the most elegant of solutions but TypeScript isn't smart enough yet to resolve this kind of type recursion outside of type declaration.

### DIFF
--- a/src/function.ts
+++ b/src/function.ts
@@ -233,112 +233,19 @@ export function flip(f: Function): Function {
  *
  * @since 2.0.0
  */
-export function flow<A extends ReadonlyArray<unknown>, B>(ab: (...a: A) => B): (...a: A) => B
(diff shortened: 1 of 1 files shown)


john-batch:
I have to mention that the TypeScript type inference algorithm imposes a limit on the depth of recursive types. I managed to pass up to 48 functions to pipe() and 49 functions to flow() before TypeScript reported TS2589 (Type instantiation is excessively deep and possibly infinite).

I guess it would be enough for #1610, and 99.9% of use cases really.

Issue #1916, opened 2023-12-04
Comment-related warnings when building with Vite


branko-d:
### Current Behavior

When building a production bundle with Vite, I get a bunch of warnings like this:

```
A comment

"/*#__PURE__*/"

in "node_modules/fp-ts/es6/Either.js" contains an annotation that Rollup cannot interpret due to the position of the comment. The comment will be removed to avoid issues.
```

### Expected behavior

No build warnings.

### Reproducible example

1. Go here: https://stackblitz.com/edit/vitejs-vite-y7ddkf?file=src%2FApp.tsx
2. This is just a standard Vite's [react-ts](https://vite.new/react-ts) example with "fp-ts" added to package.json and imported in App.tsx.
3. Click on the Terminal panel at the bottom.
4. Enter q + ENTER to quid Vite dev server.
5. Type: `vite build`.

You should be able to see the warnings in the build log.

### Suggested solution(s)

Not sure - probably fix the comments somehow?

### Your environment

_Which versions of fp-ts are affected by this issue?_

| Software   | Version(s) |
| ---------- | ---------- |
| fp-ts      | 2.16.1           |
| TypeScript | 5.2.2           |

_Did this work in previous versions of fp-ts?_

Not sure - we just transitioned away from Create React App (where we didn't get these warnings) to Vite recently.

kblcuk:
FIY I stumbled upon the same issue and after some investigation I think I might have a fix 🤞

cfgs:
> FIY I stumbled upon the same issue and after some investigation I think I might have a fix 🤞

How is the fix going? :)

kblcuk:
@cfgs you can see a bit above your comment that this issue was closed in https://github.com/gcanti/fp-ts/pull/1919 :)

So once you update to the latest fp-ts, the warnings should disappear for you as well!

PR #1919, opened 2024-01-03
Fix __PURE__ annotation placement

kblcuk:
Currently there are some arrow functions that have /*#__PURE__*/
annotation before function call to identify that that function call is
side effect free.

However since this project is built to es5 target, which doesn't have
arrow functions support, the code is then translated into regular
function that loos something like this:

```
/** @internal */
export var flatMapNullable = function (F, M) {
    /*#__PURE__*/ return dual(3, function (self, f, onNullable) {
        return M.flatMap(self, liftNullable(F)(f, onNullable));
    });
};
```

However this makes some built tools (Vite (which uses Rollup under the
hood) in our case) unhappy, and build produces a lot of warnings about
that __PURE__ annotation being in the wrong place.

Checking the Rollup docs [0], it seems that pure annotation should be
placed right before function invocation, so in this particular case
between `return` keyword and the actual function.

So this simply changes arrow functions in question to have explicit
return keyword and annotation before the function call, which leaves no
interpretation to ts compiler, and makes our code build process
warning-free.

This also potentially fixes https://github.com/gcanti/fp-ts/issues/1916

[0] https://rollupjs.org/configuration-options/#pure

### DIFF
--- a/src/Either.ts
+++ b/src/Either.ts
@@ -1198,7 +1198,7 @@ export const chainOptionK: <E>(
  */
 export const chainOptionKW: <E2>(
   onNone: LazyArg<E2>
-) => <A, B>(f: (a: A) => Option<B>) => <E1>(ma: Either<E1, A>) => Either<E1 | E2, B> = /*#__PURE__*/ chainOptionK as any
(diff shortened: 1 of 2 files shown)


gcanti:
Thank you @kblcuk, going to release this asap

Issue #1929, opened 2024-03-25
[ReadonlyArray.ts] `flatten` can trigger `RangeError: Maximum call stack size exceeded`


christiantakle:
### Current Behavior

`flatten( [ <large-inner-array>, ...] )` trigger
```
RangeError: Maximum call stack size exceeded
```

### Expected behavior

Not to blow up.

### Reproducible example

```
import * as RA from "fp-ts/ReadonlyArray";

const value = {};
type Inner = Array<typeof value>;
// const a: Array<Array<typeof value>> = [new Array(112201).fill(value)];
/**
 * 112201, the number that failed in production and Apple  M1 Max.
 * I had to increase the number to see a similar failure on my desktop workstation hence the
 * `* 2`
 */
const ass: Array<Inner> = [new Array(112201 * 2).fill(value)];

const main = () => {
  console.log({ assLength: ass.length });
  try {
    const as = RA.chainWithIndex((_, x: Inner) => x)(ass);
    console.log(as.length);
  } catch (err: any) {
    console.log("[original][catch]", err.message);
  }
};

main();
```

Smaller example of the exception without using fp-ts:
```
// b.js
const xs = [];
xs.push(...new Array(100_000 * 2).fill(0));
```

```
❯ node b.js
/tmp/b.js:2
xs.push(...new Array(100_000 * 2).fill(0));
   ^

RangeError: Maximum call stack size exceeded
    at Object.<anonymous> (/tmp/b.js:2:4)
    at Module._compile (node:internal/modules/cjs/loader:1376:14)
    at Module._extensions..js (node:internal/modules/cjs/loader:1435:10)
    at Module.load (node:internal/modules/cjs/loader:1207:32)
    at Module._load (node:internal/modules/cjs/loader:1023:12)
    at Function.executeUserEntryPoint [as runMain] (node:internal/modules/run_main:135:12)
    at node:internal/main/run_main_module:28:49

Node.js v20.11.1
```

### Suggested solution(s)

```
Use nested `for` loop
```

### Your environment

> Which versions of fp-ts are affected by this issue?

All versions that include the current implementation.

> Did this work in previous versions of fp-ts?

Don't think so but have not seen it since it depends on input.

| Software   | Version(s) |
| ---------- | ---------- |
| fp-ts      | 2.16.3     |
| TypeScript | 5.4.2      |

PR #1930, opened 2024-03-25
Update ReadonlyArray.ts

christiantakle:
Fix: #1929

Not sure about adding a test here since the size of the input depend on nodejs and/or the machine its running on.

# Example using the change

Note the source code is inline so its simpler to play with.

```
/** Inlined from source code */
const isEmpty = <A>(as: ReadonlyArray<A>): as is readonly [] => as.length === 0;
// const empty: ReadonlyArray<never> = RNEA.empty
const empty: ReadonlyArray<never> = [];

const value = {};
// const a: Array<Array<typeof value>> = [new Array(112201).fill(value)];
/**
 * 112201, the number that failed in production and on an Apple M1 Max 64gb.
 * I had to increase the number to see a similar failure on my desktop workstation hence the
 * `* 2`
 */
const ass: Array<Array<typeof value>> = [new Array(112201 * 2).fill(value)];

// chainWithIndex from ReadonlyArray
const chainWithIndex =
  <A, B>(f: (i: number, a: A) => ReadonlyArray<B>) =>
  (as: ReadonlyArray<A>): ReadonlyArray<B> => {
    if (isEmpty(as)) {
      return empty;
    }
    const out: Array<B> = [];
    for (let i = 0; i < as.length; i++) {
      out.push(...f(i, as[i]));
    }
    return out;
  };

// Alternative implementation
export const chainWithIndex2 =
  <A, B>(f: (i: number, a: A) => ReadonlyArray<B>) =>
  (as: ReadonlyArray<A>): ReadonlyArray<B> => {
    if (isEmpty(as)) {
      return [];
    }
    const out: Array<B> = [];
    for (let i = 0; i < as.length; i++) {
      const bs = f(i, as[i]);
      for (let j = 0; j < bs.length; j++) {
        out.push(bs[j]);
      }
    }
    return out;
  };

const main = () => {
  console.log({ assLength: ass.length });

  {
    console.log("original start");
    try {
      const as = chainWithIndex((_, x) => x)(ass);
      console.log({ asLength: as.length });
    } catch (err: any) {
      console.log("[original][catch]", err.message);
    }
    console.log("original end");
  }

  {
    console.log("alternative start");
    const as = chainWithIndex2((_, x) => x)(ass);
    console.log({ asLength: as.length });
    console.log("alternative end");
  }
  console.log("done");
};

main();
```

Command used to run it:
```
ts-node --transpile-only x.ts
```

Output:
```
{ assLength: 1 }
original start
[original][catch] Maximum call stack size exceeded
original end
alternative start
{ asLength: 224402 }
alternative end
done
```

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "fp-ts",
-  "version": "2.15.0",
+  "version": "2.16.4",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
       "name": "fp-ts",
-      "version": "2.15.0",
+      "version": "2.16.4",
       "license": "MIT",
       "devDependencies": {
         "@effect/language-service": "^0.0.19",
--- a/src/Array.ts
+++ b/src/Array.ts
@@ -404,7 +404,10 @@ export const chainWithIndex =
   (as: Array<A>): Array<B> => {
     const out: Array<B> = []
     for (let i = 0; i < as.length; i++) {
-      out.push(...f(i, as[i]))
+      const bs = f(i, as[i])
+      for (let j = 0; j < bs.length; j++) {
+        out.push(bs[j])
+      }
     }
     return out
   }
--- a/src/NonEmptyArray.ts
+++ b/src/NonEmptyArray.ts
@@ -569,7 +569,10 @@ export const chainWithIndex =
   (as: NonEmptyArray<A>): NonEmptyArray<B> => {
     const out: NonEmptyArray<B> = fromReadonlyNonEmptyArray(f(0, head(as)))
     for (let i = 1; i < as.length; i++) {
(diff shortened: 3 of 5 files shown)


gcanti:
Thanks @christiantakle

The issue lies with the use of Array.push in conjunction with the spread operator. There are three other instances of this problem. You can locate them by searching for `".push(..."`. Could you please address these occurrences as well?

christiantakle:
@gcanti Done. 

I do make a change to `package-lock.json` when running `npm install` assume its okay to include it here.

christiantakle:
If you want I can merge the commits for the `.push(...` change or kill this pr and make a new one with that change just to reduce noise and co-locate import change.

christiantakle:
Prefer #1931

PR #1949, opened 2024-07-17
Support `strictBuiltinIteratorReturn`

rbuckton:
This updates the `fp-ts` build to be forwards compatible with microsoft/TypeScript#58243, which uses a stricter return type for built-in iterators under the new `strictBuiltinIteratorReturn` option, which is enabled by default when compiling with `strict`.

### DIFF
--- a/src/Map.ts
+++ b/src/Map.ts
@@ -217,10 +217,7 @@ export function pop<K>(E: Eq<K>): (k: K) => <A>(m: Map<K, A>) => Option<[A, Map<
   }
 }
 
-interface Next<A> {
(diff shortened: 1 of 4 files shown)


rbuckton:
I've filed this as a Draft PR in the event there are any other changes to https://github.com/microsoft/TypeScript/pull/58243 that must be addressed. Once that PR has merged, I will mark this PR as "Ready for review".

gcanti (approved):

gcanti:
Thank you @rbuckton 👍

